PR #1, opened 2026-04-18
claude github integrated workflow

xbrianh:


bin/ghimplement.sh
```diff
@@ -0,0 +1,91 @@
...
+REPO=$(gh repo view --json nameWithOwner -q .nameWithOwner) \
+  || die "not in a gh-recognized repo"
+
+CLAUDE_FLAGS=(--permission-mode bypassPermissions --output-format stream-json --verbose)
+
+extract_url() {
+  local out="$1" pattern="$2" label="$3"
+  local result url
+  result=$(jq -r 'select(.type=="result") | .result' <<<"$out" | tail -1)
+  url=$(grep -oE "$pattern" <<<"$result" | tail -1 || true)
```
xbrianh: `tail -1` of any URL matching the pattern is brittle: if the model's output ever mentions a *different* issue/PR URL in context (a linked ticket, the base repo, a comment excerpt), this silently grabs the wrong one and the rest of the script proceeds against the wrong resource.

Safer options: (a) scan the stream-json for the `tool_result` of `gh issue create` / `gh pr create` and parse the URL from there; or (b) ask the final message to emit a distinctive sentinel (`NEW_ISSUE_URL=<url>`) and match that specifically.

bin/ghimplement.sh
```diff
@@ -0,0 +1,91 @@
...
+    echo "--- raw claude output ($label) ---" >&2
+    echo "$out" >&2
+    echo "--- end raw output ---" >&2
+    die "failed to extract $label URL"
+  fi
+  echo "$url"
+}
+
+echo "==> [1/6] running /ghplan"
+PLAN_OUT=$(claude -p "${CLAUDE_FLAGS[@]}" "/ghplan ${REF} ${INSTRUCTIONS}" | progress_tee)
```
xbrianh: `${REF}` and `${INSTRUCTIONS}` are interpolated unquoted into the prompt. Two issues:

1. Empty `REF` produces `/ghplan  <instructions>` (double space) — harmless but ugly; use `${REF:+$REF }` to add a trailing space only when non-empty.
2. Combined with `--permission-mode bypassPermissions` on line 32, arbitrary CLI args become part of a prompt given to an unrestricted agent. Worth validating `REF` against a git-ref charset (`[A-Za-z0-9._/-]`) and documenting that `INSTRUCTIONS` is effectively a prompt, not an opaque arg.

bin/ghimplement.sh
```diff
@@ -0,0 +1,91 @@
...
+ISSUE_NUM=$(basename "$ISSUE_URL")
+echo "    issue: $ISSUE_URL"
+
+echo "==> [2a/6] implementing plan"
+claude -p "${CLAUDE_FLAGS[@]}" \
+  "Implement the plan in GitHub issue $ISSUE_URL. Read the issue body for the full plan, then make the code changes in this repo. Do not commit or push yet." \
+  | progress_tee >/dev/null
+
+echo "==> [2b/6] committing + opening PR"
+PR_OUT=$(claude -p "${CLAUDE_FLAGS[@]}" --continue \
```
xbrianh: `claude -p --continue` resumes the *most recent* claude session on this machine. That implicitly assumes nothing else started a newer session between step 2a and 2b — which is fine interactively but fragile if this script is run in parallel, or if the `-p` stream-json flow ever changes what counts as the 'latest' session.

Consider capturing the session ID from 2a's stream-json (`.session_id` on the `init` event) and resuming with `--resume <id>` instead, so the binding is explicit.

bin/ghimplement.sh
```diff
@@ -0,0 +1,91 @@
...
+echo "==> [1/6] running /ghplan"
+PLAN_OUT=$(claude -p "${CLAUDE_FLAGS[@]}" "/ghplan ${REF} ${INSTRUCTIONS}" | progress_tee)
+ISSUE_URL=$(extract_url "$PLAN_OUT" 'https://github\.com/[^ )]+/issues/[0-9]+' "issue")
+ISSUE_NUM=$(basename "$ISSUE_URL")
+echo "    issue: $ISSUE_URL"
+
+echo "==> [2a/6] implementing plan"
+claude -p "${CLAUDE_FLAGS[@]}" \
+  "Implement the plan in GitHub issue $ISSUE_URL. Read the issue body for the full plan, then make the code changes in this repo. Do not commit or push yet." \
+  | progress_tee >/dev/null
```
xbrianh: No verification between 2a and 2b that the model actually modified files and didn't already commit/push (despite the instruction). If 2a no-ops or commits prematurely, 2b's 'create branch from default branch' either opens an empty PR or drops the work.

A quick guard before 2b would help — e.g. `git status --porcelain | grep -q .` to assert uncommitted changes exist, plus `git rev-parse HEAD` compared to the default branch to assert no new commits yet.

bin/ghimplement.sh
```diff
@@ -0,0 +1,91 @@
...
+
+echo "==> [4/6] running /ghreview"
+claude -p "${CLAUDE_FLAGS[@]}" "/ghreview $PR_URL" | progress_tee >/dev/null
+
+echo "==> [5/6] waiting for Copilot review (20s interval, 10min timeout)"
+deadline=$(( $(date +%s) + 600 ))
+while true; do
+  state=$(gh api "repos/$REPO/pulls/$PR_NUM/reviews" \
+    --jq '.[] | select(.user.login | test("[Cc]opilot")) | .state' 2>/dev/null \
+    | grep -E 'COMMENTED|CHANGES_REQUESTED|APPROVED' | head -1 || true)
```
xbrianh: This filter only accepts `COMMENTED | CHANGES_REQUESTED | APPROVED`. If Copilot returns `DISMISSED` (or any future state), the loop waits out the full 10-minute timeout and then `die`s — even though a terminal review state was actually reached.

Simpler: treat any non-empty, non-`PENDING` state as terminal, e.g. `grep -vE '^(PENDING)?$'`.

xbrianh (commented):
Solid orchestration of plan → implement → review → address. Main concerns are brittleness in coupling this shell wrapper to the `claude` agent: URL extraction via `tail -1` can grab unrelated URLs, `--continue` implicitly binds the commit step to 'most recent session' (fragile under concurrency), and the Copilot poll filter is narrower than the set of states GitHub can return. With `bypassPermissions` in effect, unvalidated user input also flows straight into the prompt. Comments below are concrete fixes.

bin/ghimplement.sh
xbrianh: Switched to extracting URLs from the stream-json tool_result paired to the specific `gh issue create` / `gh pr create` Bash tool_use (matched by tool_use_id). The prior fallback on the final result text is still there as a last resort but should no longer be what we hit in practice, so a linked ticket or context URL can't be mistaken for the created resource.

bin/ghimplement.sh
xbrianh: Fixed both:
1. Now interpolated as `${REF:+$REF }` so empty REF no longer produces a double space.
2. Added a charset check on REF (`^[A-Za-z0-9._/#-]+$`) before it ever gets concatenated into a prompt. Also added a comment documenting that INSTRUCTIONS is deliberately treated as a prompt to an unrestricted agent — callers should pass it with that expectation.

bin/ghimplement.sh
xbrianh: Good catch. Replaced `--continue` with an explicit `--resume <session_id>`. Step 2a now captures `.session_id` from the stream-json `system/init` event and step 2b resumes that exact session, so parallel runs (or any future change to what "latest" means) won't silently bind to the wrong conversation.

bin/ghimplement.sh
xbrianh: Added a guard between 2a and 2b:
- Record `git rev-parse HEAD` before 2a.
- After 2a, `die` if HEAD advanced (the model committed despite being told not to) or if `git status --porcelain` is empty (the model no-op'd).

Either condition fails fast rather than opening an empty PR or dropping the work.

bin/ghimplement.sh
xbrianh: Switched the filter to `grep -vE '^(PENDING)?$'` so any non-empty, non-PENDING state (including DISMISSED or any state GitHub adds later) is treated as terminal. Added a comment on-site explaining why we do not enumerate states.

PR #2, opened 2026-04-18
script to sync claude config

xbrianh:


scripts/sync.sh
```diff
@@ -0,0 +1,146 @@
...
+BASE_FLAGS=(-a --itemize-changes)
+[[ $DRY -eq 1 ]] && BASE_FLAGS+=(--dry-run)
+
+sync_file() {
+    local src="$1" dst="$2"
+    if [[ ! -e "$src" ]]; then
+        echo "skip: $src (not present)"
+        return
+    fi
+    mkdir -p "$(dirname "$dst")"
```
xbrianh, line 94: `rsync --delete` on `push` is a significant footgun. If the repo's `skills/` (or `agents/`, `commands/`) is empty — because of a stale branch, a bad merge, or a freshly-cloned checkout on another machine where `pull` was never run — `./sync.sh push -y` wipes `~/.claude/skills/` without warning. The single confirm prompt doesn't distinguish 'overwrite a few files' from 'delete everything'.

Mitigations, in order of effort:
1. Surface the delete count in the confirm prompt by running an rsync `--dry-run` first and counting `*deleting` lines.
2. Add a `--force` requirement when the pending deletion count exceeds a threshold (say, >5 files).
3. Default `push`/`pull` to dry-run; require `--apply` to actually write. Safer but more verbose.

scripts/sync.sh
```diff
@@ -0,0 +1,146 @@
...
+
+Commands:
+  pull        Copy from ~/.claude/ into this repo (mirror: deletes extras in repo)
+  push        Copy from this repo into ~/.claude/ (mirror: deletes extras in ~/.claude)
+  diff        Show differences between repo and ~/.claude/
+  status      Alias for diff
+
+Options:
+  -y, --yes       Skip confirmation prompt
+  -n, --dry-run   Show what would change without writing
```
xbrianh, line 42: `usage()` always `exit 1`s and writes to stdout, which conflates help with error. Two consequences:

- `./sync.sh --help` (no command) ends up here via the `*)` case and exits 1, so `echo $?` after `--help` reports failure.
- Redirecting usage to stdout means `./sync.sh --help > help.txt` works but `./sync.sh bogus 2> err.txt` loses the usage text to the terminal.

Typical pattern: `usage() { cat <<EOF >&$1 ...; exit $2; }` then `usage 1 1` for errors, `usage 2 0` for `-h|--help`. Or split into `help()` and `die_usage()`.

scripts/sync.sh
```diff
@@ -0,0 +1,146 @@
...
+EOF
+    exit 1
+}
+
+[[ $# -lt 1 ]] && usage
+
+CMD="$1"; shift || true
+YES=0
+DRY=0
```
xbrianh, line 67: `-h|--help` is only recognized *after* a command is parsed, because this loop runs after `CMD="$1"; shift`. So `./sync.sh --help` takes the `--help` flag as the command, falls through to the `*)` case, and prints usage via the error path (exit 1, stdout).

Handle the help flag before extracting `CMD`:

```bash
case "${1:-}" in
    -h|--help|"") usage ;;
esac
```

scripts/sync.sh
```diff
@@ -0,0 +1,146 @@
...
+do_push() {
+    for pair in "${FILE_PAIRS[@]}"; do
+        split_pair "$pair"
+        sync_file "$REPO_PATH" "$HOME_PATH"
+    done
+    for pair in "${DIR_PAIRS[@]}"; do
+        split_pair "$pair"
+        sync_dir "$REPO_PATH" "$HOME_PATH"
+    done
+}
```
xbrianh, line 180: When a tracked file exists on only one side (e.g. `bin/ghimplement.sh` exists in `~/bin/` but not yet in the repo), `diff -u` emits `diff: <path>: No such file or directory` to stderr and sets `rc=1`. The outer loop handles the rc, but the stderr noise is confusing.

The directory branch below already uses `-N` (treat absent files as empty) — add it here too: `diff -uN "$REPO_PATH" "$HOME_PATH"`.

scripts/sync.sh
```diff
@@ -0,0 +1,146 @@
...
+    pull)
+        confirm "Pull tracked files into $REPO_ROOT/ (will delete extras in repo dirs)?" || { echo "aborted"; exit 1; }
+        do_pull
+        ;;
+    push)
+        confirm "Push tracked files from $REPO_ROOT/ into \$HOME (will delete extras in home dirs)?" || { echo "aborted"; exit 1; }
+        do_push
+        ;;
+    diff|status)
+        do_diff || true
```
xbrianh, line 211: Minor wording: the confirm says 'will delete extras in home dirs', but `--delete` is only applied to `DIR_PAIRS` entries. Files in `FILE_PAIRS` (`CLAUDE.md`, `settings.json`, `bin/ghimplement.sh`) are overwritten but never deleted. The message implies files in `$HOME` could disappear, which isn't true.

Suggest: `'Push tracked files from ... (will mirror skills/agents/commands into \$HOME, deleting extras in those dirs)?'`.

xbrianh (commented):
Script is clear and the file/dir pair abstraction is nice. Main concern is the destructive surface of `--delete` on `push`: an empty or stale repo dir will silently wipe the corresponding home dir with a single yes/no confirm. A few smaller nits on help/usage ergonomics and diff noise.

scripts/sync.sh
xbrianh: Addressed in 913acfd. Before the confirm prompt, push/pull now runs an rsync --dry-run --delete across the DIR_PAIRS and surfaces a per-directory deletion count on stderr. If the total exceeds DELETE_THRESHOLD (5), the script refuses to proceed unless --force is passed (dry-run bypasses the threshold). This keeps the one-liner `./sync.sh push -y` safe: a stale/empty repo dir would now abort with a visible count rather than silently wiping `~/.claude/skills/`.

scripts/sync.sh
xbrianh: Addressed in 913acfd. Split into `help_text()` (writes to stdout, no exit) and `die_usage()` (optional error message to stderr, then help_text to stderr, exit 1). `--help` now exits 0 with usage on stdout; argument errors exit 1 with the error message and usage on stderr.

scripts/sync.sh
xbrianh: Addressed in 913acfd. Added a `case "${1:-}"` block before `CMD="$1"` that handles `-h|--help` (exit 0 via help_text) and empty args (die_usage). Verified `./sync.sh --help` now prints to stdout and exits 0.

scripts/sync.sh
xbrianh: Addressed in 913acfd: file branch now uses `diff -uN` too, so a file present on only one side is treated as empty on the other rather than emitting "No such file or directory" to stderr.

scripts/sync.sh
xbrianh: Addressed in 913acfd. Reworded both confirm prompts to make clear that only `skills/`, `agents/`, and `commands/` are mirrored with deletions; the tracked files (`CLAUDE.md`, `settings.json`, `bin/ghimplement.sh`) are only overwritten.

PR #8, opened 2026-04-18
Colocate ghimplement.sh with its skill to drop the ~/bin dependency

### LINKED ISSUE #7: Move ghimplement.sh into skills/ghimplement/ to drop the ~/bin dependency
## Context

`bin/ghimplement.sh` is currently synced to `~/bin/ghimplement.sh`, but `~/bin` is not universal — anyone adopting this repo as their Claude Code config must have it on `PATH` (or create it) for the `/ghimplement` skill to work. By contrast, every Claude Code user has `~/.claude/skills/`, which the sync already mirrors via `DIR_PAIRS`.

Colocating the script with its sole consumer (the `ghimplement` skill) makes the skill self-contained, removes the `~/bin` dependency, and eliminates a `FILE_PAIRS` entry.

## Approach

Move the script to `skills/ghimplement/ghimplement.sh` so it rides along with the skill via the existing `skills:$CLAUDE_DIR/skills` `DIR_PAIR`. The file moves as-is — it does not reference its own path, and rsync's `-a` preserves the executable bit. Then update every reference (sync script, SKILL.md, repo CLAUDE.md, README) to the new path.

Considered alternatives:
- **Leave it where it is**: keeps the `~/bin` PATH assumption baked in.
- **Inline the script into SKILL.md**: loses the ability to run/test it standalone and bloats the skill prompt.

## Tasks

- [ ] Move the script: `git mv bin/ghimplement.sh skills/ghimplement/ghimplement.sh`; remove the now-empty `bin/` directory.
- [ ] `scripts/sync.sh`: delete the `\"bin/ghimplement.sh:\$HOME/bin/ghimplement.sh\"` entry from `FILE_PAIRS` (line 22) and drop the `~/bin/ghimplement.sh` line from the header comment (lines 2–7).
- [ ] `skills/ghimplement/SKILL.md`: update three path references — `description:` frontmatter, `allowed-tools:` (`Bash(~/.claude/skills/ghimplement/ghimplement.sh:*)`), and the body prose + fenced invocation block.
- [ ] `CLAUDE.md` (repo): drop \"(plus `~/bin/ghimplement.sh`)\" on line 3; remove the `bin/ghimplement.sh → ~/bin/ghimplement.sh` entry from the `FILE_PAIRS` list on line 13; update the link on line 26 to ``[\`skills/ghimplement/ghimplement.sh\`](skills/ghimplement/ghimplement.sh)``.
- [ ] `README.md`: drop \"(plus `~/bin/ghimplement.sh`)\" on line 3; remove the `bin/ghimplement.sh` line from the layout diagram (line 11); rewrite the `ghimplement/` sub-bullet (line 14) to note the colocated script; change `~/.claude and ~/bin` → `~/.claude` on line 20; update the link target on line 44 and the path on line 48.
- [ ] PR description: call out the post-merge cleanup needed on each user's machine (`FILE_PAIRS` has no delete semantics, so the stale `~/bin/ghimplement.sh` will not be removed by `push`).

## Verification

1. `scripts/sync.sh diff` — expect the new script to appear under `skills/` and no `bin/ghimplement.sh` entry.
2. `scripts/sync.sh push --dry-run` — confirm the script rsyncs into `~/.claude/skills/ghimplement/ghimplement.sh` with the exec bit preserved (`-rwxr-xr-x`), and no `DELETE_THRESHOLD` warnings appear.
3. `scripts/sync.sh push` — perform the sync.
4. `ls -l ~/.claude/skills/ghimplement/ghimplement.sh` — confirm presence + exec bit.
5. Smoke test: run `~/.claude/skills/ghimplement/ghimplement.sh` with no args — expect the usage `die` message (exit non-zero, prints `usage: …`).
6. Invoke `/ghimplement` with a trivial instruction in a scratch repo to confirm the skill wrapper resolves the new path.
7. Manual cleanup on each user's machine after merge: `rm ~/bin/ghimplement.sh && rmdir ~/bin 2>/dev/null || true`.

## Open questions

- Should `scripts/sync.sh` grow a one-off cleanup shim that removes `~/bin/ghimplement.sh` if present, so adopters don't need a manual step? (Probably not worth the complexity for a single file, but flagging.)
- Worth keeping a top-level `bin/` directory as a deliberate \"no shared scripts here\" marker, or fully remove it? Plan currently removes it.

xbrianh:
## Summary

- Move `bin/ghimplement.sh` → `skills/ghimplement/ghimplement.sh` so the script rides along with the skill via the existing `skills:$CLAUDE_DIR/skills` `DIR_PAIR`, removing the `~/bin` PATH assumption.
- Drop the `bin/ghimplement.sh:$HOME/bin/ghimplement.sh` entry from `scripts/sync.sh` `FILE_PAIRS` (and the matching header-comment line).
- Update all references — `skills/ghimplement/SKILL.md` (description, `allowed-tools`, prose, fenced invocation), repo `CLAUDE.md`, and `README.md` (layout diagram, sub-bullet, sync-target line, two link targets).

## Post-merge cleanup

`FILE_PAIRS` has no delete semantics, so `scripts/sync.sh push` will **not** remove the stale `~/bin/ghimplement.sh` already installed on each user's machine. After merging, run manually:

```
rm ~/bin/ghimplement.sh
rmdir ~/bin 2>/dev/null || true
```

## Test plan

- [x] `git mv` preserved the executable bit on `skills/ghimplement/ghimplement.sh`.
- [x] `./skills/ghimplement/ghimplement.sh` with no args prints the expected `usage: …` message and exits non-zero.
- [x] `scripts/sync.sh push --dry-run --yes` shows `ghimplement.sh` rsyncing into `~/.claude/skills/ghimplement/`, no `bin/` entry, no `DELETE_THRESHOLD` warnings.
- [ ] After merge: `scripts/sync.sh push` on a dev machine and confirm `~/.claude/skills/ghimplement/ghimplement.sh` lands with exec bit, then smoke-test `/ghimplement` in a scratch repo.
- [ ] After merge: remove stale `~/bin/ghimplement.sh` per the cleanup block above.

Closes #7


scripts/sync.sh
```diff
@@ -19,7 +18,6 @@ DELETE_THRESHOLD=5
 FILE_PAIRS=(
     "home/CLAUDE.md:$CLAUDE_DIR/CLAUDE.md"
     "settings.json:$CLAUDE_DIR/settings.json"
-    "bin/ghimplement.sh:$HOME/bin/ghimplement.sh"
 )
```
xbrianh, line 21: Minor suggestion — since `FILE_PAIRS` has no delete semantics, the stale `~/bin/ghimplement.sh` will sit around on every existing install until users manually run the cleanup block from the PR description. Could add a one-shot detect-and-warn at the top of `do_push` / `do_pull`, e.g.:

```bash
if [[ -f "$HOME/bin/ghimplement.sh" ]]; then
    echo "note: $HOME/bin/ghimplement.sh is no longer tracked; remove it with: rm $HOME/bin/ghimplement.sh" >&2
fi
```

Nice-to-have, not a blocker — the README/PR body already documents the manual step.

skills/ghimplement/SKILL.md
```diff
@@ -1,11 +1,11 @@
 ---
 name: ghimplement
-description: Run the end-to-end plan → implement → review → address workflow by invoking ~/bin/ghimplement.sh. Creates a GitHub issue, opens a PR implementing it, collects Copilot + Claude reviews, and addresses them.
+description: Run the end-to-end plan → implement → review → address workflow by invoking ~/.claude/skills/ghimplement/ghimplement.sh. Creates a GitHub issue, opens a PR implementing it, collects Copilot + Claude reviews, and addresses them.
 argument-hint: [-r <ref>] <instructions>
-allowed-tools: Bash(~/bin/ghimplement.sh:*)
+allowed-tools: Bash(~/.claude/skills/ghimplement/ghimplement.sh:*)
```
xbrianh, line 5: Nit: the absolute path `~/.claude/skills/ghimplement/ghimplement.sh` is now hardcoded in three places in this file (`description`, `allowed-tools`, and the fenced invocation block on line 21). Not a change required here, but worth noting that a future rename would touch all three. If Claude Code ever exposes a skill-local root variable (e.g. `$CLAUDE_PLUGIN_ROOT`), switching to that would make this self-locating. Fine as-is for now.

xbrianh (commented):
Clean, well-scoped refactor. The `git mv` preserves the exec bit, every tracked reference (sync.sh FILE_PAIRS, its header comment, SKILL.md frontmatter + prose, CLAUDE.md, README.md layout diagram + link targets) is consistently updated, and the post-merge implication (FILE_PAIRS has no delete semantics, so `~/bin/ghimplement.sh` stays behind) is called out in the PR body.

One soft suggestion inline about automating that stale-file cleanup so users don't have to remember it. Otherwise this looks ready.

scripts/sync.sh
xbrianh: Good call — added `warn_stale_files` helper called at the top of `do_push` / `do_pull` in 037850d. Tested locally (the stale file was actually present on my dev box): `./scripts/sync.sh push --dry-run --yes` now emits `note: /Users/xath/bin/ghimplement.sh is no longer tracked; remove it with: rm /Users/xath/bin/ghimplement.sh` before the rsync itemize block.

skills/ghimplement/SKILL.md
xbrianh: Noted — agreed it's worth flagging. Leaving as-is for this PR since the reviewer marked it non-blocking, but filed the `$CLAUDE_PLUGIN_ROOT`-style self-locating refactor idea as a follow-up worth revisiting if/when Claude Code exposes such a variable.

PR #18, opened 2026-04-20
Move workflow runtime state out of ~/.claude/ to XDG_STATE_HOME

### LINKED ISSUE #17: Move all workflow runtime state out of ~/.claude/ to XDG_STATE_HOME
# Move all workflow runtime state out of `~/.claude/` to `XDG_STATE_HOME`

## Context

Two related problems with where `/ghimplement` and `/localimplement` put their runtime data:

1. **`/localimplement` is probabilistically broken.** PR #14 moved plan and review artifacts into `~/.claude/workflows/<WF_ID>/artifacts/` to keep scaffolding off product branches. But `~/.claude/` is classified as a *sensitive* directory by Claude Code, and writes there are blocked **even under `--permission-mode bypassPermissions`** — the sensitive-file check lives above that flag. Subagent stages that need to `Write`/`Bash(cat >)`/`Bash(tee)` an artifact file there hit the guardrail. Two `/localimplement` runs survived by accident (subagents stumbled into `python -c` reading stdin, `tee <<EOF`); a third run failed mid-pipeline when the detail-reviewer subagent asked for permission and gave up, tripping `die \"review sonnet did not produce …\"`. The bug is real and will recur.
2. **`/ghimplement` shares the guardrail fragility.** `/ghimplement` doesn't write artifacts, but every workflow's `state.json`, `log`, `pid`, `finished`, `acknowledged` live under `~/.claude/workflows/<WF_ID>/` and are written by `launch.sh`, `finish.sh`, `set-stage.sh`, and read by `workflows.sh` / `session-summary.sh`. Those writes currently happen in raw shell (not inside a Claude subagent) so the guardrail doesn't fire today — but any future change that has a subagent touch those paths would trip. Colocating **all** workflow runtime state in a non-sensitive directory removes the class of bug, not just this instance.

## Goals

- Every path a workflow touches — state, log, pid, markers, artifacts, stream jsonl — lives under `${XDG_STATE_HOME:-$HOME/.local/state}/claude-workflows/`. Nothing under `~/.claude/` related to workflow runtime.
- Applies to both `/ghimplement` and `/localimplement` pipelines, `/workflows` status skill, and the `SessionStart` / `UserPromptSubmit` hook (`session-summary.sh`).
- Direct-CLI `localimplement.sh \"<instructions>\"` (no `WF_ID`) also migrates — its in-tree `.claude-workflow/<ts>/` fallback is retired, replaced by an `XDG_STATE_HOME`-rooted path for symmetry with the backgrounded case.
- `settings.json` auto-approves writes to the new state root so interactive sessions don't prompt.

## Non-goals

- Migrating existing state under `~/.claude/workflows/`. Orphan the old directories in place; the user can `rm -rf` them when done inspecting history.
- Changing the `bg/localimplement/<id>` branch semantics, the `/ghimplement` issue/PR flow, the workflow ID scheme, the `state.json` shape, or the `/workflows` listing output format.
- Re-adding artifacts to the product branch.

## Why `XDG_STATE_HOME` and not `/tmp`

An earlier draft proposed `/tmp/claude-workflows/`. Rejected because `/tmp` is tmpfs on many Linux VMs (evaporates on reboot) and macOS's `periodic daily` sweeps `/private/tmp` entries older than 3 days. Workflow state is debugging-valuable days after a run — wrong tier.

`XDG_STATE_HOME` (defaults to `$HOME/.local/state`) is the XDG-spec-designated location for \"state of long-running processes, logs\" — exactly our case. Persists across reboots, survives in VM snapshots, not swept.

Why not `XDG_CONFIG_HOME` (the root the user first suggested): that's for stable configuration. Workflow logs and per-run artifacts aren't config.

### Path: sibling, not child of Claude's own state dir

Claude Code itself already writes to `~/.local/state/claude/` (observed: `locks/` subdirectory). Use a **sibling**, not a child:

- ✅ `${XDG_STATE_HOME:-$HOME/.local/state}/claude-workflows/` — our root.
- ❌ `${XDG_STATE_HOME:-$HOME/.local/state}/claude/workflows/` — under Claude's own state dir, which may attract future guardrail extensions.

## Current write map

All paths below currently live under `\$HOME/.claude/workflows/<WF_ID>/`:

| Path | Written by | Read by |
|---|---|---|
| `state.json` | `launch.sh` (init) · `set-stage.sh` · `finish.sh` | `workflows.sh` · `session-summary.sh` · `liveness.sh` |
| `log` | `launch.sh` (redirects pipeline stdout+stderr) | user · hooks |
| `pid` | `launch.sh` | debug |
| `finished` marker | `finish.sh` | `session-summary.sh` · liveness classifier |
| `acknowledged` marker | `session-summary.sh` · `workflows.sh --ack` | `workflows.sh` · `session-summary.sh` |
| `artifacts/plan.md` | `localimplement.sh` subagent | `localimplement.sh` subagent |
| `artifacts/review-code-holistic-<model>.md` | `localimplement.sh` subagent | `localimplement.sh` subagent |
| `artifacts/review-code-detail-<model>.md` | `localimplement.sh` subagent | `localimplement.sh` subagent |
| `artifacts/stream-*.jsonl` | `localimplement.sh` (via `tee` on stream-json output) | user · debug |

Target: identical shape, different root — `${XDG_STATE_HOME:-$HOME/.local/state}/claude-workflows/<WF_ID>/`.

## Constraints

- **Single source of truth.** Introduce one constant/variable for the root path used by every script. A shell snippet like `STATE_ROOT=\"${XDG_STATE_HOME:-$HOME/.local/state}/claude-workflows\"` duplicated verbatim at the top of each script is acceptable given the project's \"no new abstraction\" style — but the duplication must be kept in sync. The scatter of `\$HOME/.claude/workflows` literals across 5+ files is why this migration is invasive; fix it once.
- **Preserve charset validation on `WF_ID`.** `localimplement.sh` already validates `WF_ID` against `^[A-Za-z0-9._-]+\$` before interpolating it into a path. Keep that check at every site where the new root concatenates `\$WF_ID` (script-by-script audit).
- **`mkdir -p` the root before first write.** Unlike `~/.claude/`, `\$XDG_STATE_HOME/claude-workflows/` is not guaranteed to exist. Every script that writes under it should `mkdir -p` defensively (matches the existing pattern at `launch.sh:160`).
- **No data migration, no symlinks.** New root is created fresh; existing `~/.claude/workflows/*` is orphaned. No bridging.

## Acceptance criteria

1. A completed `/localimplement` run writes `plan.md`, both review files, all `stream-*.jsonl`, `state.json`, `log`, `pid`, and markers entirely under `${XDG_STATE_HOME:-$HOME/.local/state}/claude-workflows/<WF_ID>/`. Nothing new appears under `~/.claude/workflows/`.
2. A completed `/ghimplement` run writes `state.json`, `log`, `pid`, and markers under `${XDG_STATE_HOME:-$HOME/.local/state}/claude-workflows/<WF_ID>/`. Nothing new under `~/.claude/workflows/`.
3. `/workflows`, `/workflows --here`, `/workflows --ack <id>`, and `/workflows --ack-all` all read and write under the new state root.
4. The `SessionStart` / `UserPromptSubmit` hook (`session-summary.sh`) reads the new state root and surfaces running + just-finished workflows correctly.
5. Direct-CLI `skills/localimplement/localimplement.sh \"<instructions>\"` creates its session directory under the new state root (exact subpath scheme is implementation choice — e.g. `<ts>-direct-<rand>` — documented in-script). The in-tree `.claude-workflow/<ts>/` directory is NOT created.
6. `settings.json` has entries under `permissions.allow` that auto-approve writes to `${XDG_STATE_HOME:-$HOME/.local/state}/claude-workflows/**` (write-related tools and matching `Bash` patterns). Since shell-based permission matchers may not expand `${XDG_STATE_HOME:-…}` the way a script does, prefer a hardcoded `$HOME/.local/state/claude-workflows/**` pattern in the allowlist — it matches the 99%-case (no override) exactly, and users who set `XDG_STATE_HOME` explicitly are expected to also extend their allowlist manually.
7. User-facing docs point at the new paths: `README.md`, `skills/localimplement/SKILL.md`, `skills/ghimplement/SKILL.md`, `skills/workflows/SKILL.md`. The launcher's stdout banner (\"workflow id / workdir / log / state file / pid\") emits new-root paths.
8. No-scaffolding-in-repo defense clauses in `localimplement.sh`'s and `ghimplement.sh`'s implement-stage prompts remain (they're vestigial after this change but still defend against an over-eager subagent creating `.claude-workflow/` or `plan.md` in the product tree).

## Files to touch

- `skills/_bg/launch.sh` — `STATE_ROOT` (line ~158), launcher stdout banner.
- `skills/_bg/finish.sh` — `STATE_DIR` (line ~21).
- `skills/_bg/set-stage.sh` — `STATE_FILE` (line ~19).
- `skills/_bg/session-summary.sh` — `STATE_ROOT` (line ~13) and header comment.
- `skills/workflows/workflows.sh` — `STATE_ROOT` (line ~10) and header comment.
- `skills/localimplement/localimplement.sh` — collapse the `WF_ID`-set vs. unset branches into a unified new-root path; update comments at lines ~14, ~26, ~102, ~104, ~107, ~112–117, ~271.
- `skills/localimplement/SKILL.md` — artifact-path references at lines ~3, ~12, ~21–24.
- `skills/ghimplement/SKILL.md` — state-path reference at line ~12.
- `skills/workflows/SKILL.md` — description at lines ~3, ~8.
- `README.md` — per-workflow-state references at lines ~52, ~62.
- `settings.json` — append to `permissions.allow`.

## Open questions

- **`mkdir -p` ownership.** Who creates `${XDG_STATE_HOME:-$HOME/.local/state}/claude-workflows/` first? Having every script `mkdir -p` independently is simplest; the alternative (a dedicated bootstrap) is more indirection than this codebase otherwise uses.
- **Permission-entry granularity.** `Write($HOME/.local/state/claude-workflows/**)` alone, or add matching `Bash` patterns (`Bash(tee $HOME/.local/state/claude-workflows/*)`, etc.)? Prefer the minimum needed to silence prompts on the real flows; add more as observed gaps appear.

xbrianh:
## Summary

Closes #17.

- Relocate every workflow runtime path (state.json, log, pid, markers, artifacts, stream jsonl) from `~/.claude/workflows/<id>/` to `${XDG_STATE_HOME:-$HOME/.local/state}/claude-workflows/<id>/`, escaping Claude Code's sensitive-file guardrail that was probabilistically breaking `/localimplement` artifact writes under `--permission-mode bypassPermissions`.
- Collapse `localimplement.sh`'s WF_ID-set vs. unset branches into a single path: direct-CLI runs now synthesize `<ts>-direct-<rand>` session ids under the same XDG root instead of writing to an in-tree `.claude-workflow/<ts>/` dir.
- `settings.json`: auto-approve `Write`/`Edit`/`Bash` writes under `$HOME/.local/state/claude-workflows/**` so interactive sessions don't prompt. Uses a hardcoded `$HOME` path (not `${XDG_STATE_HOME:-...}`) per issue guidance — shell-based matchers don't expand parameter defaults.

Existing state under `~/.claude/workflows/` is intentionally **not** migrated — orphaned in place per the issue's non-goals.

## Files changed

- `skills/_bg/launch.sh`, `finish.sh`, `set-stage.sh`, `session-summary.sh` — new STATE_ROOT.
- `skills/workflows/workflows.sh` — new STATE_ROOT.
- `skills/localimplement/localimplement.sh` — unified session-dir logic.
- `skills/localimplement/SKILL.md`, `skills/ghimplement/SKILL.md`, `skills/workflows/SKILL.md`, `README.md` — doc path updates.
- `settings.json` — permission allowlist entries.

## Test plan

- [ ] Run `/localimplement <task>` end-to-end; confirm `plan.md`, both review files, `stream-*.jsonl`, `state.json`, `log`, `pid`, and `finished`/`acknowledged` markers all land under `$HOME/.local/state/claude-workflows/<id>/` and nothing new appears under `~/.claude/workflows/`.
- [ ] Run `/ghimplement <task>` end-to-end; confirm state/log/pid/markers land under the new root and nothing new under `~/.claude/workflows/`.
- [ ] Run `/workflows`, `/workflows --here`, `/workflows --ack <id>`, `/workflows --ack-all` — all read/write the new root.
- [ ] Confirm `SessionStart`/`UserPromptSubmit` hook surfaces running + just-finished workflows correctly against the new root.
- [ ] Run `skills/localimplement/localimplement.sh "<instructions>"` directly (no WF_ID); confirm session dir at `$HOME/.local/state/claude-workflows/<ts>-direct-<rand>/artifacts/` and that no `.claude-workflow/` dir is created in the repo tree.
- [ ] Confirm no permission prompts fire on artifact writes during a fresh `/localimplement` run.

### DIFF
--- a/settings.json
+++ b/settings.json
@@ -4,12 +4,12 @@
       "Bash(gh *)",
       "Bash(grep *)",
       "Bash(find *)",
-      "Write($HOME/.local/state/claude-workflows/**)",
-      "Edit($HOME/.local/state/claude-workflows/**)",
-      "Bash(mkdir -p $HOME/.local/state/claude-workflows/**)",
-      "Bash(tee $HOME/.local/state/claude-workflows/**)",
-      "Bash(cat > $HOME/.local/state/claude-workflows/**)",
-      "Bash(cat >> $HOME/.local/state/claude-workflows/**)"
+      "Write(~/.local/state/claude-workflows/**)",
+      "Edit(~/.local/state/claude-workflows/**)",
+      "Bash(mkdir -p ~/.local/state/claude-workflows/**)",
+      "Bash(tee ~/.local/state/claude-workflows/**)",
+      "Bash(cat > ~/.local/state/claude-workflows/**)",
+      "Bash(cat >> ~/.local/state/claude-workflows/**)"
     ],
     "defaultMode": "auto"
   },
--- a/settings.json
+++ b/settings.json
@@ -3,7 +3,13 @@
     "allow": [
       "Bash(gh *)",
       "Bash(grep *)",
-      "Bash(find *)"
+      "Bash(find *)",
+      "Write($HOME/.local/state/claude-workflows/**)",
+      "Edit($HOME/.local/state/claude-workflows/**)",
+      "Bash(mkdir -p $HOME/.local/state/claude-workflows/**)",
+      "Bash(tee $HOME/.local/state/claude-workflows/**)",
+      "Bash(cat > $HOME/.local/state/claude-workflows/**)",
+      "Bash(cat >> $HOME/.local/state/claude-workflows/**)"
     ],
     "defaultMode": "auto"
   },
--- a/skills/localimplement/SKILL.md
+++ b/skills/localimplement/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: localimplement
-description: Run the end-to-end plan → implement → review-code → address-code workflow in the background by invoking ~/.claude/skills/_bg/launch.sh. Plan and code reviews land in `~/.claude/workflows/<workflow-id>/artifacts/` alongside the run log (kept off the product branch); the code review uses two different models in parallel. The launcher returns immediately; you'll be notified when the pipeline finishes.
+description: Run the end-to-end plan → implement → review-code → address-code workflow in the background by invoking ~/.claude/skills/_bg/launch.sh. Plan and code reviews land in `${XDG_STATE_HOME:-$HOME/.local/state}/claude-workflows/<workflow-id>/artifacts/` alongside the run log (kept off the product branch); the code review uses two different models in parallel. The launcher returns immediately; you'll be notified when the pipeline finishes.
 argument-hint: [-a <model>] [-b <model>] <instructions>
 allowed-tools: Bash(~/.claude/skills/_bg/launch.sh:*)
 ---
@@ -9,7 +9,7 @@ You are running the `localimplement` workflow **in the background**. The skill i
 
 1. Creates an isolated git worktree of the current project on a fresh branch named `bg/localimplement/<workflow-id>` (or `cp -a` copies the tree for non-git projects).
 2. Spawns the real pipeline (`~/.claude/skills/localimplement/localimplement.sh`) detached from this session — it survives Ctrl-C, shell exit, and Claude Code quitting.
-3. Records per-workflow state under `~/.claude/workflows/<workflow-id>/` (`state.json`, combined `log`, markers).
+3. Records per-workflow state under `${XDG_STATE_HOME:-$HOME/.local/state}/claude-workflows/<workflow-id>/` (`state.json`, combined `log`, markers).
 4. Returns within ~1s.
 
 A `SessionStart` / `UserPromptSubmit` hook notifies a future Claude session for this project when the workflow finishes.
@@ -18,10 +18,10 @@ A `SessionStart` / `UserPromptSubmit` hook notifies a future Claude session for 
 
 Plan and code-review artifacts live outside the product branch — they are scaffolding, not product. Point the user at:
 
-- `~/.claude/workflows/<workflow-id>/artifacts/plan.md` — the implementation plan.
-- `~/.claude/workflows/<workflow-id>/artifacts/review-code-holistic-<model>.md` and `review-code-detail-<model>.md` — the two parallel code reviews.
-- `~/.claude/workflows/<workflow-id>/log` — combined stdout/stderr of the pipeline.
-- `~/.claude/workflows/<workflow-id>/state.json` — workflow status, exit code, workdir path, branch name.
+- `${XDG_STATE_HOME:-$HOME/.local/state}/claude-workflows/<workflow-id>/artifacts/plan.md` — the implementation plan.
+- `${XDG_STATE_HOME:-$HOME/.local/state}/claude-workflows/<workflow-id>/artifacts/review-code-holistic-<model>.md` and `review-code-detail-<model>.md` — the two parallel code reviews.
+- `${XDG_STATE_HOME:-$HOME/.local/state}/claude-workflows/<workflow-id>/log` — combined stdout/stderr of the pipeline.
+- `${XDG_STATE_HOME:-$HOME/.local/state}/claude-workflows/<workflow-id>/state.json` — workflow status, exit code, workdir path, branch name.
 - `bg/localimplement/<workflow-id>` — durable branch with **only** the code changes (no scaffolding). From the main working tree: `git checkout bg/localimplement/<workflow-id>` to inspect, merge, or discard. A squash-merge pulls in product code cleanly.
 
 Commits on the branch, in order: implementation → "Address review feedback" (absent if reviewers found nothing).
--- a/skills/localimplement/SKILL.md
+++ b/skills/localimplement/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: localimplement
-description: Run the end-to-end plan → implement → review-code → address-code workflow in the background by invoking ~/.claude/skills/_bg/launch.sh. Plan and code reviews land in `${XDG_STATE_HOME:-$HOME/.local/state}/claude-workflows/<workflow-id>/artifacts/` alongside the run log (kept off the product branch); the code review uses two different models in parallel. The launcher returns immediately; you'll be notified when the pipeline finishes.
+description: Run the end-to-end plan → implement → review-code → address-code workflow in the background by invoking ~/.claude/skills/_bg/launch.sh. Plan and code reviews land in `~/.local/state/claude-workflows/<workflow-id>/artifacts/` alongside the run log (kept off the product branch); the code review uses two different models in parallel. The launcher returns immediately; you'll be notified when the pipeline finishes.
 argument-hint: [-a <model>] [-b <model>] <instructions>
 allowed-tools: Bash(~/.claude/skills/_bg/launch.sh:*)
 ---
@@ -9,7 +9,7 @@ You are running the `localimplement` workflow **in the background**. The skill i
 
 1. Creates an isolated git worktree of the current project on a fresh branch named `bg/localimplement/<workflow-id>` (or `cp -a` copies the tree for non-git projects).
 2. Spawns the real pipeline (`~/.claude/skills/localimplement/localimplement.sh`) detached from this session — it survives Ctrl-C, shell exit, and Claude Code quitting.
-3. Records per-workflow state under `${XDG_STATE_HOME:-$HOME/.local/state}/claude-workflows/<workflow-id>/` (`state.json`, combined `log`, markers).
+3. Records per-workflow state under `~/.local/state/claude-workflows/<workflow-id>/` (or `$XDG_STATE_HOME/claude-workflows/<workflow-id>/` if `XDG_STATE_HOME` is set) — `state.json`, combined `log`, markers.
 4. Returns within ~1s.
 
 A `SessionStart` / `UserPromptSubmit` hook notifies a future Claude session for this project when the workflow finishes.
@@ -18,10 +18,10 @@ A `SessionStart` / `UserPromptSubmit` hook notifies a future Claude session for 
 
 Plan and code-review artifacts live outside the product branch — they are scaffolding, not product. Point the user at:
 
-- `${XDG_STATE_HOME:-$HOME/.local/state}/claude-workflows/<workflow-id>/artifacts/plan.md` — the implementation plan.
-- `${XDG_STATE_HOME:-$HOME/.local/state}/claude-workflows/<workflow-id>/artifacts/review-code-holistic-<model>.md` and `review-code-detail-<model>.md` — the two parallel code reviews.
-- `${XDG_STATE_HOME:-$HOME/.local/state}/claude-workflows/<workflow-id>/log` — combined stdout/stderr of the pipeline.
-- `${XDG_STATE_HOME:-$HOME/.local/state}/claude-workflows/<workflow-id>/state.json` — workflow status, exit code, workdir path, branch name.
+- `~/.local/state/claude-workflows/<workflow-id>/artifacts/plan.md` — the implementation plan.
+- `~/.local/state/claude-workflows/<workflow-id>/artifacts/review-code-holistic-<model>.md` and `review-code-detail-<model>.md` — the two parallel code reviews.
+- `~/.local/state/claude-workflows/<workflow-id>/log` — combined stdout/stderr of the pipeline.
+- `~/.local/state/claude-workflows/<workflow-id>/state.json` — workflow status, exit code, workdir path, branch name.
 - `bg/localimplement/<workflow-id>` — durable branch with **only** the code changes (no scaffolding). From the main working tree: `git checkout bg/localimplement/<workflow-id>` to inspect, merge, or discard. A squash-merge pulls in product code cleanly.
 
 Commits on the branch, in order: implementation → "Address review feedback" (absent if reviewers found nothing).
--- a/skills/localimplement/localimplement.sh
+++ b/skills/localimplement/localimplement.sh
@@ -11,7 +11,8 @@ trap 'trap - INT TERM; kill -- -$$ 2>/dev/null; exit 130' INT TERM
 die() { echo "error: $*" >&2; exit 1; }
 
 # Stage helper: only meaningful when this script runs under the _bg launcher
-# (which exports WF_ID and creates ~/.claude/workflows/<WF_ID>/state.json).
+# (which exports WF_ID and creates the workflow's state.json under
+# ${XDG_STATE_HOME:-$HOME/.local/state}/claude-workflows/<WF_ID>/).
 # A direct CLI invocation has no WF_ID and set_stage no-ops.
 SET_STAGE_SH="$HOME/.claude/skills/_bg/set-stage.sh"
 set_stage() {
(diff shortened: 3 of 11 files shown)


settings.json
```diff
@@ -3,7 +3,13 @@
     "allow": [
       "Bash(gh *)",
       "Bash(grep *)",
-      "Bash(find *)"
+      "Bash(find *)",
+      "Write($HOME/.local/state/claude-workflows/**)",
+      "Edit($HOME/.local/state/claude-workflows/**)",
+      "Bash(mkdir -p $HOME/.local/state/claude-workflows/**)",
```
xbrianh: **Severity:** major

`Bash(...)` permission matchers compare against the command string as the tool sees it. A subagent issuing `mkdir -p /Users/xath/.local/state/claude-workflows/foo` won't match the literal `$HOME` in this pattern — by the time a `Bash` tool call reaches the permission check, `$HOME` would have to appear *literally* in the command (which never happens in practice). Same issue for the `tee` / `cat >` / `cat >>` entries below.

**Fix:** swap `$HOME` for `~` — which Claude Code does expand in matcher patterns (cf. `Bash(~/.claude/skills/_bg/launch.sh:*)` elsewhere in the repo). Even with that, these Bash matchers only cover a few specific shell verbs; they won't help a subagent that uses redirection (`>file`) or a different tool. The real coverage comes from the `Write(...)` and `Edit(...)` entries above — the Bash entries are belt-and-braces and arguably droppable if they're going to be this fragile.

settings.json
```diff
@@ -3,7 +3,13 @@
     "allow": [
       "Bash(gh *)",
       "Bash(grep *)",
-      "Bash(find *)"
+      "Bash(find *)",
+      "Write($HOME/.local/state/claude-workflows/**)",
```
xbrianh: **Severity:** major

Same `$HOME` concern: verify that Claude Code expands `$HOME` in `Write(...)` / `Edit(...)` path matchers against absolute paths like `/Users/xath/.local/state/claude-workflows/foo/plan.md`. `~` is the safer character to use here — it's consistently expanded across the permission matcher surface. If these two entries don't match, the whole acceptance criterion 6 ("no permission prompts on artifact writes") silently fails and the guardrail fragility the PR sets out to fix returns.

skills/localimplement/localimplement.sh
```diff
@@ -98,23 +100,26 @@ done
...
   [[ "$WF_ID" =~ ^[A-Za-z0-9._-]+$ ]] || die "invalid WF_ID: $WF_ID"
-  SESSION_DIR="$HOME/.claude/workflows/$WF_ID/artifacts"
+  SESSION_ID="$WF_ID"
 else
-  SESSION_DIR=".claude-workflow/$TS"
+  RAND=$(LC_ALL=C tr -dc 'a-f0-9' </dev/urandom 2>/dev/null | head -c 6 || true)
+  [[ -n "$RAND" ]] || RAND="xxxxxx"
+  SESSION_ID="$TS-direct-$RAND"
 fi
+SESSION_DIR="$STATE_ROOT/$SESSION_ID/artifacts"
```
xbrianh: **Severity:** minor

Direct-CLI runs now write to `$STATE_ROOT/<ts>-direct-<rand>/artifacts/` — but no `state.json` is ever created there, and no `finished` / `acknowledged` marker is ever dropped. Consequence: the 14-day prune in `session-summary.sh:189` (`find -maxdepth 2 -name acknowledged -mtime +14`) never matches these, so direct-CLI session dirs accumulate forever in the state root.

Two cheap fixes:
- Nest direct-CLI runs under a dedicated subdir (e.g. `$STATE_ROOT/direct/<ts>-<rand>/artifacts/`) so they're visually separated from real workflow state dirs and can be pruned by a different heuristic.
- Or, extend the `session-summary.sh` prune to also sweep artifact-only dirs older than N days.

The old in-tree `.claude-workflow/<ts>/` had the same issue but was at least visible in `git status` as a nudge to clean up; the XDG root is out of sight.

skills/localimplement/SKILL.md
```diff
@@ -1,6 +1,6 @@
 ---
 name: localimplement
-description: Run the end-to-end plan → implement → review-code → address-code workflow in the background by invoking ~/.claude/skills/_bg/launch.sh. Plan and code reviews land in `~/.claude/workflows/<workflow-id>/artifacts/` alongside the run log (kept off the product branch); the code review uses two different models in parallel. The launcher returns immediately; you'll be notified when the pipeline finishes.
+description: Run the end-to-end plan → implement → review-code → address-code workflow in the background by invoking ~/.claude/skills/_bg/launch.sh. Plan and code reviews land in `${XDG_STATE_HOME:-$HOME/.local/state}/claude-workflows/<workflow-id>/artifacts/` alongside the run log (kept off the product branch); the code review uses two different models in parallel. The launcher returns immediately; you'll be notified when the pipeline finishes.
```
xbrianh: **Severity:** nit

The `${XDG_STATE_HOME:-$HOME/.local/state}/claude-workflows/<workflow-id>/artifacts/` form shows up multiple times in user-facing skill prose (description, body). It's shell-accurate but reads awkwardly — most users never set `XDG_STATE_HOME`. Consider `~/.local/state/claude-workflows/<workflow-id>/artifacts/` in prose with a single parenthetical disclaimer somewhere ("or `$XDG_STATE_HOME/claude-workflows/` if set"), and keep the `${VAR:-default}` form only in the scripts where it has runtime semantics.

skills/localimplement/localimplement.sh
```diff
@@ -98,23 +100,26 @@ done
...
+# WF_ID is interpolated into a filesystem path, so validate it against a
+# conservative charset to prevent path traversal (e.g. "../") or embedded
+# slashes when it's set externally. The _bg launcher produces IDs matching
+# ^[a-z0-9-]+$, so this pattern is a strict superset.
+STATE_ROOT="${XDG_STATE_HOME:-$HOME/.local/state}/claude-workflows"
 TS=$(date +%Y%m%d-%H%M%S)
 if [[ -n "${WF_ID:-}" ]]; then
   [[ "$WF_ID" =~ ^[A-Za-z0-9._-]+$ ]] || die "invalid WF_ID: $WF_ID"
-  SESSION_DIR="$HOME/.claude/workflows/$WF_ID/artifacts"
+  SESSION_ID="$WF_ID"
```
xbrianh: **Severity:** nit

Minor consistency: when `WF_ID` is set the session id reuses it verbatim; when it isn't, the synthesized id is `<ts>-direct-<rand>`. The `<ts>` prefix on the direct-CLI path is nice for chronological sorting, but `WF_ID`-backed sessions have no timestamp in the name (the slug+hex form). Not incorrect — just worth noting that listing `$STATE_ROOT/` in sorted order won't group _bg-launcher runs chronologically. Cheap fix if you care: add a `started_at` field to the (future) direct-CLI state file, or drop the `<ts>` from the direct path for symmetry.

xbrianh (commented):
The migration is mechanically clean and touches every file called out in the issue. Three concerns worth addressing before merge:

1. **Bash permission matchers probably won't match.** The `Bash(mkdir -p $HOME/...)` / `Bash(tee $HOME/...)` / `Bash(cat > $HOME/...)` entries compare against the literal command string, but by the time a Bash tool call is evaluated the shell has resolved `$HOME` to an absolute path — and no subagent is going to write the literal string `$HOME` into a command. `~` is the idiomatic character Claude Code expands in matcher patterns (see how the rest of the codebase uses `Bash(~/.claude/skills/_bg/launch.sh:*)`); same concern applies to the `Write(...)` / `Edit(...)` path matchers.

2. **Direct-CLI sessions accumulate forever.** The new code writes `$STATE_ROOT/<ts>-direct-<rand>/artifacts/` but never drops a `state.json` or `acknowledged` marker there, so the 14-day prune in `session-summary.sh` (which finds `acknowledged` via `find -maxdepth 2`) will never clean them up. Worth either nesting direct-CLI runs under a separate subdir (e.g. `$STATE_ROOT/direct/`) or extending the prune to match artifacts-only dirs.

3. **User-facing SKILL.md descriptions now read as raw shell syntax.** `${XDG_STATE_HOME:-$HOME/.local/state}/claude-workflows/<id>/` is accurate but awkward. Most users never set `XDG_STATE_HOME` — a simpler `~/.local/state/claude-workflows/<id>/` in prose (with the shell-syntax form kept in the scripts where it actually does something) would be easier to read.

All three are addressable without changing the core migration. Nothing that warrants a re-architecture.

settings.json
xbrianh: Fixed in e95c097 — swapped `$HOME` → `~` across all six `Bash(...)` entries. Kept them in the allowlist alongside the `Write(...)` / `Edit(...)` entries as belt-and-braces; agreed they're redundant in most flows but cheap to keep now that the pattern is right.

settings.json
xbrianh: Fixed in e95c097 — swapped `$HOME` → `~` in the `Write(...)` and `Edit(...)` matchers as well. `~` is expanded in Claude Code's path matchers against absolute paths, so these should now cover writes to `/Users/xath/.local/state/claude-workflows/<id>/...` and keep acceptance criterion 6 intact.

skills/localimplement/localimplement.sh
xbrianh: Fixed in e95c097. Took the first suggestion — direct-CLI runs now nest under `$STATE_ROOT/direct/<ts>-<rand>/` (instead of sitting as siblings of real workflow dirs), and `session-summary.sh` grew a second prune pass that sweeps `$STATE_ROOT/direct/*` dirs older than 14 days by directory mtime. The sweep is guarded by a `case` pattern so the `rm -rf` can't escape the state root.

skills/localimplement/SKILL.md
xbrianh: Fixed in e95c097 — user-facing prose across `localimplement/SKILL.md`, `ghimplement/SKILL.md`, `workflows/SKILL.md`, and `README.md` now reads `~/.local/state/claude-workflows/<id>/...` with a single parenthetical disclaimer on first mention in each doc. Shell-syntactic `${XDG_STATE_HOME:-$HOME/.local/state}` form is preserved in the scripts (launch.sh, workflows.sh, finish.sh, set-stage.sh, session-summary.sh, localimplement.sh) where it has runtime semantics.

skills/localimplement/localimplement.sh
xbrianh: Acknowledged but not changed. With the restructure from the sibling comment (direct-CLI runs now live under `$STATE_ROOT/direct/`), `_bg`-launcher runs and direct-CLI runs are no longer peers in the same directory, so the asymmetric naming doesn't hurt sort order for any single `ls`. If a `started_at` field shows up in a future direct-CLI state file we can revisit; for now keeping the `<ts>-<rand>` form under `direct/` since the `<ts>` prefix is still useful for sorting within that subdir.

PR #20, opened 2026-04-21
Port localimplement pipeline from bash to Python

### LINKED ISSUE #19: Port localimplement pipeline from bash to Python
## Context

`skills/localimplement/localimplement.sh` (350 lines of bash) is the pipeline script behind the `/localimplement` background workflow: plan → implement → triple-review → address-code. It's grown enough that several bits are awkward in bash:

- `log_stream` is a tee-to-jq-to-stderr pipeline whose previous `tee >(jq ...)` form silently failed under the `_bg` launcher because process-substitution stderr didn't reach the workflow log; the current synchronous form works but is delicate.
- The triple-reviewer fan-out uses subshell wrappers + `kill -0` polling + `wait` to harvest exit codes, with comments calling out `$!` capture pitfalls and PIPESTATUS gymnastics.
- Argument parsing is six getopts flags with hand-rolled regex validation.
- JSON munging for `set_stage` sub_stage updates leans on `jq -cn` from inside bash.

Python gives us first-class `subprocess`, `json`, `argparse`, `threading`, `signal` — the same idioms the existing `skills/workflows/workflows.py` already uses successfully in this repo. Porting reduces the foot-guns without changing behavior.

## Approach

Rewrite the pipeline in Python (stdlib only — same surface as `workflows.py`), preserve the exact externally-observable contract, and update the launcher to discover `.py` pipelines:

- **Same external contract.** Same CLI flags (`-p -i -x -a -b -c`), same defaults (`sonnet`), same artifact filenames in the same `$STATE_ROOT/$WF_ID/artifacts/` (or `$STATE_ROOT/direct/<ts>-<rand>/artifacts/`) directory, same stage names sent to `set-stage.sh` (`plan` / `implement` / `review-code` / `address-code`), same sub_stage JSON shape (`{holistic, detail, scope}` keyed by lens, value `\"<status> (<model>)\"`).
- **Stdlib only.** `argparse`, `subprocess`, `threading`, `signal`, `json`, `re`, `os`, `pathlib`, `time`, `datetime` — matches `workflows.py` and avoids introducing a venv requirement.
- **Threads, not asyncio**, for the triple-reviewer fan-out: each reviewer is dominated by subprocess I/O (reading stream-json from `claude -p`); threads map 1:1 to the existing bash subshells and let us reuse the same polling-emit-sub_stage pattern without an event loop.
- **Keep helper bash scripts.** `_bg/launch.sh`, `_bg/finish.sh`, `_bg/set-stage.sh` stay bash — they're small, stable, and shared with `ghimplement`. The pipeline shells out to `set-stage.sh` exactly as today.
- **Launcher discovery** (`_bg/launch.sh`) gains a `.py`-then-`.sh` fallback so `ghimplement.sh` continues to work unchanged. This is a one-line surgical change, not a launcher rewrite.

Why this over alternatives:

- **Versus a partial port (e.g. only `log_stream` to a Python helper):** the bash↔Python boundary would land in the middle of the parallel-reviewer dance, which is the most subtle part. A full port is simpler than a hybrid.
- **Versus also porting `ghimplement.sh` in the same PR:** scope discipline. `ghimplement` has its own complexity (issue creation, PR push, Copilot review wait); bundling doubles the diff and makes review harder.
- **Versus introducing a third-party dep (e.g. `rich`, `click`):** adds a venv requirement that would break the launcher's bare `command -v claude` / `command -v jq` style of dependency probing. Stdlib keeps the install footprint identical.

## Tasks

- [ ] Create `skills/localimplement/localimplement.py` with `#!/usr/bin/env python` shebang, executable bit set, and a module docstring mirroring the bash header (role under the `_bg` launcher; XDG state layout; why artifacts live outside the product branch).
- [ ] Port argument parsing to `argparse`: `-p/--plan-model`, `-i/--impl-model`, `-x/--address-model`, `-a/--holistic-model`, `-b/--detail-model`, `-c/--scope-model` (each defaulting to `sonnet`), positional `instructions` (joined). Reproduce the `^[A-Za-z0-9._-]+$` model-alias validation.
- [ ] Port `die()` and the SIGINT/SIGTERM trap: install signal handlers that terminate all tracked child `subprocess.Popen` instances (`terminate()` then `kill()` after a short grace period) before exiting 130. Spawn children with `start_new_session=False` so they stay in our process group; track them in a module-level list under a lock.
- [ ] Port `set_stage()` as a Python helper that shells out to `~/.claude/skills/_bg/set-stage.sh` with `WF_ID` (no-op when `WF_ID` unset or the helper is missing/non-executable; sub_stage payload built with `json.dumps`, never raises).
- [ ] Port session-dir resolution: validate `WF_ID` against `^[A-Za-z0-9._-]+$` (`die` on mismatch), compute `SESSION_DIR` as `$STATE_ROOT/$WF_ID/artifacts` when set, otherwise `$STATE_ROOT/direct/<YYYYMMDD-HHMMSS>-<rand6hex>/artifacts`. `mkdir -p` it. Print `==> session: <path>` to match current output.
- [ ] Port `log_stream()`: take `(label, raw_path)`, return a callable that consumes a `Popen.stdout` line iterator. For each line, write raw to `raw_path` (open append-binary, flush per line), then `try: json.loads(line)` — on failure, skip silently (preserve the bash `|| true` semantic so a truncated JSON line cannot abort the stage). Format human-readable trace to `sys.stderr` per event type:
  - `system` + `subtype=init` → `[<label>] init session=... model=... cwd=...`
  - `assistant` text → `[<label>] text: <truncated 200>`
  - `assistant` tool_use → `[<label>] tool: <name> <first non-empty of file_path/command/pattern/url/output_file truncated 200>`
  - `user` tool_result → `[<label>] result[ ERROR]: <body truncated 200>`
  - `result` → `[<label>] final: subtype=... turns=... cost=...`
  Replace `\n` with spaces in all displayed strings (parity with the existing `gsub`).
- [ ] Port the **plan** stage: emit `==> [1/4] planning (model: <m>) -> <plan.md>`, run `claude -p --model <m> --permission-mode bypassPermissions --output-format stream-json --verbose <prompt>` via `subprocess.Popen` with stdout piped through `log_stream`, raise via `die()` if `plan.md` is empty after exit. Use the existing prompt verbatim (Context / Approach / Tasks / Open questions structure).
- [ ] Port the **implement** stage: capture `PRE_HEAD` via `git rev-parse HEAD` when in a git repo, otherwise touch a `.pre-impl` sentinel inside `SESSION_DIR`. Compose `IMPL_COMMIT_INSTR` exactly as today (referencing `plan.md` by basename, not absolute path; forbidding meta/scaffolding files in the repo). Run `claude -p` with the implement prompt. Verify changes occurred:
  - In git: `git rev-parse HEAD` differs from `PRE_HEAD` **or** `git status --porcelain` is non-empty.
  - Outside git: at least one file under cwd has mtime newer than the sentinel, excluding `$SESSION_DIR/*` and `.git/*`.
  Otherwise `die(\"implementation stage produced no changes; aborting\")`.
- [ ] Port `run_review()` as a Python function that builds the same prompt template (model-namespaced output file, lens prose, `where_field` placeholder, `# Review ($model)` / `## Summary` / `## Findings` structure, no-issues escape clause, \"do NOT make any code changes\" guardrail) and runs `claude -p` via the same `Popen` + `log_stream` plumbing.
- [ ] Port `run_triple_review()` to threads: spawn three `threading.Thread` workers (one per lens), each running `run_review()` and capturing exceptions. Emit the initial `set_stage review-code {holistic: \"running (...)\", ...}` sub_stage. Loop with a 2s sleep, polling each thread via `is_alive()`; whenever one transitions to done, re-emit the sub_stage with that lens flipped to `done`. After all three join, raise `die(\"one or more reviews failed\")` if any worker recorded an exception, then `die` if any of the three output files is empty.
- [ ] Port the **address-code** stage: emit `==> [4/4] addressing code reviews (model: <m>)`, build `ADDRESS_COMMIT_INSTR` (only in git; \"Address review feedback\" title; no push), run `claude -p` with the existing three-reviews prompt referencing the holistic / detail / scope output files by their absolute paths and model names, log to `stream-address.jsonl`.
- [ ] Update `skills/_bg/launch.sh` line 71 (`PIPELINE=\"$HOME/.claude/skills/$KIND/$KIND.sh\"`) to discover `.py` first then fall back to `.sh`:
  ```
  for ext in py sh; do
      candidate=\"$HOME/.claude/skills/$KIND/$KIND.$ext\"
      if [[ -x \"$candidate\" ]]; then PIPELINE=\"$candidate\"; break; fi
  done
  [[ -n \"${PIPELINE:-}\" ]] || die \"no executable pipeline at $HOME/.claude/skills/$KIND/$KIND.{py,sh}\"
  ```
  Keep `ghimplement.sh` untouched — it'll be picked up by the `.sh` fallback.
- [ ] Delete `skills/localimplement/localimplement.sh` once `localimplement.py` reaches parity. `scripts/sync.sh push` will mirror the deletion (under the 5-file `DELETE_THRESHOLD`, no `--force` needed). Git history preserves the bash version.
- [ ] Manual verification once the rewrite is in place:
  - `~/.claude/skills/_bg/launch.sh localimplement \"add a comment to README.md\"` — confirm workflow id printed, `state.json` `pid` populated.
  - Tail `~/.local/state/claude-workflows/<id>/log` — confirm `[plan] init …`, `[implement] tool: Edit …`, `[review-code:sonnet] …` × 3, `[address-code] …` traces appear (proving `log_stream` parity).
  - `cat ~/.local/state/claude-workflows/<id>/state.json` mid-run — confirm `stage` flips through `plan` / `implement` / `review-code` / `address-code` and `sub_stage` shows `{holistic, detail, scope}` flipping `running` → `done`.
  - On completion, confirm all four artifacts (`plan.md`, three `review-code-*.md`) are non-empty and the branch contains both an implementation commit and an `Address review feedback` commit.
  - Ctrl-C a fresh direct invocation (`./skills/localimplement/localimplement.py \"…\"`) and verify no orphan `claude` processes survive (`pgrep -af claude`).

## Open questions

- **`python` interpreter availability.** Bash is universally present; Python is one more probe. The launcher already requires `claude` and `jq`, and `workflows.py` already requires `python` — so this is a consistent stretch, but worth confirming we don't want a `command -v python` probe inside the launcher's `KIND==localimplement` path.
- **`.py`-then-`.sh` fallback in the launcher vs. KIND-specific lookup.** Generic fallback is one fewer hard-coded list to keep in sync; KIND-specific lookup is more explicit. Recommendation: generic, since it's a 4-line shell tweak and matches how the launcher already infers the pipeline path from `$KIND`.
- **Delete `localimplement.sh` in the same PR or leave it as a one-cycle deprecated copy?** Leaving both alive risks drift and confuses `sync.sh`. Recommendation: delete in the same PR; rollback is `git revert`.
- **Should we also port `ghimplement.sh`?** Out of scope for this issue — the two pipelines share no code and would double the diff. If this port goes well, file a follow-up.
- **Threading vs asyncio for the triple-reviewer fan-out.** Threading is simpler, maps 1:1 to the current bash subshells, and matches the polling pattern. asyncio would be marginally cleaner for the wait-loop but adds an event loop just for three subprocesses. Recommendation: threading.

xbrianh:
## Summary

- Rewrites `skills/localimplement/localimplement.sh` as `localimplement.py` (stdlib only) to remove the delicate bash idioms: `tee|jq` log streaming, subshell + `kill -0` + PIPESTATUS for the triple-reviewer fan-out, and `jq`-from-bash sub_stage JSON munging.
- Preserves the external contract: same flags (`-p/-i/-x/-a/-b/-c`), `sonnet` defaults, artifact filenames under `$STATE_ROOT/$WF_ID/artifacts/`, `set-stage.sh` stage names, and `{holistic, detail, scope}` sub_stage shape.
- `skills/_bg/launch.sh` now discovers `.py` then `.sh`, so `ghimplement.sh` keeps working via the fallback; the bash pipeline is deleted in the same commit.

Closes #19

## Test plan

- [ ] `~/.claude/skills/_bg/launch.sh localimplement "add a comment to README.md"` — workflow id printed, `state.json` `pid` populated.
- [ ] Tail `~/.local/state/claude-workflows/<id>/log` — `[plan] init …`, `[implement] tool: Edit …`, `[review-code:sonnet] …` × 3, `[address-code] …` traces appear.
- [ ] Mid-run, `state.json` flips `stage` through `plan` / `implement` / `review-code` / `address-code` and `sub_stage` shows `{holistic, detail, scope}` flipping `running` → `done`.
- [ ] On completion, all four artifacts (`plan.md`, three `review-code-*.md`) are non-empty and the branch contains both an implementation commit and an `Address review feedback` commit.
- [ ] Ctrl-C a fresh direct invocation (`./skills/localimplement/localimplement.py "…"`) and verify no orphan `claude` processes survive (`pgrep -af claude`).
- [ ] `ghimplement` still launches — confirms `.sh` fallback in `launch.sh` works.


xbrianh (commented):

skills/localimplement/localimplement.py
```diff
@@ -0,0 +1,682 @@
...
+
+def run_claude(model: str, prompt: str, label: str, raw_path: pathlib.Path) -> None:
+    """Spawn `claude -p`, stream its stdout through log_stream, and raise
+    RuntimeError on non-zero exit."""
+    cmd = ["claude", "-p", "--model", model, *CLAUDE_FLAGS, prompt]
+    p = subprocess.Popen(
+        cmd,
+        stdout=subprocess.PIPE,
+        stderr=None,
+        bufsize=0,
```
xbrianh: `bufsize=0` on a binary pipe makes `FileIO.readline()` read one byte per `os.read()` syscall while scanning for `\n`. The stream-json output from `claude -p` during the implement stage can be large (many events per tool call, each carrying a sizable `content` blob), and every single byte of that becomes a syscall under this setting. The bash version pipelined through `tee | jq`, which gets natural OS pipe buffering.

Suggest dropping the explicit `bufsize` (default `-1` → `io.DEFAULT_BUFFER_SIZE`, i.e. `BufferedReader` with 8 KiB chunks). `readline()` on a `BufferedReader` doesn't block waiting for the buffer to fill — it returns as soon as a newline is seen or EOF — so streaming latency is preserved and per-line throughput jumps by orders of magnitude.

skills/localimplement/localimplement.py
```diff
@@ -0,0 +1,682 @@
...
+        if p.poll() is None:
+            try:
+                p.kill()
+            except Exception:
+                pass
+
+
+def _signal_handler(signum, frame):
+    _reap_all()
+    sys.exit(130)
```
xbrianh, line 101: `_signal_handler` runs synchronously on the main thread at a bytecode boundary and calls `_reap_all`, which does `with _children_lock:`. If the signal lands while the main thread is already inside the `with _children_lock:` block in `_track` or `_untrack`, the handler deadlocks — `threading.Lock` isn't reentrant.

The window is narrow (two bytecodes in the current code) and in practice the lock is usually free, but this is exactly the class of bug that's trivial to preempt and hellish to diagnose later. Two low-cost fixes:
- Make `_children_lock = threading.RLock()` (same acquire API, safe to re-enter from the same thread).
- Or don't do real work in the handler: set a flag / write a byte to a self-pipe, and reap from the main loop.

skills/localimplement/localimplement.py
```diff
@@ -0,0 +1,682 @@
...
+        '[-x <address-model>] [-a <holistic-review-model>] '
+        '[-b <detail-review-model>] [-c <scope-review-model>] "<instructions>"'
+    )
+    parser = argparse.ArgumentParser(add_help=False, usage=usage)
+    parser.add_argument("-p", "--plan-model", dest="plan", default="sonnet")
+    parser.add_argument("-i", "--impl-model", dest="impl", default="sonnet")
+    parser.add_argument("-x", "--address-model", dest="address", default="sonnet")
+    parser.add_argument("-a", "--holistic-model", dest="holistic", default="sonnet")
+    parser.add_argument("-b", "--detail-model", dest="detail", default="sonnet")
+    parser.add_argument("-c", "--scope-model", dest="scope", default="sonnet")
```
xbrianh: Minor contract note: the bash version used `getopts "p:i:x:a:b:c:"`, which accepts only the short forms. `argparse.add_argument("-p", "--plan-model", ...)` silently expands the grammar so `--plan-model`, `--impl-model`, etc. are now valid too, while the `usage` string documents only the short forms. Not harmful, but if preserving the external contract is a goal, either drop the long names or add them to the `usage` string so the surface area is explicit.

skills/localimplement/localimplement.py
```diff
@@ -0,0 +1,682 @@
...
+    cwd = pathlib.Path(".").resolve()
+    try:
+        session_resolved = session_dir.resolve()
+    except Exception:
+        session_resolved = session_dir
+    for dirpath, dirnames, filenames in os.walk(cwd):
+        dirnames[:] = [d for d in dirnames if d != ".git"]
+        dp = pathlib.Path(dirpath)
+        try:
+            if dp.resolve() == session_resolved or session_resolved in dp.resolve().parents:
```
xbrianh: Nit: `dp.resolve()` is called twice per iteration here — once for the `==` check and once for `.parents`. On a deep tree that's 2×N stat walks. Hoist into a local (`dp_resolved = dp.resolve()`) and reuse.

skills/localimplement/localimplement.py
```diff
@@ -0,0 +1,682 @@
...
+    parser.add_argument("-i", "--impl-model", dest="impl", default="sonnet")
+    parser.add_argument("-x", "--address-model", dest="address", default="sonnet")
+    parser.add_argument("-a", "--holistic-model", dest="holistic", default="sonnet")
+    parser.add_argument("-b", "--detail-model", dest="detail", default="sonnet")
+    parser.add_argument("-c", "--scope-model", dest="scope", default="sonnet")
+    parser.add_argument("instructions", nargs="*")
+    try:
+        args = parser.parse_args(argv)
+    except SystemExit:
+        die(usage)
```
xbrianh: Nit: when argparse fails it already prints its own `usage: ...\nlocalimplement.py: error: <specific msg>` to stderr before raising `SystemExit`. The `die(usage)` here then prints the usage a second time with an `error:` prefix, so a bad `-p` value produces three lines of output and buries argparse's more specific error. Either override `parser.error` to route through `die`, or just let argparse handle it (drop the try/except and the `die(usage)` branch).

xbrianh (commented):
Clean 1:1 port of the bash pipeline. The `launch.sh` `.py`→`.sh` fallback is a tidy migration hook, and replacing the `( ... claude -p | log_stream ) & ... $! + kill -0 + PIPESTATUS` dance with a thread + Popen tracker makes the Ctrl-C path actually legible. External contract (flags, `sub_stage` shape, artifact filenames, stage names) looks preserved.

Two things worth a closer look: the `Popen(..., bufsize=0)` choice forces byte-per-syscall reads on every stream-json line, and the SIGINT handler re-enters `_children_lock` via `_reap_all` — non-reentrant `Lock` means a theoretical deadlock if the signal fires while the main thread is holding it in `_track`/`_untrack`. A couple smaller notes on argparse and a hoistable `.resolve()` call inline.

skills/localimplement/localimplement.py
```diff
@@ -0,0 +1,682 @@
+#!/usr/bin/env python
+"""Background pipeline for the /localimplement skill.
+
+Runs under the _bg launcher (which exports WF_ID and manages state.json under
+${XDG_STATE_HOME:-$HOME/.local/state}/claude-workflows/<WF_ID>/). Direct
```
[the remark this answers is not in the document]
xbrianh: Updated the shebang to the py-3 form and rewrote the 3.9+/3.10+ annotations (`list[…]`, `tuple[…]`, `Exception | None`, `Path | None`) to use `typing.{List,Optional,Tuple}` so the file parses on older Python 3 interpreters too.

skills/localimplement/localimplement.py
```diff
@@ -0,0 +1,682 @@
...
+        ReviewWorker(
+            model=model_c, out_file=out_c, focus=focus_c, context=context,
+            where_field=where_field, label=f"review-code:{model_c}",
+            raw_path=session_dir / f"stream-review-code-{model_c}.jsonl",
+        ),
+    ]
+    # Stable lens labels (not model names) as sub_stage keys so the shape is
+    # unambiguous when two lenses share a model. Model name is embedded in
+    # the value so status output can show it.
+    lens_keys = ("holistic", "detail", "scope")
```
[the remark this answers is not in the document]
xbrianh: Fixed — the raw stream path now embeds the lens key (`stream-review-code-{holistic,detail,scope}-<model>.jsonl`), so the default sonnet×3 case no longer has three threads concurrently appending to one file. Also hoisted `lens_keys` above the worker construction so the suffixes are read from a single source.

skills/localimplement/localimplement.py
```diff
@@ -0,0 +1,682 @@
...
+        "holistic": SCRIPT_DIR / "lens-holistic-code.md",
+        "detail": SCRIPT_DIR / "lens-detail-code.md",
+        "scope": SCRIPT_DIR / "lens-scope-code.md",
+    }
+    for name, path in lens_files.items():
+        if not path.exists() or path.stat().st_size == 0:
+            die(f"missing or empty lens file: {path}")
+    focus_a = lens_files["holistic"].read_text()
+    focus_b = lens_files["detail"].read_text()
+    focus_c = lens_files["scope"].read_text()
```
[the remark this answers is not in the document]
xbrianh: Done — all three lens reads now pass `encoding="utf-8"` explicitly, so a non-UTF-8 locale (e.g. `LANG=C` in a minimal container) can't crash the pipeline on the em-dashes in those files.

skills/localimplement/localimplement.py
xbrianh: Dropped `bufsize=0`. Default is now `-1` → `BufferedReader` with 8 KiB reads, which scans for `\n` in-buffer instead of one `os.read()` per byte. Added a comment explaining the reasoning and that streaming latency is preserved (readline returns on newline or EOF, not on buffer-fill).

skills/localimplement/localimplement.py
xbrianh: Switched `_children_lock` to `threading.RLock()` — same acquire API, safe for the signal handler to re-enter if it lands while the main thread is already inside `_track`/`_untrack`. Updated the block comment to flag the reason.

skills/localimplement/localimplement.py
xbrianh: Dropped the long-form flags — `-p`/`-i`/`-x`/`-a`/`-b`/`-c` are now the only accepted grammar, matching the bash `getopts` contract the `usage` string documents. Verified: `--plan-model opus` now prints `unrecognized arguments: --plan-model` and exits 2.

skills/localimplement/localimplement.py
xbrianh: Hoisted — `dp_resolved = dp.resolve()` once per iteration, reused for both the equality check and the `.parents` lookup.

skills/localimplement/localimplement.py
xbrianh: Dropped the `try/except SystemExit: die(usage)` wrapper. argparse's own `usage: …\nlocalimplement.py: error: <specific>` now reaches stderr cleanly — no more triple-line output burying the specific error. Added a comment inline explaining why we don't wrap it.

PR #26, opened 2026-04-21
Mark gremlin state finished after successful rescue

### LINKED ISSUE #24: Fix rescue→land handoff in /gremlins: mark state finished on successful rescue
## Context

After `/gremlins rescue <id>` runs the inline `claude -p ...` agent and it exits 0, the gremlin's `state.json` is **not** updated. It still carries the terminal failure state from the original crash (`status=\"stopped\"`, `exit_code=1`, no `finished` marker), so `liveness_of_state_file` keeps returning `dead:exit 1`. As a result, `/gremlins land <id>` refuses the gremlin (`do_land` gates localgremlins on `live == \"dead:finished\"`), and the operator has to hand-edit `state.json` to proceed — which is exactly what the liveness safety check exists to prevent.

The intended handoff is: a successful rescue → `dead:finished` → `land` proceeds naturally.

## Approach

On the success path of `do_rescue` (returncode == 0), before printing the completion message, patch `state.json` to the finished shape and touch the `<wdir>/finished` marker. This mirrors the conventions already used by `do_stop` (gremlins.py:354-367) and by the background `finish.sh`, so `liveness_of_state_file` will classify the gremlin as `dead:finished` (per the terminal branch at gremlins.py:105-109) and `_land_local` will accept it.

Why this approach:
- Minimal, surgical — one file, one function, ~5 lines added.
- Reuses existing liveness semantics; no new states or flags.
- Failure path is deliberately untouched: if rescue exits non-zero, the gremlin remains `dead:exit <code>` so the operator can diagnose via the log/worktree (as today).

Alternative considered: relaxing `do_land`'s liveness gate to allow landing from `dead:exit` after a rescue. Rejected — that weakens the safety check for all callers, and the real bug is that state.json is lying about the current status.

## Tasks

- [ ] In `do_rescue` (skills/gremlins/gremlins.py ~493-505), on the `result.returncode == 0` branch, before the \"Rescue completed for …\" print:
  - [ ] Compute `now_iso = datetime.datetime.now(datetime.UTC).strftime(\"%Y-%m-%dT%H:%M:%SZ\")` (explicitly use `datetime.UTC` to avoid the `utcnow()` DeprecationWarning already present elsewhere in this file).
  - [ ] Set `state[\"status\"] = \"finished\"`, `state[\"exit_code\"] = 0`, `state[\"ended_at\"] = now_iso`.
  - [ ] Write `state` back to `sf` with `json.dump(..., indent=2)` inside a try/except `OSError` that prints a `warning:` line on failure (matches the pattern at gremlins.py:363-367).
  - [ ] `pathlib.Path(os.path.join(wdir, \"finished\")).touch()` inside a try/except `OSError` (matches gremlins.py:355-358).
- [ ] Leave the `returncode != 0` branch untouched so unsuccessful rescues still surface as `dead:exit <code>`.
- [ ] Leave the other `utcnow()` call-sites (gremlins.py:359, :396) alone — migrating them is out of scope for this fix.

## Verification

- [ ] Manual: take a gremlin stuck at `dead:exit 1`, run `/gremlins rescue <id>`, let it exit 0, then run `/gremlins status <id>` and confirm liveness is `dead:finished`. Follow with `/gremlins land <id>` and confirm it proceeds without hand-editing state.
- [ ] Manual negative: simulate a rescue that exits non-zero (e.g. Ctrl-C the subprocess indirectly or run with a bad prompt); confirm state.json is unchanged and liveness still reports `dead:exit <code>`.

## Open questions

None — touch point, timestamp format, and marker convention are all specified by existing code.

xbrianh:
## Summary

On the rescue success path in `do_rescue` (skills/gremlins/gremlins.py), patch `state.json` to a finished-shape record and touch the `<wdir>/finished` marker so `liveness_of_state_file` classifies the gremlin as `dead:finished` — letting `/gremlins land <id>` proceed naturally instead of refusing on `dead:exit <code>`.

- `status` → `"finished"`, `exit_code` → `0`, `ended_at` → now (UTC).
- `<wdir>/finished` marker touched (mirrors `finish.sh`).
- Failure branch untouched: a non-zero rescue still surfaces as `dead:exit <code>` for diagnosis.

Closes #24

## Test plan
- [ ] Take a gremlin stuck at `dead:exit 1`, run `/gremlins rescue <id>`, let the agent exit 0, then check `/gremlins status <id>` reports `dead:finished`.
- [ ] Run `/gremlins land <id>` on the rescued gremlin and confirm it lands without hand-editing `state.json`.
- [ ] Negative: abort a rescue with Ctrl-C (non-zero exit) and confirm `state.json` is unchanged and liveness still reports `dead:exit <code>`.

### DIFF
--- a/skills/gremlins/gremlins.py
+++ b/skills/gremlins/gremlins.py
@@ -493,6 +493,20 @@ def do_rescue(target: str) -> bool:
     try:
         result = subprocess.run(["claude", "-p", prompt], cwd=workdir)
         if result.returncode == 0:
+            now_iso = datetime.datetime.now(datetime.UTC).strftime("%Y-%m-%dT%H:%M:%SZ")
+            state["status"] = "finished"
+            state["exit_code"] = 0
+            state["ended_at"] = now_iso
+            try:
+                with open(sf, "w", encoding="utf-8") as fh:
+                    json.dump(state, fh, indent=2)
+            except OSError as e:
+                print(f"warning: could not patch state.json: {e}")
+            try:
+                pathlib.Path(os.path.join(wdir, "finished")).touch()
+            except OSError as e:
+                print(f"warning: could not touch finished marker: {e}")
+
             print()
             print(f"Rescue completed for {gr_id}.")
             print(f"Run /gremlins land {gr_id} if you are satisfied with the result.")
--- a/skills/gremlins/gremlins.py
+++ b/skills/gremlins/gremlins.py
@@ -493,8 +493,15 @@ def do_rescue(target: str) -> bool:
     try:
         result = subprocess.run(["claude", "-p", prompt], cwd=workdir)
         if result.returncode == 0:
-            now_iso = datetime.datetime.now(datetime.UTC).strftime("%Y-%m-%dT%H:%M:%SZ")
-            state["status"] = "finished"
+            # Marker first, then state.json — mirrors finish.sh. The marker is
+            # the authoritative signal for `dead:finished`; if the state write
+            # fails after the marker exists, liveness still classifies correctly.
+            try:
+                pathlib.Path(os.path.join(wdir, "finished")).touch()
+            except OSError as e:
+                print(f"warning: could not touch finished marker: {e}")
+            now_iso = datetime.datetime.now(datetime.timezone.utc).strftime("%Y-%m-%dT%H:%M:%SZ")
+            state["status"] = "done"
             state["exit_code"] = 0
             state["ended_at"] = now_iso
             try:
@@ -502,10 +509,9 @@ def do_rescue(target: str) -> bool:
                     json.dump(state, fh, indent=2)
             except OSError as e:
                 print(f"warning: could not patch state.json: {e}")
-            try:
-                pathlib.Path(os.path.join(wdir, "finished")).touch()
-            except OSError as e:
-                print(f"warning: could not touch finished marker: {e}")
+            # Unlike finish.sh, we do not `git worktree remove` here — the
+            # rescued worktree is preserved so the operator can inspect before
+            # running `/gremlins land`, which has its own cleanup path.
 
             print()
             print(f"Rescue completed for {gr_id}.")


xbrianh (commented):

skills/gremlins/gremlins.py
```diff
@@ -493,6 +493,20 @@ def do_rescue(target: str) -> bool:
     try:
         result = subprocess.run(["claude", "-p", prompt], cwd=workdir)
         if result.returncode == 0:
+            now_iso = datetime.datetime.now(datetime.UTC).strftime("%Y-%m-%dT%H:%M:%SZ")
```
xbrianh: `datetime.UTC` is Python 3.11+. The rest of this file (lines 359, 396) uses `datetime.datetime.utcnow()` (deprecated in 3.12 but widely compatible) or you could use `datetime.timezone.utc` — either would match the existing style and avoid an implicit 3.11 floor. Not a blocker if you're already on 3.11+, but worth a conscious choice.

skills/gremlins/gremlins.py
```diff
@@ -493,6 +493,20 @@ def do_rescue(target: str) -> bool:
...
+            state["status"] = "finished"
```
xbrianh: Minor convention nit: `finish.sh` writes `status="done"` on EC==0 and `do_stop` writes `status="stopped"`. Nothing in the repo actually reads `"finished"`, so this is functionally inert — `liveness_of_state_file` keys off the `finished` marker plus `exit_code`. But if a human (or future code) ever greps state.json for status values, `"finished"` will be the odd one out. Consider `"done"` to match `finish.sh`.

skills/gremlins/gremlins.py
```diff
@@ -493,6 +493,20 @@ def do_rescue(target: str) -> bool:
...
+            state["ended_at"] = now_iso
+            try:
+                with open(sf, "w", encoding="utf-8") as fh:
+                    json.dump(state, fh, indent=2)
+            except OSError as e:
+                print(f"warning: could not patch state.json: {e}")
+            try:
+                pathlib.Path(os.path.join(wdir, "finished")).touch()
+            except OSError as e:
+                print(f"warning: could not touch finished marker: {e}")
```
xbrianh: Ordering consideration: if `json.dump` fails (OSError) but the `finished` marker touch succeeds, `liveness_of_state_file` still sees the old non-zero `exit_code` in state.json and returns `dead:exit <code>` — so `/gremlins land` will still be blocked with a misleading message despite the warning. `finish.sh` sidesteps this by touching the marker *first* and treating it as the authoritative terminal signal, then best-effort updating state; flipping the order here would match. Low-probability failure mode (local disk OSError), so feel free to dismiss.

skills/gremlins/gremlins.py
```diff
@@ -493,6 +493,20 @@ def do_rescue(target: str) -> bool:
...
+            state["status"] = "finished"
+            state["exit_code"] = 0
+            state["ended_at"] = now_iso
+            try:
+                with open(sf, "w", encoding="utf-8") as fh:
+                    json.dump(state, fh, indent=2)
+            except OSError as e:
+                print(f"warning: could not patch state.json: {e}")
+            try:
+                pathlib.Path(os.path.join(wdir, "finished")).touch()
```
xbrianh: Unlike `finish.sh`, the rescue success path does not `git worktree remove` the workdir. That's arguably correct — the user will want to inspect the rescue result before running `/gremlins land`, and land has its own cleanup path — but it does mean a rescued gremlin is in a slightly different terminal shape than a naturally-finished one (state says finished, but worktree is still there). Worth a one-line comment noting the intentional divergence.

xbrianh (commented):
Nicely scoped patch — the minimal addition on the rescue success path correctly produces the `dead:finished` classification `/gremlins land` expects. A few minor consistency nits worth considering: `datetime.UTC` requires Python 3.11+ while the rest of the file uses `datetime.datetime.utcnow()`; the `status` value convention on a successful exit elsewhere is `"done"` (from `finish.sh`), not `"finished"`; and the write ordering means a `json.dump` failure followed by a successful marker-touch leaves liveness reporting `dead:exit <old-code>`. None of these block the stated goal — they're polish items.

skills/gremlins/gremlins.py
xbrianh: Fixed in acb77ee — switched to `datetime.timezone.utc` so the success path doesn't AttributeError on pre-3.11 Pythons.

skills/gremlins/gremlins.py
xbrianh: Fixed in acb77ee — using `status="done"` to match finish.sh's vocabulary.

skills/gremlins/gremlins.py
xbrianh: Fixed in acb77ee — touch the finished marker first, then patch state.json, mirroring finish.sh. Added a short comment noting the invariant.

skills/gremlins/gremlins.py
xbrianh: Added in acb77ee — inline comment noting rescue success preserves the worktree (unlike finish.sh), since the operator inspects before running /gremlins land.

skills/gremlins/gremlins.py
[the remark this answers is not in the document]
xbrianh: Fixed in acb77ee — switched to `datetime.timezone.utc` so this works pre-3.11. Thanks for catching the compatibility issue.

skills/gremlins/gremlins.py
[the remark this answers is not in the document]
xbrianh: Fixed in acb77ee — now using `status="done"` to match finish.sh. Agreed the repo already has a status vocabulary (running/stopped/done) and introducing a fourth value wasn't worth it.

PR #28, opened 2026-04-21
Strengthen reviewer file-output contract for mechanical-case short-circuit

### LINKED ISSUE #25: localgremlin: holistic reviewer short-circuit violates file-output contract
## Context

The localgremlin pipeline crashes with `review <model> did not produce <out>` when the holistic reviewer short-circuits on a mechanical change.

The runner prompt and the lens disagree about where the verdict goes:

- `skills/localgremlin/localgremlin.py:299-301` builds the reviewer prompt as `{focus}` + `` Write your review to `{out_file}`. `` — so the runner's contract is "every review lands in the file."
- `skills/localgremlin/lens-holistic-code.md:3` tells the reviewer, in the mechanical case, to "stop immediately and output exactly one line" with the mechanical verdict. In practice the model honors "stop immediately" and emits the line to chat instead of writing to `{out_file}`.
- `skills/localgremlin/localgremlin.py:418` then `die()`s because the output file is empty:
  ```python
  if not out.exists() or out.stat().st_size == 0:
      die(f"review {w.model} did not produce {out}")
  ```

The empty-file check is a correct safety net — the bug is upstream, in the prompt contract.

## Approach

Strengthen the **runner prompt** so the file-output contract is non-negotiable and explicitly covers the mechanical short-circuit case. Keep the lens focused on review heuristics; don't re-assert the contract in every lens.

Rationale for preferring the runner-prompt fix over amending the lens:
- The invariant ("every review writes to `{out_file}`") is a runner-level concern, not a lens-level one. It belongs where it applies to all lenses.
- Fixing it in the lens would require remembering to add the same guardrail to any future lens that introduces a short-circuit. The scope and detail lenses don't currently have equivalent short-circuits, but the runner-level fix protects future ones too.
- The mechanical short-circuit line in the lens reads more naturally if it stays a review heuristic ("here's when to bail early") without being tangled with plumbing ("...and by the way, write to this path").

The empty-file check at `localgremlin.py:418` stays untouched — it's a correct safety net and catches any remaining contract violations.

## Tasks

- [ ] Edit the reviewer prompt in `skills/localgremlin/localgremlin.py` (around line 299-301) so the `` Write your review to `{out_file}` `` instruction is strengthened into a non-negotiable file-output contract that explicitly covers the mechanical short-circuit case. Rough shape: tell the reviewer that `{out_file}` is the canonical output location in every case, including any one-line mechanical verdict, and that emitting the verdict only to chat is not acceptable.
- [ ] Leave `lens-holistic-code.md` unchanged (the mechanical short-circuit instruction stays a review heuristic).
- [ ] Leave the empty-file check at `localgremlin.py:418` unchanged — it's the correct safety net.
- [ ] Validate by reading back the constructed prompt string (e.g. a small local dry-run that prints the final prompt given a stub `focus` + `out_file`) and reasoning through what a model would do when the lens says "stop immediately and output exactly one line" and the runner says `{out_file}` is the canonical sink for every case, including that one-liner. The pipeline itself is hard to unit-test without a live `claude` invocation, so prompt-string inspection + reasoning is the expected validation.

## Open questions

- Wording choice for the strengthened contract is the only real degree of freedom. Something like: "`{out_file}` is the canonical and required location for your review output in every case — including any short-circuit one-liner. Do not emit the verdict only to chat; write it to the file and then stop." Final wording is up to the implementer.

xbrianh:
## Summary

- The `/localgremlin` pipeline was dying with `review <model> did not produce <out>` when the holistic reviewer short-circuited on a mechanical change (see #25).
- Root cause: the holistic lens (`lens-holistic-code.md:3`) tells the reviewer to *"stop immediately and output exactly one line"*; the model honored that literally and printed to chat, bypassing the outer prompt's `` Write your review to `{out_file}` `` instruction. The triple-review guard at `localgremlin.py:418` then killed the run.
- Fix: strengthen the runner prompt in `run_review()` so `{out_file}` is explicitly the canonical output location in every case, including any short-circuit one-liner. The lens files and the empty-file safety net stay untouched.

## Test plan

- [ ] Read back the constructed prompt and confirm the short-circuit case is covered unambiguously.
- [ ] Future `/localgremlin` run against a mechanical change (e.g. a pure rename) completes through address-code without the `did not produce` failure.

Closes #25

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -25,11 +25,15 @@ Create the file under the corresponding repo directory (`skills/<name>/SKILL.md`
 
 The `ghgremlin` skill invokes [`skills/ghgremlin/ghgremlin.sh`](skills/ghgremlin/ghgremlin.sh) to run an end-to-end GitHub-issue-driven workflow: `/ghplan`, an implementation + PR creation stage, `/ghreview` (with a scope reviewer running in parallel), and `/ghaddress`.
 
+## The local gremlin
+
+The `localgremlin` skill runs plan → implement → three parallel reviewers (holistic, detail, scope) → address-code locally via [`skills/localgremlin/localgremlin.py`](skills/localgremlin/localgremlin.py). All artifacts land in `~/.local/state/claude-gremlins/<id>/artifacts/` off the product branch.
+
 Both `/ghgremlin` and `/localgremlin` accept `--design` as a first flag to invoke `/design` before the gremlin, running an interactive spec conversation and handing off automatically when the user is ready.
 
 ## Additional skills
 
-- [`/gremlins`](skills/gremlins/SKILL.md) — on-demand status of all background gremlins on this machine. Key subcommands: `stop <id>` (SIGTERM), `rescue <id>` (diagnose and resume inline), `rm <id>` (delete state directory, worktree directory, and branch), `land <id>` (squash-land a local gremlin or merge a gh PR and clean up). Use `--here` to filter to the current repo.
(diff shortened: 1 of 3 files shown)


xbrianh:
Closing — the branch was based on a local main that had diverged from origin/main, which made the PR diff look like it reverted landed commits. Recreating cleanly against the current default.

PR #31, opened 2026-04-21
Split /gremlins rescue into foreground diagnosis + background resume

### LINKED ISSUE #30: /gremlins rescue: split into foreground debug + background resume
## Context

Today `/gremlins rescue <id>` runs the entire rescue in a single inline `claude -p` that diagnoses, fixes, **and** completes all remaining stages before exiting. Two problems:

1. The gremlin blocks the foreground for the full pipeline tail, even though only the diagnosis phase needs a human nearby.
2. Once the inline call finishes, the gremlin never rejoins the pool — `/gremlins` no longer lists it under its id.

We want rescue to feel like a natural continuation of a gremlin's life: short interactive diagnosis up front, then the gremlin quietly goes back to work in the background, visible in `/gremlins` under its **original id** with a rescue indicator. Full design in `/tmp/design-gremlins-rescue.md`.

## Approach

Split `do_rescue` in `skills/gremlins/gremlins.py` into two phases:

1. **Phase A — foreground diagnosis/fix (`claude -p` inline, today's mechanism).** The prompt is narrowed: diagnose + fix the underlying issue, then exit. Do **not** re-run remaining stages. This keeps the chat-driven interactive feel the current implementation provides.
2. **Phase B — background resume.** If Phase A exits 0, `gremlins.py` invokes `~/.claude/skills/_bg/launch.sh --resume <GR_ID>`, which reuses the existing state dir, worktree, and branch and relaunches the gremlin pipeline starting at the failed stage. Launcher returns immediately; gremlin runs under `nohup` like any fresh launch.

Reusing `launch.sh` for the resume step (rather than inventing a second launch path) satisfies the design's constraint that the background-resume step should reuse the existing mechanism. Preserving `GR_ID` and the worktree keeps downstream tools (`/gremlins land`, `stop`, `rm`, `close`, the session-summary hook) unchanged.

Each pipeline script gets a `--resume-from <stage>` flag. Stages are gated by their required artifacts — e.g. resuming into `review-code` requires `plan.md` and an implementation commit; resuming into `address-code` requires three review files. Artifacts already land in `$STATE_DIR/artifacts/` on failure (finish.sh preserves worktree + artifacts on non-zero exit), so the resume is a straight replay from whichever stage is given.

**Why not alternatives:**
- *Have Phase A also do the remaining stages but in a detached subshell* — works but duplicates launch.sh's PID/state/nohup bookkeeping, and the prompt would still need a cut-point between diagnosis and background resume. Splitting cleanly at the `claude -p` exit is simpler.
- *Skip `launch.sh` and spawn the pipeline directly from `gremlins.py`* — loses the session-detachment, PID capture, and finish.sh teardown that launch.sh already does. Worth reusing.

## Tasks

- [ ] Narrow the rescue prompt in `build_rescue_prompt()` (gremlins.py): instruct the agent to diagnose + fix the underlying issue and **stop**; do not complete remaining stages. Keep the rescue-note write. Mention that a background resume will run the remaining stages after the agent exits.
- [ ] Rework `do_rescue()` (gremlins.py):
  - Run the foreground `claude -p` diagnostic call as today.
  - On exit 0 → invoke `~/.claude/skills/_bg/launch.sh --resume <GR_ID>` and print its output; return success.
  - On non-zero exit or `KeyboardInterrupt` → leave state.json unchanged, do **not** touch a `finished` marker, print a note that rescue was aborted and the gremlin's existing state is preserved. User can rescue again, `rm`, or `close`.
  - Remove the current logic that touches `finished` + patches `status=done` after a successful rescue — the background resume owns that now via `finish.sh`.
- [ ] Add `--resume <GR_ID>` mode to `skills/_bg/launch.sh`:
  - Skip ID generation, slugification, state-dir creation, and worktree/branch creation.
  - Load existing `state.json`; extract `workdir`, `branch`, `kind`, `stage` (= failed stage), `instructions`.
  - Remove the `finished` and `summarized` markers so the rescued gremlin gets announced on next completion.
  - Patch `state.json`: `status=running`, clear `exit_code`/`ended_at`, bump `rescue_count` (new field, default 0 → 1 → 2 …), write `rescued_at=<now>`, record `resumed_from_stage=<failed stage>`.
  - Resolve the pipeline script the same way as today and spawn it under the same `nohup` subshell, passing `--resume-from <stage>` plus the original `instructions`. Capture the new PID into state.json.
  - Refuse if the gremlin is `running` or `stalled:` or `dead:finished` (should be enforced by gremlins.py too, but defense in depth).
- [ ] Add `--resume-from <stage>` to `skills/localgremlin/localgremlin.py`:
  - New argparse flag; valid stages: `plan`, `implement`, `review-code`, `address-code`.
  - When set, skip to the named stage. Precondition checks on each skip:
    - `implement` → `plan.md` exists and non-empty.
    - `review-code` → `plan.md` exists; assert implementation commit exists (`git log -1` touches expected files) or working tree is non-empty — mirror the existing post-implement invariant.
    - `address-code` → all three `review-code-*.md` files exist and non-empty.
  - A resumed `review-code` simply re-runs the whole triple reviewer fan-out; partial sub_stage state from a prior half-finished run is overwritten. Document this in a short comment.
- [ ] Add `--resume-from <stage>` to `skills/ghgremlin/ghgremlin.sh`:
  - Valid stages match `GREMLIN_STAGES["ghgremlin"]`: `plan`, `implement`, `commit-pr`, `request-copilot`, `ghreview`, `wait-copilot`, `ghaddress`.
  - Skip earlier stages, rehydrating the shell variables later stages need:
    - `ISSUE_URL`/`ISSUE_NUM`/`ISSUE_BODY`: persist `issue_url` into `state.json` during the plan stage (new — today only `pr_url` is persisted), then fetch body from `gh issue view` on resume.
    - `PR_URL`/`PR_NUM`/`SCOPE_PR_DIFF`: `pr_url` already in state.json; derive `PR_NUM` from it, fetch diff on resume.
    - `IMPL_SESSION`: not persisted; simplest path is to refuse resuming into `commit-pr` if `IMPL_SESSION` is unavailable and force the resume to start from `implement` instead. Document this trade-off in a short comment. (Alternative: persist `impl_session_id` during 2a — noted as an open question.)
- [ ] Expose rescue state in `/gremlins` output (`gremlins.py`):
  - `build_row` reads `rescue_count` from state; when > 0, append ` (rescue)` or ` (rescue x<N>)` to the `LIVENESS` column (for running + dead cases alike).
  - Drill-in view already prints every state.json field, so `rescue_count`, `rescued_at`, and `resumed_from_stage` surface naturally.
- [ ] Refresh docs:
  - `skills/gremlins/SKILL.md`: update the `rescue <id>` bullet to describe two-phase behavior and the persistent rescue indicator.
  - `CLAUDE.md` (repo root): the `/gremlins` paragraph mentions rescue — add a clause noting the foreground/background split.
  - `README.md` — if it mentions rescue at the same level of detail, update in kind; otherwise leave alone.
- [ ] Smoke-test end to end (manual; document commands in PR description):
  - Start a localgremlin, force a failure mid-implement (e.g. run with a known-broken instruction), observe `dead:*` in `/gremlins`.
  - `/gremlins rescue <id>` — confirm foreground debug exits, then `/gremlins` shows the same id running with `(rescue)` indicator.
  - Let it finish; confirm `finish.sh` fires, `dead:finished` appears, `/gremlins land <id>` still works.
  - Deliberately fail the same gremlin a second time post-rescue and confirm a second rescue works.

## Open questions

1. **Persist `impl_session_id` for ghgremlin resume?** Today `ghgremlin.sh` uses `--resume "$IMPL_SESSION"` in stage 2b (commit-pr) to land commits in the same Claude session as stage 2a. If we don't persist it, resuming into `commit-pr` is impossible and we have to rewind to `implement`. That re-does work but is simpler. Worth the extra field? (Proposal: skip for v1; add later if rewinding proves painful.)
2. **How does the operator signal "unsalvageable"?** Design says "the transition to background happens implicitly — no explicit cue". Propose: `claude -p` exit code is the signal. Exit 0 → ready to resume. Non-zero / Ctrl-C / user tells the agent to bail → no handoff. Is that good enough, or should there be a `/rescue-abort` convention for the agent to emit?
3. **Should a resumed `wait-copilot` skip the wait?** If ghgremlin died because Copilot timed out, resuming re-enters the same 10-minute wait. Probably fine — if the operator needs to skip, they can nudge during Phase A. Flagging for visibility.
4. **Rescue-loop throttling?** Design non-goal, but should we at least emit a warning when `rescue_count > 3`? (Proposal: no, keep it off.)
5. **Should Phase A be allowed to re-plan?** If `rescue-from=plan` is permitted, the resumed gremlin will overwrite `plan.md`. Acceptable?

xbrianh:
## Summary

Reworks `/gremlins rescue <id>` from a single inline agent that completes
the entire pipeline into a two-phase flow:

- **Phase A (foreground)** — inline `claude -p` diagnoses the failure and
  fixes the underlying issue, then exits. The prompt is narrowed: no more
  replaying remaining stages from the rescue agent.
- **Phase B (background)** — on Phase A exit 0, `do_rescue` invokes
  `~/.claude/skills/_bg/launch.sh --resume <GR_ID>`, which reuses the
  existing state dir, worktree, and branch, clears the `finished`/
  `summarized` markers, bumps `rescue_count` / records
  `rescued_at` + `resumed_from_stage` in `state.json`, and relaunches the
  gremlin under `nohup` with `--resume-from <stage>`. The gremlin rejoins
  `/gremlins` under its original id with a `(rescue)` / `(rescue x<N>)`
  marker appended to the `LIVENESS` column.

Each pipeline gains a `--resume-from <stage>` flag:

- `localgremlin.py` — stages `plan`, `implement`, `review-code`,
  `address-code`. Preconditions: `plan.md` exists for implement+;
  implementation evidence in the worktree for review-code+; all three
  review files for address-code.
- `ghgremlin.sh` — stages `plan`, `implement`, `commit-pr`,
  `request-copilot`, `ghreview`, `wait-copilot`, `ghaddress`. The plan
  stage now persists `issue_url` to `state.json` (alongside the existing
  `pr_url`) so skipped-plan resumes can rehydrate without re-running
  `/ghplan`. `--resume-from commit-pr` auto-rewinds to `implement`
  because `IMPL_SESSION` isn't persisted.

Aborts (Phase A Ctrl-C or non-zero exit) leave state untouched — operator
can retry rescue, rm, or close.

## Test plan

- [ ] Start a localgremlin, force an implement-stage failure, observe `dead:*` in `/gremlins`
- [ ] `/gremlins rescue <id>` — confirm foreground debug exits, `/gremlins` shows same id running with `(rescue)` marker
- [ ] Let the resumed gremlin finish; confirm `finish.sh` fires, `dead:finished` appears, `/gremlins land <id>` still works
- [ ] Force a second failure after rescue; confirm second rescue bumps marker to `(rescue x2)`
- [ ] Ghgremlin rescue: force a `ghreview`-stage failure, rescue, confirm `issue_url` rehydrates without re-running `/ghplan`

Closes #30


skills/_bg/launch.sh
```diff
@@ -40,19 +40,139 @@ slugify() {
...
+    STATE_ROOT="${XDG_STATE_HOME:-$HOME/.local/state}/claude-gremlins"
+    STATE_DIR="$STATE_ROOT/$RESUME_GR_ID"
+    STATE_FILE="$STATE_DIR/state.json"
+    [[ -d "$STATE_DIR" && -f "$STATE_FILE" ]] || die "no state at $STATE_DIR"
+
+    RESUME_KIND=$(jq   -r '.kind         // ""' "$STATE_FILE")
+    WORKDIR=$(jq       -r '.workdir      // ""' "$STATE_FILE")
+    BRANCH=$(jq        -r '.branch       // ""' "$STATE_FILE")
+    STAGE=$(jq         -r '.stage        // ""' "$STATE_FILE")
+    INSTRUCTIONS=$(jq  -r '.instructions // ""' "$STATE_FILE")
```
xbrianh: Truncation hazard: `.instructions` in state.json is written as `INSTR_SUMMARY` (launch.sh:207), which is `${INSTR_RAW:0:200}` — so on resume, `INSTRUCTIONS` is the 200-char-truncated version, not the full prompt. Today this is latent because (a) spec-file handoffs from `/design` pass a short path, and (b) resume-from=plan is the only path that re-feeds INSTRUCTIONS to a model (localgremlin's plan prompt and ghgremlin's `/ghplan`). But a rescue of a gremlin that crashed mid-plan on a ≥200-char instruction string will silently re-plan against truncated text, which is the exact scenario rescue is supposed to make safe. Worth either storing the full instructions separately (e.g. `instructions_full` alongside the display-truncated `instructions`) or carrying the full string out-of-band (e.g. `$STATE_DIR/instructions.txt`) so resume can rehydrate faithfully.

skills/_bg/launch.sh
```diff
@@ -40,19 +40,139 @@ slugify() {
...
+       --arg    resumed_from_stage "$STAGE" \
+       '.status = $status
+        | .stage = $stage
+        | del(.exit_code)
+        | del(.ended_at)
+        | .rescue_count = ((.rescue_count // 0) + 1)
+        | .rescued_at = $rescued_at
+        | .resumed_from_stage = $resumed_from_stage
+        | .pid = null' \
+       "$STATE_FILE" > "$STATE_TMP" || die "failed to patch state.json"
```
xbrianh, line 164: Consider adding `| del(.sub_stage)` to this jq filter. A localgremlin that crashed mid-`review-code` leaves `.sub_stage = {holistic: ..., detail: ..., scope: ...}` in state.json. After this patch, status/stage are freshened but sub_stage retains the pre-crash values until the resumed pipeline calls `set_stage("review-code")` (which, via set-stage.sh without a sub_stage arg, does del-sub_stage). Cosmetic only — the gremlin displays stale sub_stage info in `/gremlins` for the few seconds between launch.sh patching state and the resumed pipeline emitting its first set_stage call — but trivial to eliminate here.

skills/_bg/launch.sh
```diff
@@ -40,19 +40,139 @@ slugify() {
...
+        printf '\n--- resume at %s (from stage: %s) ---\n' "$NOW_ISO" "$STAGE"
+    } >> "$STATE_DIR/log" 2>/dev/null || true
+
+    export PIPELINE
+    export GR_ID="$RESUME_GR_ID"
+
+    (
+        cd "$WORKDIR"
+        nohup bash -c '"$PIPELINE" "$@"; EC=$?; "$HOME/.claude/skills/_bg/finish.sh" "$GR_ID" "$EC"' \
+            -- --resume-from "$STAGE" "$INSTRUCTIONS" </dev/null >>"$STATE_DIR/log" 2>&1 &
```
xbrianh: The invocation only forwards `--resume-from <stage>` and `$INSTRUCTIONS`; original flags aren't preserved. For localgremlin, if the original run used custom `-a/-b/-c` models, the resumed pipeline runs with defaults (sonnet×3), which means:
- resume-from=review-code writes new files named `review-code-*-sonnet.md` while the old `review-code-*-<custommodel>.md` are left on disk untouched — land logic and /gremlins views still see both sets.
- resume-from=address-code will hit the precondition die at localgremlin.py:623–625 with a clear error, which is at least loud.

For ghgremlin, `-r <ref>` isn't persisted either, but it's only consumed at plan, and rescue-from=plan on an already-issue-created gremlin has bigger problems (see other comment). The localgremlin case is the more material one — either persist the original argv in state.json and forward it here, or call out the limitation in gremlins.py's rescue prompt / SKILL.md so an operator rescuing a custom-model gremlin isn't surprised.

skills/ghgremlin/ghgremlin.sh
```diff
@@ -119,15 +183,28 @@ extract_session_id() {
...
+  PLAN_OUT=$(claude -p "${CLAUDE_FLAGS[@]}" "/ghplan ${REF:+$REF }${INSTRUCTIONS}" | progress_tee)
+  ISSUE_URL=$(extract_gh_url "$PLAN_OUT" \
+    'https://github\.com/[^ )]+/issues/[0-9]+' \
+    'gh issue create' \
+    "issue")
+  ISSUE_NUM=$(basename "$ISSUE_URL")
+  echo "    issue: $ISSUE_URL"
+  # Persist issue_url so --resume-from implement (or later) can rehydrate
+  # ISSUE_URL without re-running /ghplan.
+  patch_state '.issue_url = $url' --arg url "$ISSUE_URL"
```
xbrianh, line 219: Duplicate-issue window: `/ghplan` creates the GitHub issue via the agent's `gh issue create` call, then we extract the URL and only afterwards persist `.issue_url` via `patch_state`. If anything between `gh issue create` succeeding and this line fails (claude -p exits non-zero after issue creation, extract_gh_url dies, patch_state warns-and-returns), state.stage stays at `plan` with no `issue_url`, and a subsequent `/gremlins rescue` will re-run `/ghplan` and create a **duplicate issue**.

Mitigations worth considering:
- Have the plan-stage agent print the issue URL to a file we can snapshot to disk before state is patched, so we don't depend on both the extraction regex succeeding AND state.json being writable.
- Or, detect the duplicate pre-create: have the plan prompt check for an existing open issue whose body matches the instructions before creating a new one.
- At minimum: if extract_gh_url fails, scan raw stream-json for any `gh issue create` tool_use that succeeded and surface that URL in the die message so an operator can hand-populate state.json before rescue.

skills/gremlins/gremlins.py
```diff
@@ -487,48 +492,51 @@ def do_rescue(target: str) -> bool:
...
+        print("Gremlin state preserved — rerun /gremlins rescue, rm, or close.")
+        print(f"Inspect the log at {log_path} and worktree at {workdir} for details.")
+        return False
+
+    # Phase B: hand off to launch.sh --resume so the remaining stages run in the
+    # background under the same GR_ID. launch.sh patches state.json, clears the
+    # finished/summarized markers, and relaunches the pipeline with
+    # --resume-from <stage>.
+    launcher = os.path.expanduser("~/.claude/skills/_bg/launch.sh")
+    if not os.path.isfile(launcher):
```
xbrianh: `os.path.isfile` only — an un-chmod'd launcher (0644) would pass this check, then `subprocess.run([launcher, ...])` would raise `PermissionError`, which isn't caught by the `except FileNotFoundError` below. Worth either:
```python
if not os.access(launcher, os.X_OK):
    print(f"error: launcher not executable at {launcher}")
    return False
```
or catching `OSError`/`PermissionError` in the subprocess block. Low-severity; only bites if someone hand-edits skills/ without `chmod +x`.

skills/gremlins/gremlins.py
```diff
@@ -423,11 +430,9 @@ def build_rescue_prompt(state, wdir, log_tail, artifact_paths):
...
 
 1. Diagnose the failure from the log above.
-2. Fix the underlying issue in this worktree (working directory: {workdir}).
-3. Re-run the failed stage ({stage}) and then complete the remaining stages: {', '.join(remaining) or '(none)'}.
-   For each stage, perform what the pipeline script would have done.
-   Commit your changes when the implement stage is complete.
-4. Write a brief rescue note to `{rescue_note_path}` describing what failed and what you fixed.
+2. Fix the underlying issue in this worktree (working directory: {workdir}) so that rerunning the failed stage ({stage}) will succeed. This may mean editing code, cleaning up partial state, or staging missing artifacts.
+3. Write a brief rescue note to `{rescue_note_path}` describing what failed and what you fixed.
+4. STOP. Do NOT re-run the failed stage or any remaining stages yourself — after you exit, a background resume will relaunch the gremlin pipeline starting at {stage} and complete the rest automatically. Exit 0 when the fix is in place; exit non-zero (or abort) if the failure is unsalvageable so the operator can decide next steps.
```
xbrianh: "Exit 0 when the fix is in place; exit non-zero (or abort) if the failure is unsalvageable" reads well, but `claude -p` returns 0 on any successful generation regardless of what the agent concluded — the agent would have to explicitly invoke `Bash: exit 1` to propagate an unsalvageable verdict, which isn't a natural motion. In practice, Phase B always fires as long as Phase A didn't crash or get Ctrl-C'd, even when the agent decided it couldn't fix the issue. Consider either:
- Having the agent write a decision file (`$wdir/artifacts/rescue-decision.txt` with `salvageable` or `unsalvageable`) that do_rescue inspects before launching Phase B, or
- Dropping the "exit non-zero" framing and trusting the operator to Ctrl-C Phase A when they see the agent give up.

As written, the prompt implies a gating behavior that isn't actually enforced.

xbrianh (commented):
Two-phase rescue is a clean split and the `--resume-from` plumbing mirrors the stage order well. Two real concerns worth addressing before landing: (1) on resume, `INSTRUCTIONS` is rehydrated from state.json's `.instructions`, which is stored truncated to 200 chars — a resume-from=plan with a long task will re-plan against a truncated prompt; (2) ghgremlin persists `issue_url` only *after* URL extraction, so a plan failure between `gh issue create` and `patch_state` will silently create a duplicate issue when the gremlin is rescued. A few smaller nits inline (stale `.sub_stage` on resume, missing model-flag passthrough, launcher `isfile` vs executable check, aspirational `exit 0`/`exit non-zero` semantics in the rescue prompt).

xbrianh (commented):

skills/localgremlin/localgremlin.py
```diff
@@ -579,11 +587,49 @@ def main(argv: List[str]) -> int:
...
+                ["git", "status", "--porcelain"],
+                capture_output=True, text=True, check=False,
+            )
+            has_dirty = bool(porcelain.stdout.strip())
+            r = subprocess.run(
+                ["git", "rev-list", "--count", "HEAD"],
+                capture_output=True, text=True, check=False,
+            )
+            has_commits = (r.returncode == 0 and int(r.stdout.strip() or "0") > 0)
+            if not has_dirty and not has_commits:
```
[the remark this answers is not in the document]
xbrianh: Fixed in 8f3430c. The non-git branch of the review-code/address-code precondition now falls back to walking the worktree for any non-metadata file (skipping .git and the session artifacts dir). The cross-resume sentinel isn't available, so a strictly "modified since pre-impl" check isn't possible; empty worktree is the only unambiguous "nothing was implemented" signal in non-git mode, which is analogous to "empty HEAD history" in git mode.

skills/_bg/launch.sh
```diff
@@ -40,19 +40,139 @@ slugify() {
...
+    rm -f "$STATE_DIR/finished" "$STATE_DIR/summarized" 2>/dev/null || true
+
+    NOW_ISO="$(date -u +%Y-%m-%dT%H:%M:%SZ)"
+    STATE_TMP="$STATE_FILE.tmp"
+    jq --arg    status             "running" \
+       --arg    stage              "$STAGE" \
+       --arg    rescued_at         "$NOW_ISO" \
+       --arg    resumed_from_stage "$STAGE" \
+       '.status = $status
+        | .stage = $stage
```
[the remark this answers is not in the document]
xbrianh: Fixed in 8f3430c. Added `del(.sub_stage) | del(.stage_updated_at)` to the resume jq filter so /gremlins doesn't show stale pre-crash values during the window between launch.sh patching state and the resumed pipeline's first set-stage call.

skills/_bg/launch.sh
xbrianh: Fixed in 8f3430c. launch.sh now writes the full untruncated $INSTR_RAW to $STATE_DIR/instructions.txt on initial launch; --resume reads from the sidecar and only falls back to .instructions if the file is missing (older gremlins). Rescue mid-plan on a ≥200-char instruction string now re-feeds the original prompt faithfully.

skills/_bg/launch.sh
xbrianh: Fixed in 8f3430c — see also Copilot's sibling comment. The resume jq filter now includes `del(.sub_stage) | del(.stage_updated_at)`.

skills/_bg/launch.sh
xbrianh: Fixed in 8f3430c. launch.sh now captures the original pipeline-level flags (everything before the first positional, i.e. `${_args[@]:0:$_i}`) into state.json as a `pipeline_args` JSON array on initial launch. On --resume, those args are rehydrated and re-prepended to the pipeline invocation before `--resume-from $STAGE $INSTRUCTIONS`, so localgremlin's -a/-b/-c and ghgremlin's -r ref survive a rescue. Gremlins launched before this change have no `pipeline_args` field and fall back to empty (defaults), same as today's behavior — called out in the comment.

skills/ghgremlin/ghgremlin.sh
xbrianh: Partially addressed in 8f3430c. I didn't retrofit pre-create duplicate detection (that requires restructuring the plan prompt / extending `/ghplan`), but the plan stage now tees its stream-json to `$STATE_DIR/artifacts/ghplan-out.jsonl`, and extract_gh_url prints a hint pointing at that file on failure. That means if claude -p exits non-zero after `gh issue create` succeeded, the operator has the raw tool_use/tool_result events on disk to hand-extract the URL, patch `state.json` with `.issue_url`, and then rescue without a duplicate-create. Pre-create dedup is a larger change — deferred until we see this fire in practice.

skills/gremlins/gremlins.py
xbrianh: Fixed in 8f3430c. Switched to `os.access(launcher, os.X_OK)` so a non-executable launcher surfaces a clean error instead of a PermissionError the FileNotFoundError handler would miss.

skills/gremlins/gremlins.py
xbrianh: Fixed in 8f3430c. Rewrote that bullet to drop the "exit 0 / exit non-zero" framing (claude -p doesn't propagate agent verdicts via exit code) and instead tell the agent to say "unsalvageable" clearly in its final message + rescue note, leaving the worktree untouched so the operator can Ctrl-C Phase A if they agree. Didn't go with the decision-file approach — the operator is already watching Phase A output live, so a prose verdict + Ctrl-C is the simpler contract.

PR #33, opened 2026-04-23
only sync for non-empty diffs

xbrianh:


scripts/sync.sh
```diff
@@ -83,12 +83,30 @@ confirm() {
...
+# between src_dir and dst_dir. Feeding these to rsync via --exclude-from makes rsync
+# skip the file entirely — not just the content transfer but the mtime/perms update
+# that --checksum alone still performs on content-identical files.
+identical_files() {
+    local src_dir="$1" dst_dir="$2"
+    [[ ! -d "$src_dir" || ! -d "$dst_dir" ]] && return 0
+    (cd "$src_dir" && find . -type f) | while IFS= read -r rel; do
+        rel="${rel#./}"
+        if [[ -f "$dst_dir/$rel" ]] && cmp -s "$src_dir/$rel" "$dst_dir/$rel"; then
+            printf '/%s\n' "$rel"
```
xbrianh: rsync treats each line in `--exclude-from` as a filter pattern, not a literal path. A tracked filename containing `*`, `?`, `[`, `]`, or a leading `#` (comment) would be mis-interpreted — either matching too much or being silently dropped. Not a live issue for the current config tree, but `find -print0` + `printf '/%s\0'` + `--from0 --exclude-from=...` would make this robust against future additions (e.g. a `cmd?.md`).

If you'd rather not add `--from0`, a one-line note in the comment block that the helper assumes no filter-metacharacters in filenames would at least document the assumption.

scripts/sync.sh
```diff
@@ -100,7 +118,11 @@ sync_dir() {
         return
     fi
     mkdir -p "$dst"
-    rsync "${BASE_FLAGS[@]}" --delete "$src/" "$dst/"
+    local exclude_file
+    exclude_file=$(mktemp)
+    identical_files "$src" "$dst" > "$exclude_file"
+    rsync "${BASE_FLAGS[@]}" --delete --exclude-from="$exclude_file" "$src/" "$dst/"
+    rm -f "$exclude_file"
```
xbrianh: With `set -euo pipefail` at the top of the script, a non-zero rsync exit will abort the function before this `rm -f` runs, leaking the temp file into `$TMPDIR`. A `trap` immediately after `mktemp` is a small hardening win:

```bash
exclude_file=$(mktemp)
trap 'rm -f "$exclude_file"' RETURN
```

(Be careful with `trap RETURN` if this function is ever called from a subshell where a parent trap matters — for this file's usage pattern it's fine.)

scripts/sync.sh
```diff
@@ -83,12 +83,30 @@ confirm() {
...
 [[ $DRY -eq 1 ]] && BASE_FLAGS+=(--dry-run)
 
+# Emit rsync-anchored paths ("/relpath") for regular files that are byte-identical
+# between src_dir and dst_dir. Feeding these to rsync via --exclude-from makes rsync
+# skip the file entirely — not just the content transfer but the mtime/perms update
+# that --checksum alone still performs on content-identical files.
+identical_files() {
+    local src_dir="$1" dst_dir="$2"
+    [[ ! -d "$src_dir" || ! -d "$dst_dir" ]] && return 0
+    (cd "$src_dir" && find . -type f) | while IFS= read -r rel; do
```
xbrianh: Filenames with embedded newlines will split across iterations of the `while read` loop. Config files basically never have this, but the `find -print0` / `read -d ''` pattern would handle it cleanly if you're already going to `--from0` for the metacharacter issue above.

xbrianh (commented):
Approach is correct and targets a real bug in the prior `--checksum`-only fix (that flag skips the content transfer but still updates mtime/perms on identical files, which is exactly the noise this PR removes). The pre-filter + `--exclude-from` dance is portable across macOS and Linux. Two small concerns worth addressing before this is battle-hardened: rsync filter metacharacters in generated exclude patterns, and a temp-file leak on rsync failure.

scripts/sync.sh
xbrianh: Fixed in 6e421ef — switched `identical_files` to `find -print0` and rsync `--from0 --exclude-from=` so paths are NUL-delimited literals rather than filter patterns. Verified locally with `cmd?.md` and `#hash.md` fixtures: both are now correctly excluded.

scripts/sync.sh
xbrianh: Fixed in 6e421ef — the rsync call now runs inside a subshell with `trap 'rm -f "$exclude_file"' EXIT`, so the temp file is cleaned up even if rsync exits non-zero under `set -e`. Went with a subshell+EXIT (rather than a function-scoped RETURN) because an early `return` path in `sync_dir` was leaving a stale RETURN trap that fired on a later call with `exclude_file` unbound — the subshell scopes the trap tightly to just the rsync invocation.

scripts/sync.sh
xbrianh: Fixed in 6e421ef — now using `find -print0` with `while IFS= read -r -d ''` so embedded newlines in filenames are preserved.

scripts/sync.sh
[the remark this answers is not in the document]
xbrianh: Fixed in 6e421ef — `identical_files` now only excludes files where contents AND permission bits match, via a new portable `file_mode` helper (macOS `stat -f '%Lp'` / Linux `stat -c '%a'`). Mode-only drift (e.g. a dropped exec bit on a `.sh`) no longer gets skipped, so rsync -a can still repair it. Verified with a chmod-drift fixture: rsync dry-run emits `.f...p...` for the mode-only case while still excluding the fully-identical one.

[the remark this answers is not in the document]
xbrianh: Fixed in 6e421ef — rsync now runs inside a subshell with `trap 'rm -f "$exclude_file"' EXIT`, so the temp file is cleaned up on success, on rsync failure under `set -e`, and on any other early exit path.

scripts/sync.sh
```diff
@@ -83,12 +83,30 @@ confirm() {
...
+
 sync_file() {
     local src="$1" dst="$2"
     if [[ ! -e "$src" ]]; then
         echo "skip: $src (not present)"
         return
     fi
+    if [[ -f "$src" && -f "$dst" ]] && cmp -s "$src" "$dst"; then
+        return
+    fi
```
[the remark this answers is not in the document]
xbrianh: Fixed in 6e421ef — `sync_file` now uses the new `fully_identical` helper, which checks both content (`cmp -s`) and permission bits. A mode-only drift (e.g. an exec bit missing on `~/.claude/skills/_bg/set-stage.sh`) no longer short-circuits, so rsync -a still reconciles it.

PR #34, opened 2026-04-23
Stop gremlin status from polluting claude -p commit messages

xbrianh:
## Summary

- `session-summary.sh` injects a "Background gremlins — finished since last check" block with an `IMPORTANT: surface this verbatim` directive into every `claude` invocation's context via `additionalContext`.
- When `gremlins.py` `land` shells out to `claude -p` to synthesize a squash-merge commit message, that directive overrides the prompt's "output only the commit message" instruction, so landed commits end up with a subject like `**Background gremlins — finished since last check:**`.
- Fix: the hook now exits early on `CLAUDE_SKIP_GREMLIN_SUMMARY=1`, and `gremlins.py` sets that env var only when invoking `claude -p` for commit-message synthesis. Other `claude -p` callers (the pipelines themselves, the rescue agent) are untouched and still see the summary.

## Test plan

- [ ] Land a local gremlin with a non-empty unsummarized finished peer in the state dir; confirm the composed subject is a real commit title, not the status block.
- [ ] Confirm `SessionStart` / `UserPromptSubmit` hooks in a normal Claude Code session still produce the gremlin summary (env var is not set outside the synthesis path).
- [ ] Confirm the hook exits cleanly on stdin containing a hook-event JSON when `CLAUDE_SKIP_GREMLIN_SUMMARY=1`.

### DIFF
--- a/skills/_bg/session-summary.sh
+++ b/skills/_bg/session-summary.sh
@@ -9,6 +9,12 @@
 # session.
 set -u
 
+# Opt-out for callers that invoke `claude -p` for structured text output (e.g.
+# commit-message synthesis in gremlins.py `land`). The hook's "surface this
+# verbatim" directive otherwise prepends the status block to the model's reply
+# and corrupts the caller's parse.
+[[ "${CLAUDE_SKIP_GREMLIN_SUMMARY:-0}" == "1" ]] && exit 0
+
 # Missing jq → nothing to report. Bail before `set -e` would bite us anywhere.
 command -v jq >/dev/null 2>&1 || exit 0
 
--- a/skills/_bg/session-summary.sh
+++ b/skills/_bg/session-summary.sh
@@ -12,10 +12,11 @@ set -u
 # Opt-out for callers that invoke `claude -p` for structured text output (e.g.
 # commit-message synthesis in gremlins.py `land`). The hook's "surface this
 # verbatim" directive otherwise prepends the status block to the model's reply
-# and corrupts the caller's parse.
-[[ "${CLAUDE_SKIP_GREMLIN_SUMMARY:-0}" == "1" ]] && exit 0
+# and corrupts the caller's parse. Repo-scoped prefix (not `CLAUDE_`) so we
+# don't collide with Claude Code's own env-var namespace.
+[[ "${GREMLIN_SKIP_SUMMARY:-0}" == "1" ]] && exit 0
 
-# Missing jq → nothing to report. Bail before `set -e` would bite us anywhere.
+# Missing jq → nothing to report, so exit quietly.
 command -v jq >/dev/null 2>&1 || exit 0
 
 STATE_ROOT="${XDG_STATE_HOME:-$HOME/.local/state}/claude-gremlins"
--- a/skills/gremlins/gremlins.py
+++ b/skills/gremlins/gremlins.py
@@ -849,9 +849,13 @@ Requirements:
 
 Output only the commit message text, nothing else."""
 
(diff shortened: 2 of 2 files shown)


skills/_bg/session-summary.sh
```diff
@@ -9,6 +9,12 @@
 # session.
 set -u
 
+# Opt-out for callers that invoke `claude -p` for structured text output (e.g.
+# commit-message synthesis in gremlins.py `land`). The hook's "surface this
+# verbatim" directive otherwise prepends the status block to the model's reply
+# and corrupts the caller's parse.
+[[ "${CLAUDE_SKIP_GREMLIN_SUMMARY:-0}" == "1" ]] && exit 0
```
xbrianh: Claude Code itself uses the `CLAUDE_` env-var prefix for its own variables (e.g. `CLAUDE_PROJECT_DIR` from the hook docs). A user-defined opt-out sharing that namespace risks (a) future collision if Claude Code ships an env var with a similar name, and (b) reader confusion about whether this is a stock Claude Code mechanism or a repo-local one. Consider a repo-scoped prefix instead — e.g. `GREMLIN_SKIP_SUMMARY` or `AMORPHOUS_SKIP_GREMLIN_SUMMARY`. The paired `gremlins.py` line would change with it.

skills/gremlins/gremlins.py
```diff
@@ -849,9 +849,13 @@ def _synthesize_commit_message_ai(inputs: dict) -> tuple:
 
 Output only the commit message text, nothing else."""
 
+    # Suppress the session-summary hook so its "surface this verbatim" directive
+    # doesn't prepend the gremlin status block to the commit message.
+    env = os.environ.copy()
+    env["CLAUDE_SKIP_GREMLIN_SUMMARY"] = "1"
```
xbrianh: This is the only `claude -p` call in the repo that needs a clean text output today, so setting the env inline is fine. But if a second text-parse caller ever appears (e.g. a future `claude -p` used to distill review feedback), the env var has to be remembered again. A tiny helper — `run_claude_p_text(prompt: str) -> str` that owns the env mutation and the `subprocess.run` kwargs — would make the contract harder to drift from. Not a blocker for this PR.

xbrianh (commented):
Fix is small, surgical, and targets the right seam: the hook can't filter itself by caller, so the caller signals intent via env. One naming nit on the env var (CLAUDE_ prefix collides with Claude Code's own namespace), plus a note that the fix is scoped to a single `claude -p` invocation — any future text-parse caller in this repo would need the same env set, so a thin wrapper might age better.

skills/_bg/session-summary.sh
```diff
@@ -9,6 +9,12 @@
...
+
 # Missing jq → nothing to report. Bail before `set -e` would bite us anywhere.
```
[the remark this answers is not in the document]
xbrianh: Reworded in 418bfdb — dropped the misleading `set -e` reference.

skills/_bg/session-summary.sh
xbrianh: Renamed to `GREMLIN_SKIP_SUMMARY` in 418bfdb (both hook and caller updated) to stay out of the `CLAUDE_` namespace.

skills/gremlins/gremlins.py
xbrianh: Done in 418bfdb — extracted `_run_claude_p_text(prompt)` that owns the `GREMLIN_SKIP_SUMMARY` env mutation and the `subprocess.run` kwargs. Future text-parse callers inherit the opt-out automatically.

PR #36, opened 2026-04-23
Add headless mode to /gremlins rescue

### LINKED ISSUE #35: Headless rescue mode for /gremlins rescue
## Context

Today `/gremlins rescue` is interactive-only: Phase A spawns `claude -p` inline in the gremlin's worktree expecting a human at a TTY to watch the diagnosis/fix and optionally Ctrl-C to abort, and Phase B hands off to `launch.sh --resume` ([`skills/gremlins/gremlins.py:441-542`](skills/gremlins/gremlins.py)). That's fine when a person is driving, but it blocks other automation — notably a future bossgremlin — from recovering a failed child without human presence. We also want humans to be able to fire-and-forget an unattended rescue.

This issue adds a `--headless` mode that runs the full rescue end-to-end with no TTY and no user input, refuses failure classes that shouldn't be auto-fixed, caps attempts per gremlin, and leaves a machine-readable bail reason in `state.json` when it declines to proceed. The default (interactive) flow is unchanged.

Reference spec: `/tmp/design-headless-rescue.md` (local).

## Approach

Four additive pieces, staged so each is independently useful:

1. **Bail markers upstream.** Extend the review and address stages (both `/ghgremlin` and `/localgremlin` paths) to write a small, well-defined `bail_class` + `bail_detail` into `state.json` when they fail. Vocabulary: `reviewer_requested_changes`, `security`, `secrets`, `other`. Rescue reads these — it does not try to heuristically infer the class from raw logs.

2. **Attempt counter semantics.** The launcher already maintains `rescue_count` on the gremlin ([`launch.sh:160`](skills/_bg/launch.sh)), incremented during Phase B. Keep that field, and make both interactive and headless rescue respect a hard cap of 3. The cap is shared across modes — an already-rescued-twice gremlin gets one more headless attempt.

3. **`--headless` flag on `gremlins rescue`.** A new opt-in flag on `do_rescue` ([`gremlins.py:441`](skills/gremlins/gremlins.py)) that:
   - reads `bail_class` from state.json and refuses to proceed for excluded classes,
   - checks `rescue_count` against the cap before running,
   - runs Phase A with `claude -p` configured for non-interactive execution (no TTY, deterministic completion signal, bounded timeout), and
   - on Phase A success, hands off to `launch.sh --resume` exactly as today.

4. **Bail semantics.** When headless rescue declines — excluded class, cap exhausted, Phase A failure — it writes `bail_reason` to `state.json`, exits non-zero, and leaves the gremlin in a terminal state so `/gremlins` listings show it isn't coming back. Also surface `bail_reason` in the `/gremlins` drill-in view ([`gremlins.py` detail printer](skills/gremlins/gremlins.py)).

### Why this shape

- **Upstream markers, not log parsing.** The spec is explicit: rescue must not heuristically detect excluded classes from raw logs. Writing structured markers at the source (the stage that knows *why* it failed) is the only safe read path.
- **Reuse `rescue_count`.** The field already exists and persists across launcher resumes. Adding a second counter would create ambiguity about which one a caller should trust.
- **Opt-in, not a new subcommand.** A flag on the existing `rescue` keeps the discovery surface small and makes it easy to document alongside interactive mode.

### Upstream bail markers (stage-side)

- [ ] Define the bail-marker vocabulary and JSON shape in `state.json`: a `bail_class` enum (`reviewer_requested_changes` | `security` | `secrets` | `other`) and a free-form `bail_detail` string. Document the shape inline where state.json fields are enumerated.
- [ ] `/ghreview` (`skills/ghreview/SKILL.md`): when a review flags blocking issues, write `bail_class=reviewer_requested_changes` before the skill exits non-zero. If the review labels anything security-related, write `bail_class=security` instead.
- [ ] `/ghaddress` (`skills/ghaddress/SKILL.md`): when address-code cannot proceed because a comment touches secrets, write `bail_class=secrets`. When it bails for a generic reason, write `bail_class=other` with a short `bail_detail`.
- [ ] Local gremlin `review-code` and `address-code` stages in [`skills/localgremlin/localgremlin.py`](skills/localgremlin/localgremlin.py): same classification on failure, written to state.json before the subprocess returns.
- [ ] Add a tiny shared helper (sh or py) that stages call to write `{bail_class, bail_detail}` to state.json atomically — avoid each stage hand-rolling jq patches.

### Attempt cap

- [ ] Confirm `rescue_count` is the right field (currently incremented by `launch.sh` on resume) and document it as the shared counter for both modes.
- [ ] In `do_rescue` (both interactive and headless), check `rescue_count >= 3` *before* running Phase A. On exhaustion: interactive prints an error and exits non-zero; headless writes `bail_reason=attempts_exhausted` and exits non-zero.
- [ ] Decide: does a Phase A failure that doesn't reach Phase B count? (Today `rescue_count` only increments when `launch.sh --resume` fires, i.e. on Phase A success.) Proposal: keep that semantic — only successful relaunches count — so a user isn't penalized for a genuinely broken rescue agent. Flag in Open questions.

### Headless flag

- [ ] Add `--headless` to the `rescue` subcommand argument parser in `gremlins.py`; thread a `headless: bool` param into `do_rescue`.
- [ ] When `headless=True`, before spawning `claude -p`: read `bail_class` from state.json. If it's in `{reviewer_requested_changes, security, secrets}`, write `bail_reason=excluded_class:<class>` and exit non-zero without running Phase A.
- [ ] Switch the Phase A invocation path for headless:
  - Run `claude -p` with the output captured (no TTY), a bounded wall-clock timeout, and stdin closed.
  - Replace the user's Ctrl-C-based abort with a structured success signal. Simplest: require the rescue prompt to write a deterministic marker file (e.g. `artifacts/rescue-<ts>.done` with a `{status: "fixed" | "unsalvageable" | "transient"}` body) and have headless rescue read that file to decide whether to proceed to Phase B, bail, or relaunch-only.
  - On `status=transient`, skip the fix check and go straight to Phase B (the spec explicitly allows this and counts it as an attempt).
  - On `status=unsalvageable` or timeout or non-zero exit, write `bail_reason` and exit.
- [ ] Extend `build_rescue_prompt` ([`gremlins.py:386`](skills/gremlins/gremlins.py)) with a headless-specific tail that instructs the agent to produce the marker file and forbids interactive prompts. Keep the interactive prompt unchanged.

### Bail semantics + visibility

- [ ] On any headless bail path, write `bail_reason` (short, machine-readable string) plus `bail_detail` (human string) to `state.json`, set the gremlin to a terminal status so liveness classifies it as `dead:finished` (or an equivalent terminal state — verify with `liveness_of_state_file`), and exit non-zero.
- [ ] Update the `/gremlins <id>` drill-in printer to show `bail_reason` / `bail_detail` when present.
- [ ] Update `skills/gremlins/SKILL.md` to document `--headless`, the bail-reason field, and the attempt cap. Update the `rescue` bullet so the behavior-difference is clear.

### Tests / verification

- [ ] Manual: trigger a fake failure in a local gremlin, run `/gremlins rescue --headless <id>` from a script with stdin closed, confirm it completes without TTY interaction.
- [ ] Manual: construct state.json with `bail_class=reviewer_requested_changes`, run `/gremlins rescue --headless <id>`, confirm it bails with `bail_reason=excluded_class:reviewer_requested_changes` and exit code non-zero.
- [ ] Manual: set `rescue_count=3`, run headless rescue, confirm it bails with `bail_reason=attempts_exhausted`.
- [ ] Manual: run interactive rescue on the same gremlin that was headless-bailed; confirm the interactive flow still works and can proceed past a cleared `bail_reason` if the user chooses (or document the refusal if we decide interactive should respect the cap too).

## Open questions

- **Rescue-count increment timing.** Does a Phase A failure count against the cap, or only a Phase B relaunch? Proposal above is "only successful relaunches count," which matches today's `launch.sh` behavior. Confirm before implementing.
- **Marker-file vs. stdout contract for Phase A.** The marker-file approach is robust but adds an on-disk artifact. Alternative: require the agent to emit a final JSON line and parse stdout. Marker file is less likely to drift; stdout parsing is simpler. Lean toward marker file unless there's a reason to prefer stdout.
- **`claude -p` headless flags.** Need to confirm exactly which flags make `claude -p` run fully non-interactively (stdin closed + `--output-format text` + bounded timeout may be sufficient, but verify no prompt for permissions pops up mid-run).
- **Does interactive rescue also respect the 3-attempt cap?** The spec says "shared across interactive and headless rescues" for the counter, but is silent on whether *interactive* callers are refused at cap. Reasonable default: interactive warns loudly but allows the human to override; headless hard-refuses. Confirm.
- **Terminal state after a bail.** `dead:finished` currently implies success. We may need a distinct terminal liveness (e.g. `dead:bailed`) so `/gremlins` listings don't mislead. Decide during implementation.
- **Does the bossgremlin work land in this issue or follow-up?** Reading the spec, bossgremlin is a *user* of headless rescue, not a deliverable of this issue. Treat as out of scope unless told otherwise.

xbrianh:
## Summary

`/gremlins rescue --headless <id>` runs the existing two-phase diagnose-then-resume flow end-to-end with no TTY and no user input — for unattended callers (humans walking away, future bossgremlin) that need to recover a failed gremlin without human presence. Default (interactive) rescue is unchanged.

### Headless flow

- **Pre-flight refusals.** Reads `bail_class` from `state.json`. If it's `reviewer_requested_changes`, `security`, or `secrets`, hard-refuses, writes `bail_reason=excluded_class:<class>` plus `bail_detail`, touches the `finished` marker, exits non-zero.
- **Attempt cap.** Shares `rescue_count` with interactive rescue; at ≥ 3, hard-refuses with `bail_reason=attempts_exhausted`. Interactive warns and proceeds.
- **Phase A contract.** `claude -p` with stdin closed, `bypassPermissions`, `--output-format text`, wall-clock timeout (default 1800s, override via `HEADLESS_RESCUE_TIMEOUT_SECS`). Agent writes `artifacts/rescue-<ts>.done` containing `{"status": "fixed" | "transient" | "unsalvageable", "summary": "..."}`. The wrapper reads it deterministically — no log scraping.
- **Bail visibility.** Every refusal/failure writes `bail_reason` (machine) and `bail_detail` (human) to state.json; liveness gains `dead:bailed:<reason>` (both Python and shell paths); the drill-in view surfaces a dedicated `bail:` block above the raw state dump.

### Upstream bail markers

- New `skills/_bg/set-bail.sh` — atomic helper for stages to write `bail_class` + `bail_detail`. Vocabulary: `reviewer_requested_changes`, `security`, `secrets`, `other`. Only `other` is attempted by headless rescue.
- `/ghreview` and `/ghaddress` get gremlin-only bail sections in their SKILL.md (gated on `$GR_ID` so direct human invocations are unchanged).
- `ghgremlin.sh` runs `check_bail` after `/ghreview` and `/ghaddress` so a bail marker halts the pipeline cleanly.
- `localgremlin.py` adds an `emit_bail` helper and wraps the review-code/address-code stages so claude -p crashes record `bail_class=other` before the SystemExit propagates; the address-code prompt grows a gremlin-only bail-instructions section.

### Open-question decisions taken

- Phase A failures don't increment `rescue_count` — only successful Phase B relaunches do (preserves `launch.sh`'s existing semantics; users aren't penalized for a broken rescue agent).
- Marker file (not stdout JSON) — robust against truncation, doesn't need parser drift across model versions.
- Interactive rescue at cap warns and proceeds; headless hard-refuses.
- New terminal liveness `dead:bailed:<reason>` rather than overloading `dead:finished`.
- Bossgremlin is out of scope (it's the user, not the deliverable).

Closes #35

## Test plan

- [x] `gremlins.py rescue --headless <bogus-id>` → "no gremlin matched", exit 1
- [x] state.json with `bail_class=reviewer_requested_changes` → headless refuses, writes `bail_reason=excluded_class:reviewer_requested_changes`, exits non-zero
- [x] state.json with `rescue_count=3` → headless refuses, writes `bail_reason=attempts_exhausted`, exits non-zero
- [x] Drill-in view shows the `bail:` block with class/reason/detail above the raw state dump
- [x] Default listing shows `dead:bailed:<reason>` in the LIVENESS column
- [x] `set-bail.sh` writes/updates `bail_class` and `bail_detail` atomically (with and without detail)
- [x] All shell scripts pass `bash -n`; both Python files pass `ast.parse`
- [ ] End-to-end: run a real failing gremlin, invoke headless rescue from a script with stdin closed, confirm Phase A → marker → Phase B handoff (deferred — needs a real gremlin pipeline)

### DIFF
--- a/skills/_bg/liveness.sh
+++ b/skills/_bg/liveness.sh
@@ -40,7 +40,7 @@ liveness_of_state_file() {
     IFS=$'\x1f' read -r gr_status gr_pid gr_exit_code gr_bail_reason < <(
         jq -r '[.status, (.pid // "" | tostring),
                 (.exit_code // "" | tostring),
-                (.bail_reason // "")] | join("")' "$sf" 2>/dev/null || true
+                (.bail_reason // "")] | join("\u001f")' "$sf" 2>/dev/null || true
     )
 
     # Terminal: finish.sh (or headless rescue's bail path) wrote the
--- a/skills/_bg/liveness.sh
+++ b/skills/_bg/liveness.sh
@@ -31,20 +31,26 @@ liveness_of_state_file() {
     [[ -f "$sf" ]] || return 0
     # Note: avoid `status` as a local name — it's a special/readonly in zsh,
     # and this file is intended to be sourced from either bash or zsh hooks.
-    local wdir gr_status gr_pid gr_exit_code
+    local wdir gr_status gr_pid gr_exit_code gr_bail_reason
     wdir=$(dirname "$sf")
 
     # US (\x1f) separator, matching session-summary.sh and gremlins.py: bash
     # treats tab as IFS-whitespace and collapses consecutive empty columns, so
-    # a future 4th field could silently lose a value. US is non-whitespace.
-    IFS=$'\x1f' read -r gr_status gr_pid gr_exit_code < <(
+    # a future 5th field could silently lose a value. US is non-whitespace.
+    IFS=$'\x1f' read -r gr_status gr_pid gr_exit_code gr_bail_reason < <(
         jq -r '[.status, (.pid // "" | tostring),
-                (.exit_code // "" | tostring)] | join("\u001f")' "$sf" 2>/dev/null || true
+                (.exit_code // "" | tostring),
+                (.bail_reason // "")] | join("")' "$sf" 2>/dev/null || true
     )
 
-    # Terminal: finish.sh ran → `finished` marker exists.
+    # Terminal: finish.sh (or headless rescue's bail path) wrote the
+    # `finished` marker. A bail_reason takes precedence over the generic
+    # exit code so listings show *why* rescue gave up rather than just
+    # "dead:exit 2".
     if [[ -f "$wdir/finished" ]]; then
-        if [[ -n "$gr_exit_code" && "$gr_exit_code" != "0" && "$gr_exit_code" != "null" ]]; then
+        if [[ -n "$gr_bail_reason" ]]; then
+            echo "dead:bailed:$gr_bail_reason"
+        elif [[ -n "$gr_exit_code" && "$gr_exit_code" != "0" && "$gr_exit_code" != "null" ]]; then
             echo "dead:exit $gr_exit_code"
         else
             echo "dead:finished"
--- a/skills/_bg/set-bail.sh
+++ b/skills/_bg/set-bail.sh
@@ -35,7 +35,7 @@ TMP="$STATE_FILE.bail.tmp.$$"
 if jq --arg cls "$BAIL_CLASS" \
       --arg det "$BAIL_DETAIL" \
       '.bail_class = $cls
-       | (if $det == "" then . else .bail_detail = $det end)' \
+       | (if $det == "" then del(.bail_detail) else .bail_detail = $det end)' \
       "$STATE_FILE" > "$TMP" 2>/dev/null; then
     mv "$TMP" "$STATE_FILE" 2>/dev/null || rm -f "$TMP" 2>/dev/null
 else
--- a/skills/_bg/set-bail.sh
+++ b/skills/_bg/set-bail.sh
@@ -1,0 +1,44 @@
+#!/usr/bin/env bash
+# Atomically write bail_class (and optional bail_detail) to a background
+# gremlin's state.json. Stages — both shell (ghgremlin.sh, /ghreview,
+# /ghaddress) and python (localgremlin.py) — call this when they decline
+# to proceed so /gremlins rescue --headless can decide whether to attempt
+# recovery.
+#
+# Usage: set-bail.sh <gr_id> <bail_class> [bail_detail]
+#
+# Vocabulary for bail_class:
+#   reviewer_requested_changes  — code review flagged blocker findings
+#   security                    — review flagged security concern(s)
+#   secrets                     — change touches secrets/credentials
+#   other                       — generic; pair with a useful bail_detail
+#
+# Headless rescue refuses to run for the first three classes. `other` is
+# attempted with the full diagnose-and-fix path.
+#
+# Fails silently on any error: writing the marker must never break the
+# stage that's bailing.
+set -u
+
+GR_ID="${1:-}"
+BAIL_CLASS="${2:-}"
+BAIL_DETAIL="${3:-}"
+
+[[ -n "$GR_ID" && -n "$BAIL_CLASS" ]] || exit 0
+
+STATE_FILE="${XDG_STATE_HOME:-$HOME/.local/state}/claude-gremlins/$GR_ID/state.json"
+[[ -f "$STATE_FILE" ]] || exit 0
+
+command -v jq >/dev/null 2>&1 || exit 0
+
+TMP="$STATE_FILE.bail.tmp.$$"
+if jq --arg cls "$BAIL_CLASS" \
+      --arg det "$BAIL_DETAIL" \
+      '.bail_class = $cls
+       | (if $det == "" then . else .bail_detail = $det end)' \
+      "$STATE_FILE" > "$TMP" 2>/dev/null; then
+    mv "$TMP" "$STATE_FILE" 2>/dev/null || rm -f "$TMP" 2>/dev/null
+else
+    rm -f "$TMP" 2>/dev/null
+fi
+exit 0
--- a/skills/ghgremlin/ghgremlin.sh
+++ b/skills/ghgremlin/ghgremlin.sh
@@ -106,6 +106,20 @@ patch_state() {
   fi
 }
 
+# check_bail <stage-label> — if the just-completed stage wrote a bail_class
+# into state.json (via _bg/set-bail.sh), die so the gremlin pipeline halts
+# cleanly. /gremlins rescue --headless then reads bail_class to decide
+# whether to attempt automated recovery; for excluded classes (security,
+# secrets, reviewer_requested_changes) it refuses outright.
+check_bail() {
+  [[ -n "$STATE_FILE" && -f "$STATE_FILE" ]] || return 0
+  local label="${1:-stage}" cls
+  cls=$(jq -r '.bail_class // ""' "$STATE_FILE" 2>/dev/null)
+  if [[ -n "$cls" ]]; then
+    die "$label bailed: bail_class=$cls (see state.json bail_detail)"
+  fi
+}
+
 command -v claude >/dev/null || die "claude CLI not found"
 command -v gh >/dev/null     || die "gh CLI not found"
 command -v jq >/dev/null     || die "jq not found"
@@ -331,6 +345,7 @@ If the diff is scoped correctly, write exactly: 'Scoped correctly — nothing to
 
   wait "$pid_ghreview" || { kill "$pid_scope" 2>/dev/null; wait "$pid_scope" 2>/dev/null; die "/ghreview failed"; }
   wait "$pid_scope"    || die "scope reviewer failed"
+  check_bail "/ghreview"
 fi
 
 if run_stage wait-copilot; then
@@ -358,6 +373,7 @@ if run_stage ghaddress; then
   set_stage ghaddress
   echo "==> [6/6] running /ghaddress"
   claude -p "${CLAUDE_FLAGS[@]}" "/ghaddress $PR_URL" | progress_tee >/dev/null
+  check_bail "/ghaddress"
 fi
 
 echo ""
--- a/skills/gremlins/SKILL.md
+++ b/skills/gremlins/SKILL.md
@@ -1,7 +1,7 @@
 ---
 name: gremlins
 description: On-demand status of background gremlins launched by /localgremlin and /ghgremlin. Reads every ~/.local/state/claude-gremlins/<id>/state.json on the machine and prints one line per active gremlin with its kind, current stage, liveness (running / stalled / dead), description, and age. Use to check progress, spot crashed gremlins, close finished ones, stop a running gremlin, rescue a dead/stalled one, or land a finished gremlin onto its target branch. Not a project filter by default — set --here to restrict to the current repo.
-argument-hint: [stop|rescue|rm|close|land <id>] [--here] [--running] [--dead] [--stalled] [--kind local|gh] [--since <dur>] [--recent [N]] [--watch [sec]] [<id-prefix>]
+argument-hint: [stop|rescue [--headless]|rm|close|land <id>] [--here] [--running] [--dead] [--stalled] [--kind local|gh] [--since <dur>] [--recent [N]] [--watch [sec]] [<id-prefix>]
 allowed-tools: Bash(~/.claude/skills/gremlins/gremlins.py:*)
 ---
 
@@ -20,7 +20,7 @@ The script produces a small table. Each row is one gremlin:
 - `KIND` — `local` (from `/localgremlin`) or `gh` (from `/ghgremlin`).
 - `ID` — short (6-hex) gremlin identifier. Use it with `close <id>` to mark a dead gremlin as closed (which hides it from future runs of this command).
 - `STAGE` — the gremlin's current stage name (e.g. `plan`, `implement`, `review-code`). For parallel reviewers, a parenthesized sub-stage shows each reviewer's state (e.g. `review-code (opus=done,sonnet=running)`).
-- `LIVENESS` — `running`, `stalled:<reason>`, or `dead:<reason>`.
+- `LIVENESS` — `running`, `stalled:<reason>`, `dead:<reason>`, or `dead:bailed:<bail-reason>` (set when `/gremlins rescue --headless` declined to proceed; see `rescue` below).
 - `AGE` — time since launch.
 - `DESCRIPTION` — the short human phrase captured at launch.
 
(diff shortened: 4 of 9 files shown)


xbrianh (commented):

skills/ghgremlin/ghgremlin.sh
```diff
@@ -106,6 +106,20 @@ patch_state() {
...
 
+# check_bail <stage-label> — if the just-completed stage wrote a bail_class
+# into state.json (via _bg/set-bail.sh), die so the gremlin pipeline halts
+# cleanly. /gremlins rescue --headless then reads bail_class to decide
+# whether to attempt automated recovery; for excluded classes (security,
+# secrets, reviewer_requested_changes) it refuses outright.
+check_bail() {
+  [[ -n "$STATE_FILE" && -f "$STATE_FILE" ]] || return 0
+  local label="${1:-stage}" cls
+  cls=$(jq -r '.bail_class // ""' "$STATE_FILE" 2>/dev/null)
```
xbrianh, line 117: **Stale `bail_class` will false-fire on resume.**

`launch.sh --resume` only deletes `exit_code`, `ended_at`, `sub_stage`, `stage_updated_at` (it doesn't touch the new `bail_class` / `bail_reason` / `bail_detail` this PR introduces). So consider this flow:

1. `/ghreview` writes `bail_class=security`, `check_bail` dies.
2. User fixes the finding manually, runs interactive `rescue` (which doesn't filter on excluded classes).
3. Phase A exits 0, Phase B → `launch.sh --resume` → pipeline resumes at `ghreview`.
4. Resumed `/ghreview` finds nothing, writes no bail marker.
5. `check_bail "/ghreview"` reads `.bail_class` — still `"security"` from step 1 — and dies again.

The same staleness affects the Python path: `_write_bail` writes `bail_reason`, `launch.sh --resume` doesn't clear it, so after a successful rescue + clean finish, `liveness_of_state_file` sees `bail_reason` is set, `finished` marker is present, and returns `dead:bailed:<stale>` instead of `dead:finished` (bail_reason takes precedence over `exit_code=0`).

Fix is probably to extend the resume jq patch in `launch.sh`:

```
| del(.bail_class)
| del(.bail_reason)
| del(.bail_detail)
```

The docstring on `_write_bail` also says "failing to write either piece leaves the gremlin in its prior state, which is no worse than before headless rescue ran" — but resuming *is* worse than before, because of the above.

skills/_bg/liveness.sh
```diff
@@ -31,20 +31,26 @@ liveness_of_state_file() {
...
     # US (\x1f) separator, matching session-summary.sh and gremlins.py: bash
     # treats tab as IFS-whitespace and collapses consecutive empty columns, so
-    # a future 4th field could silently lose a value. US is non-whitespace.
-    IFS=$'\x1f' read -r gr_status gr_pid gr_exit_code < <(
+    # a future 5th field could silently lose a value. US is non-whitespace.
+    IFS=$'\x1f' read -r gr_status gr_pid gr_exit_code gr_bail_reason < <(
         jq -r '[.status, (.pid // "" | tostring),
-                (.exit_code // "" | tostring)] | join("\u001f")' "$sf" 2>/dev/null || true
+                (.exit_code // "" | tostring),
+                (.bail_reason // "")] | join("")' "$sf" 2>/dev/null || true
```
xbrianh: The jq separator here is a literal `\x1f` (US) byte embedded inside the quotes, which renders invisibly in most editors/diffs and is easy to strip by accident (paste normalization, reformatters, copy-through-clipboard). The pre-PR version used the jq escape `join("\u001f")`, which is visible and immune to that class of accident. Suggest restoring the escape form:

```
(.bail_reason // "")] | join("\u001f")
```

Functionally equivalent today, but robust against silent corruption.

skills/gremlins/SKILL.md
```diff
@@ -20,7 +20,7 @@ The script produces a small table. Each row is one gremlin:
 - `KIND` — `local` (from `/localgremlin`) or `gh` (from `/ghgremlin`).
 - `ID` — short (6-hex) gremlin identifier. Use it with `close <id>` to mark a dead gremlin as closed (which hides it from future runs of this command).
 - `STAGE` — the gremlin's current stage name (e.g. `plan`, `implement`, `review-code`). For parallel reviewers, a parenthesized sub-stage shows each reviewer's state (e.g. `review-code (opus=done,sonnet=running)`).
-- `LIVENESS` — `running`, `stalled:<reason>`, or `dead:<reason>`.
+- `LIVENESS` — `running`, `stalled:<reason>`, `dead:<reason>`, or `dead:bailed:<bail-reason>` (set when `/gremlins rescue --headless` declined to proceed; see `rescue` below).
```
xbrianh: Minor doc nit: this says `dead:bailed:<bail-reason>` is "set when `/gremlins rescue --headless` declined to proceed." That's accurate for the listing column, but an upstream stage writing `bail_class` via `set-bail.sh` (e.g. `/ghreview` flagging a security blocker) causes `check_bail` to `die`, and the gremlin then shows up as `dead:exit 1` in the listing — users have to drill in (`/gremlins <id>`) to see the `bail:` block explaining *why*. Worth a short sentence here pointing that out, or — better — a follow-up where the LIVENESS column surfaces upstream `bail_class` too (parallel to how it surfaces rescue-set `bail_reason`).

xbrianh (commented):
Nice job on the two-phase headless rescue — the deterministic marker-file contract, the pre-flight bail-class gating, and the parity between Python and shell liveness paths are all well considered.

Main concern: `launch.sh --resume` never clears `bail_class`, `bail_reason`, or `bail_detail`, so the new state survives across rescues. That means (a) `check_bail` will false-fire on a resumed stage that previously wrote `bail_class`, and (b) after a successful rescue + clean finish, `liveness.sh` will mislabel the gremlin as `dead:bailed:<stale-reason>` because bail_reason takes precedence over `exit_code=0`. Smaller notes on an embedded literal US byte in the jq string and a doc gap in the LIVENESS column description.

skills/gremlins/gremlins.py
```diff
@@ -32,6 +32,21 @@
...
 FMT = "%-5s  %-47s  %-22s  %-28s  %-5s  %s"
 
+# Headless rescue caps. The attempt cap is shared across interactive and
+# headless rescues — both check `rescue_count`, but interactive only warns
+# while headless hard-refuses. The wall-clock timeout bounds Phase A so a
+# stuck `claude -p` doesn't hang an unattended caller indefinitely.
+RESCUE_CAP = 3
+HEADLESS_PHASE_A_TIMEOUT_SECS = int(
+    os.environ.get("HEADLESS_RESCUE_TIMEOUT_SECS") or 1800
+)
```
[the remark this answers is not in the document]
xbrianh: Fixed in b8ba583 — wrapped the int(...) parse in try/except (ValueError, TypeError) with a fallback to 1800, so a misconfigured HEADLESS_RESCUE_TIMEOUT_SECS no longer breaks the rest of /gremlins.

skills/gremlins/gremlins.py
```diff
@@ -438,7 +496,123 @@ def build_rescue_prompt(state, wdir, log_tail, artifact_paths):
...
+    try:
+        with open(sf, encoding="utf-8") as fh:
+            state = json.load(fh)
+    except Exception:
+        return False
+    state.update(patch)
+    tmp = sf + ".bail.tmp"
+    try:
+        with open(tmp, "w", encoding="utf-8") as fh:
+            json.dump(state, fh, indent=2)
```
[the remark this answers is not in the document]
xbrianh: Fixed in b8ba583 — temp path is now pid-scoped (`state.json.bail.tmp.<pid>`), matching the `$$`-suffixed pattern used by set-stage.sh / set-bail.sh.

skills/gremlins/gremlins.py
```diff
@@ -438,7 +496,123 @@ def build_rescue_prompt(state, wdir, log_tail, artifact_paths):
...
+        "--permission-mode", "bypassPermissions",
+        "--output-format", "text",
+        prompt,
+    ]
+    try:
+        result = subprocess.run(
+            cmd,
+            cwd=workdir,
+            stdin=subprocess.DEVNULL,
+            capture_output=True,
```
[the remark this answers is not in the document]
xbrianh: Fixed in b8ba583 — stdout is DEVNULL'd and only stderr is captured via stderr=PIPE. Added a comment explaining that the results come from the marker file, not the process output.

skills/_bg/set-bail.sh
```diff
@@ -0,0 +1,44 @@
...
+STATE_FILE="${XDG_STATE_HOME:-$HOME/.local/state}/claude-gremlins/$GR_ID/state.json"
+[[ -f "$STATE_FILE" ]] || exit 0
+
+command -v jq >/dev/null 2>&1 || exit 0
+
+TMP="$STATE_FILE.bail.tmp.$$"
+if jq --arg cls "$BAIL_CLASS" \
+      --arg det "$BAIL_DETAIL" \
+      '.bail_class = $cls
+       | (if $det == "" then . else .bail_detail = $det end)' \
```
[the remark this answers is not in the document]
xbrianh: Fixed in b8ba583 — the empty-detail branch now uses `del(.bail_detail)` so a later bail that updates only the class can't inherit a stale detail.

skills/ghgremlin/ghgremlin.sh
xbrianh: Good catch — fixed in b8ba583. The resume jq patch in launch.sh now also runs `del(.bail_class) | del(.bail_reason) | del(.bail_detail)`, so a resumed stage never sees stale markers and liveness correctly returns `dead:finished` on a clean post-rescue finish. With this in, the `_write_bail` docstring's "no worse than before" claim holds again.

skills/_bg/liveness.sh
xbrianh: Fixed in b8ba583 — restored the visible jq unicode escape (`join("\\u001f")`). Confirmed jq still emits the US byte, and the bash side is unchanged (`IFS=$'\\x1f'`).

skills/gremlins/SKILL.md
xbrianh: Added a short doc note in b8ba583: the LIVENESS column entry now calls out that an upstream `bail_class` halt surfaces as `dead:exit <N>` and directs users to the drill-in view (`/gremlins <id>`) for the `bail:` block. Surfacing upstream bail_class in the column itself is a reasonable follow-up but out of scope for this PR.

PR #40, opened 2026-04-24
Split /localreview and /localaddress out of /localgremlin

### LINKED ISSUE #39: Decompose /localgremlin into /localreview and /localaddress skills
## Context

`/localgremlin` runs plan → implement → review-code (triple-lens parallel) → address-code as a single end-to-end pipeline. There is no way to invoke just the review or address step today — the whole gremlin has to run. The `/gh*` side already composes from standalone skills (`/ghplan`, `/ghreview`, `/ghaddress`); the local side should match so individual phases can be run in isolation against arbitrary local changes.

Ref: [`/tmp/design-local-skills.md`](/tmp/design-local-skills.md) (the finalized design spec).

## Approach

Factor the review-code and address-code stages out of `skills/localgremlin/localgremlin.py` into shared Python functions, then expose them as two thin CLIs (`localreview.py`, `localaddress.py`) backing two new SKILL.md files (`/localreview`, `/localaddress`). `/localgremlin` keeps its current end-to-end behavior by calling the same shared functions directly (no subprocess hop) — this guarantees byte-identical artifacts, stdout, and sub_stage emissions, which the spec requires.

Why this over alternatives:
- **Shared module + direct calls** (chosen): one code path, no subprocess/stream-json re-plumbing inside the orchestrator, and `set-stage`/`set-bail` stay guarded by `GR_ID` so standalone invocations don't try to emit gremlin bookkeeping.
- **Orchestrator shells out to standalone CLIs**: exercises the standalone path on every run but doubles the subprocess layers and makes stdout ordering fragile. Rejected.
- **SKILL.md-only (Claude runs reviews inline, like `/ghreview`)**: breaks the triple-parallel-lens model since Claude can't fan out three independent `claude -p` processes with different `--model` flags from a single skill invocation cleanly. Rejected.

Standalone invocations have no `GR_ID`, so the existing `set_stage` / `emit_bail` no-op paths already cover them. Filenames, lens prompts, and review structure stay byte-identical.

## Tasks

- [ ] Extract shared primitives from `skills/localgremlin/localgremlin.py` into a sibling `_core.py`: `run_claude`, `run_review`, `run_triple_review`, `ReviewWorker`, `log_stream`, `_emit_event`, `_trunc`, child-process tracking + signal handlers, `set_stage`, `emit_bail`, `resolve_session_dir`, `in_git_repo`, `git_head`, `die`, `MODEL_RE`, `GR_ID_RE`, `CLAUDE_FLAGS`, and the lens-file loader.
- [ ] Add a stage-level helper `run_review_code_stage(session_dir, plan_text, holistic, detail, scope, is_git)` in `_core.py` that builds the context (omitting the plan section when `plan_text` is empty), calls `run_triple_review`, and returns the three output paths. Used by both the orchestrator and `localreview.py`.
- [ ] Add a stage-level helper `run_address_code_stage(session_dir, address_model, is_git)` in `_core.py` that locates the three `review-code-{holistic,detail,scope}-*.md` files under `session_dir`, builds the prompt (matching today's verbatim), and calls `run_claude`. Used by both the orchestrator and `localaddress.py`.
- [ ] Refactor `localgremlin.py` to import from `_core.py` and call the two stage helpers. Verify by diffing produced artifacts + stdout + `state.json` sub_stage transitions against a pre-refactor run on a small task — must be identical.
- [ ] Add `skills/localgremlin/localreview.py` CLI: flags `--dir <path>` (default cwd, must exist — no silent mkdir), `--plan <path>` (optional; errors if given but missing/empty), `-a/-b/-c <model>` (same defaults as `localgremlin.py`). In git mode, error cleanly if `HEAD~1..HEAD` plus working-tree diff is empty. Invokes `run_review_code_stage`.
- [ ] Add `skills/localgremlin/localaddress.py` CLI: flags `--dir <path>` (default cwd, must exist), `-x <model>` (same default as `localgremlin.py`). Globs `review-code-{holistic,detail,scope}-*.md`; errors naming which lens is missing if any glob returns zero (or >1) matches. Invokes `run_address_code_stage`.
- [ ] Add `skills/localreview/SKILL.md`: argument-hint `[--dir <path>] [--plan <path>] [-a <model>] [-b <model>] [-c <model>]`; allowed-tools restricted to the script invocation; body explains it runs foreground (not background), writes review files to `--dir`, and points at the three output filenames.
- [ ] Add `skills/localaddress/SKILL.md`: argument-hint `[--dir <path>] [-x <model>]`; allowed-tools restricted to the script invocation; body explains it reads the three review files from `--dir`, fixes actionable findings, and in a git repo creates one `Address review feedback` commit (no push).
- [ ] Update `CLAUDE.md` and `README.md`: add one-line entries for `/localreview` and `/localaddress` under "Additional skills", and note the `_core.py` split in the "The local gremlin" section. Keep pointers accurate.
- [ ] Run `scripts/sync.sh push` once the skill dirs are populated so `~/.claude/skills/{localreview,localaddress}/` mirror the repo.
- [ ] Manual verification against the acceptance criteria in the spec:
  - [ ] `/localreview` against a one-commit-ahead worktree produces the same three filenames + content shape as a same-models `/localgremlin` review-code stage.
  - [ ] `/localaddress --dir <path>` against those three files produces the same fixes + `Address review feedback` commit.
  - [ ] `/localgremlin` end-to-end still produces identical artifacts, `state.json` sub_stage transitions, bail-class plumbing, and `--resume-from` behavior.
  - [ ] `/gremlins rescue` of a failed `/localgremlin` (force a failure in review-code or address-code) still succeeds.

## Open questions

- **Parameter name**: `--dir` per the spec. Alternatives considered and rejected: `--artifacts` (more descriptive but longer), `--session` (overlaps with gremlin "session dir" terminology and could confuse direct-invocation users). Keep `--dir` unless a reviewer pushes back.
- **Subcommand vs separate scripts**: `localreview.py` and `localaddress.py` as siblings of `localgremlin.py` is my default (clear 1:1 with SKILL.md names). Alternative: `localgremlin.py review` / `localgremlin.py address` subcommands. Either is fine — I'll go with separate scripts unless there's a reason not to.
- **Empty-plan rendering**: current review prompt hard-codes `"The plan for this change is:\n\n{plan_text}\n\n{scope}"`. When `plan_text=""` the section becomes whitespace noise. Changing the builder to omit the plan block when empty is a tiny behavior delta from today's `/localgremlin` (which always has a plan) — acceptable, I think, but flagging since the spec said "byte-identical" for the with-plan case.
- **Non-git `/localreview` scope**: spec says "dirty / recently-modified files in cwd" but standalone has no pre-impl sentinel to anchor "recent" against. Today's non-git review just tells the model to `git diff if available, otherwise inspect recently modified files` — I plan to reuse that exact prompt; standalone users outside git just need to have a reasonable-looking tree. Flag if that's too hand-wavy.
- **`/localplan` symmetry**: out of scope for this issue per the spec; mentioning only so it doesn't get silently pulled in.

xbrianh:
## Summary

- Factor review-code and address-code stage bodies out of `skills/localgremlin/localgremlin.py` into a sibling `_core.py`, sharing them with two new standalone skills.
- Add `/localreview` (foreground triple-lens parallel review over local changes) and `/localaddress` (foreground address-stage over existing `review-code-*.md` files).
- `/localgremlin` keeps its end-to-end behavior by calling the shared `run_review_code_stage` / `run_address_code_stage` helpers directly — no subprocess hop, byte-identical artifacts/stdout/sub_stage emissions for the with-plan case.

## Behavior notes

- Standalone invocations have no `GR_ID`, so `set_stage` / `emit_bail` stay no-ops outside a gremlin.
- Filenames, lens prompts, and review structure are unchanged.
- When `/localreview` is run without `--plan`, the plan block is omitted from the review prompt (vs. rendering an empty plan section). The orchestrator path always has a plan, so it is unaffected.
- `/localaddress` extracts model names from the `review-code-<lens>-<model>.md` filenames, so the address prompt's per-lens model labels match the files that were actually produced (for both orchestrator and standalone callers).

## Test plan

- [ ] `/localreview` against a one-commit-ahead worktree produces the same three filenames + content shape as a same-models `/localgremlin` review-code stage.
- [ ] `/localaddress --dir <path>` against those three files produces the same fixes + `Address review feedback` commit as `/localgremlin`'s address-code stage.
- [ ] `/localgremlin` end-to-end still produces identical artifacts, `state.json` sub_stage transitions, bail-class plumbing, and `--resume-from` behavior.
- [ ] `/gremlins rescue` of a failed `/localgremlin` (force a failure in review-code or address-code) still succeeds.
- [ ] Post-merge: run `scripts/sync.sh push` from the main checkout so `~/.claude/skills/{localreview,localaddress}/` mirror the repo.

Closes #39

### DIFF
--- a/skills/localgremlin/_core.py
+++ b/skills/localgremlin/_core.py
@@ -526,10 +526,34 @@ def run_review_code_stage(
     """Execute the review-code stage: load lenses, fan out three reviewers,
     and return the three output paths. Emits bail_class=other on failure
     when running under a gremlin (no-op otherwise). Shared by the
-    orchestrator and /localreview."""
+    orchestrator and /localreview.
+
+    Passing ``plan_text=""`` (empty string, not None) intentionally omits
+    the plan block from the review prompt entirely — this is the contract
+    that lets standalone ``/localreview`` callers run without ``--plan``.
+    Any non-empty ``plan_text`` (even whitespace-only) takes the with-plan
+    branch and is rendered verbatim into the prompt.
+
+    Stale ``review-code-<lens>-*.md`` files for each lens are unlinked
+    before spawning the reviewers so a ``--resume-from review-code`` with
+    different ``-a/-b/-c`` models cannot leave the directory with two
+    files for the same lens (which would later confuse
+    ``run_address_code_stage``'s glob-based discovery).
+    """
     review_code_a = session_dir / f"review-code-holistic-{holistic}.md"
     review_code_b = session_dir / f"review-code-detail-{detail}.md"
     review_code_c = session_dir / f"review-code-scope-{scope}.md"
+
+    # Clean up stale per-lens review files from a previous run with
+    # different reviewer models. Without this, --resume-from review-code
+    # with changed -a/-b/-c would leave two files for the same lens and
+    # break run_address_code_stage's uniqueness check.
+    for lens in ("holistic", "detail", "scope"):
+        for stale in session_dir.glob(f"review-code-{lens}-*.md"):
+            try:
+                stale.unlink()
+            except OSError:
+                pass
 
     focus_a, focus_b, focus_c = load_lenses()
 
@@ -582,51 +606,68 @@ def run_address_code_stage(
 ) -> None:
     """Execute the address-code stage: glob for the three review files,
     build the address prompt, and invoke claude. Emits bail_class=other
-    on failure when running under a gremlin (no-op otherwise). Shared by
+    on failure when running under a gremlin (no-op otherwise) — including
+    failures during glob/validation before claude is spawned. Shared by
     the orchestrator and /localaddress."""
-    review_files = {}
-    for lens in ("holistic", "detail", "scope"):
-        matches = sorted(glob.glob(str(session_dir / f"review-code-{lens}-*.md")))
-        if not matches:
-            die(f"no review-code-{lens}-*.md file found in {session_dir}")
-        if len(matches) > 1:
-            die(
-                f"multiple review-code-{lens}-*.md files in {session_dir}: "
-                f"{', '.join(matches)}"
+    # Outer try/except so *any* stage failure (missing or ambiguous review
+    # files, invalid model in a filename, read errors, or the claude -p
+    # subprocess itself) records a bail marker before SystemExit
+    # propagates. Without this wrapping the pre-claude failure paths would
+    # exit via die() without ever calling emit_bail, and headless rescue
+    # would have no bail_class to act on.
+    try:
+        review_files = {}
+        for lens in ("holistic", "detail", "scope"):
+            matches = sorted(glob.glob(str(session_dir / f"review-code-{lens}-*.md")))
+            if not matches:
+                die(f"no review-code-{lens}-*.md file found in {session_dir}")
+            if len(matches) > 1:
+                die(
+                    f"multiple review-code-{lens}-*.md files in {session_dir}: "
+                    f"{', '.join(matches)}"
+                )
+            review_files[lens] = pathlib.Path(matches[0])
+
+        # Model names are embedded in the filenames as the suffix between
+        # the lens tag and the `.md` extension. Extracting them here keeps
+        # the address prompt's per-lens model labels in sync with the files
+        # that were actually produced, for both orchestrator and standalone
+        # callers. Validate against MODEL_RE so a malformed filename (e.g.
+        # review-code-holistic-.md, or one with unexpected characters)
+        # fails loudly instead of producing an empty/garbled prompt label.
+        def _model_from(path: pathlib.Path, lens: str) -> str:
+            stem = path.stem  # review-code-<lens>-<model>
+            prefix = f"review-code-{lens}-"
+            model = stem[len(prefix):] if stem.startswith(prefix) else ""
+            if not model or not MODEL_RE.match(model):
+                die(
+                    f"cannot extract a valid model name from review file: "
+                    f"{path.name}"
+                )
+            return model
+
+        model_a = _model_from(review_files["holistic"], "holistic")
+        model_b = _model_from(review_files["detail"], "detail")
+        model_c = _model_from(review_files["scope"], "scope")
+
+        text_a = review_files["holistic"].read_text(encoding="utf-8")
+        text_b = review_files["detail"].read_text(encoding="utf-8")
+        text_c = review_files["scope"].read_text(encoding="utf-8")
+
+        address_commit_instr = ""
+        if is_git:
+            address_commit_instr = (
+                "After making all fixes, stage the changed files by name and "
+                "create a single git commit titled 'Address review feedback' whose "
+                "body references all three review files. Do not push."
             )
-        review_files[lens] = pathlib.Path(matches[0])
 
-    # Model names are embedded in the filenames as the suffix between the
-    # lens tag and the `.md` extension. Extracting them here keeps the
-    # address prompt's per-lens model labels in sync with the files that
-    # were actually produced, for both orchestrator and standalone callers.
-    def _model_from(path: pathlib.Path, lens: str) -> str:
-        stem = path.stem  # review-code-<lens>-<model>
-        prefix = f"review-code-{lens}-"
-        return stem[len(prefix):] if stem.startswith(prefix) else stem
-
-    model_a = _model_from(review_files["holistic"], "holistic")
-    model_b = _model_from(review_files["detail"], "detail")
-    model_c = _model_from(review_files["scope"], "scope")
-
-    text_a = review_files["holistic"].read_text(encoding="utf-8")
-    text_b = review_files["detail"].read_text(encoding="utf-8")
-    text_c = review_files["scope"].read_text(encoding="utf-8")
-
-    address_commit_instr = ""
-    if is_git:
-        address_commit_instr = (
-            "After making all fixes, stage the changed files by name and "
-            "create a single git commit titled 'Address review feedback' whose "
-            "body references all three review files. Do not push."
-        )
-
-    # Only attached when running under a gremlin so direct invocations
-    # (no GR_ID) don't see prompt instructions for a helper they can't
-    # usefully invoke.
-    bail_section = ""
-    if os.environ.get("GR_ID"):
-        bail_section = """
+        # Only attached when running under a gremlin so direct invocations
+        # (no GR_ID) don't see prompt instructions for a helper they can't
+        # usefully invoke.
+        bail_section = ""
+        if os.environ.get("GR_ID"):
+            bail_section = """
 
 If a finding asks you to change something that touches secrets/credentials, or you decline to address one or more findings for any other reason that should halt automated recovery, run the bail helper before finishing:
   - `~/.claude/skills/_bg/set-bail.sh "$GR_ID" secrets "<one-line reason>"` if the blocked finding touches secrets.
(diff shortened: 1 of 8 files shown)


xbrianh (commented):

skills/localgremlin/localreview.py
```diff
@@ -0,0 +1,105 @@
...
+            has_commit_diff = subprocess.run(
+                ["git", "diff", "--quiet", "HEAD~1", "HEAD"],
+                check=False,
+            ).returncode != 0
+        has_dirty = bool(subprocess.run(
+            ["git", "status", "--porcelain"],
+            capture_output=True, text=True, check=False,
+        ).stdout.strip())
+        if not has_commit_diff and not has_dirty:
+            die("nothing to review: HEAD~1..HEAD has no changes and working tree is clean")
```
xbrianh, line 96: The error message is slightly misleading in the initial-commit case: if `HEAD~1` doesn't exist, `head1_exists` is false and `has_commit_diff` stays false, so reaching this branch with a clean tree emits `HEAD~1..HEAD has no changes and working tree is clean` — but the real reason is that `HEAD~1` doesn't exist yet. Consider branching the message on `head1_exists` so the initial-commit-only case says something like `no commit history beyond HEAD and working tree is clean` instead.

skills/localgremlin/_core.py
```diff
@@ -0,0 +1,666 @@
...
+    review_files = {}
+    for lens in ("holistic", "detail", "scope"):
+        matches = sorted(glob.glob(str(session_dir / f"review-code-{lens}-*.md")))
+        if not matches:
+            die(f"no review-code-{lens}-*.md file found in {session_dir}")
+        if len(matches) > 1:
+            die(
+                f"multiple review-code-{lens}-*.md files in {session_dir}: "
+                f"{', '.join(matches)}"
+            )
```
xbrianh, line 663: This glob-based discovery changes behavior vs. the orchestrator's previous pre-computed-path approach. Scenario: a run with `-a opus` writes `review-code-holistic-opus.md`, the stage fails, then the user `--resume-from review-code` with `-a sonnet` — review-code produces `review-code-holistic-sonnet.md` but doesn't clean up the opus file, so the next stage's glob finds two matches for the `holistic` lens and dies with `multiple review-code-holistic-*.md files`. Previously this would have silently used the sonnet file and ignored opus.

This is arguably a safer failure mode, but the PR body's `byte-identical artifacts/stdout/sub_stage emissions` claim isn't quite true for the resume-with-different-models path. Either document that `--resume-from review-code` with different reviewer models requires manually removing stale files, or have review-code unlink prior `review-code-<lens>-*.md` before writing the new ones.

skills/localgremlin/_core.py
```diff
@@ -0,0 +1,666 @@
...
+        )
+
+    # Only attached when running under a gremlin so direct invocations
+    # (no GR_ID) don't see prompt instructions for a helper they can't
+    # usefully invoke.
+    bail_section = ""
+    if os.environ.get("GR_ID"):
+        bail_section = """
+
+If a finding asks you to change something that touches secrets/credentials, or you decline to address one or more findings for any other reason that should halt automated recovery, run the bail helper before finishing:
```
xbrianh, line 672: The `plan_text=""` → omit-plan-block behavior is the key contract that lets `/localreview` run without `--plan`, but it's only mentioned in the inline comment at line 547-549. Worth promoting into the docstring so future callers know the empty-string escape hatch is intentional (and that falsy non-empty plan content would still hit the with-plan branch).

xbrianh (commented):
Solid refactor — the three call sites share one code path via `_core.py`, and the orchestrator's behavior appears preserved on the happy path. A few small observations:

1. One subtle behavior deviation: `run_address_code_stage` discovers review files via glob and dies on multiple matches. Previously the orchestrator read them at pre-computed paths, so stale files from a prior run with different `-a/-b/-c` models were silently ignored. The new behavior is arguably better (fails loudly on ambiguous state) but narrows the byte-identical parity claim slightly.
2. `/localreview`'s empty-diff error message always says `HEAD~1..HEAD has no changes` even when `HEAD~1` doesn't exist (initial commit).
3. `run_review_code_stage`'s docstring could note that `plan_text=""` omits the plan block entirely — worth documenting at the function boundary since it's the contract that lets standalone callers skip `--plan`.

No blockers.

skills/localgremlin/localreview.py
```diff
@@ -0,0 +1,105 @@
...
+        die(f"--dir does not exist: {session_dir}")
+
+    plan_text = ""
+    if args.plan is not None:
+        plan_path = pathlib.Path(args.plan)
+        if not plan_path.exists():
+            die(f"--plan does not exist: {plan_path}")
+        if plan_path.stat().st_size == 0:
+            die(f"--plan is empty: {plan_path}")
+        plan_text = plan_path.read_text(encoding="utf-8")
```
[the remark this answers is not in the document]
xbrianh: Applied the suggested validation verbatim — added is_file() check, UnicodeDecodeError and OSError handlers around read_text — in 37792ed.

skills/localgremlin/_core.py
```diff
@@ -0,0 +1,666 @@
...
+    for lens in ("holistic", "detail", "scope"):
+        matches = sorted(glob.glob(str(session_dir / f"review-code-{lens}-*.md")))
+        if not matches:
+            die(f"no review-code-{lens}-*.md file found in {session_dir}")
+        if len(matches) > 1:
+            die(
+                f"multiple review-code-{lens}-*.md files in {session_dir}: "
+                f"{', '.join(matches)}"
+            )
+        review_files[lens] = pathlib.Path(matches[0])
```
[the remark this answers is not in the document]
xbrianh: Fixed in 37792ed: wrapped the glob/validation portion and _model_from() inside the existing outer try/except so any pre-claude failure (missing/ambiguous review files, invalid model in filename, read errors) records bail_class=other before SystemExit propagates.

skills/localgremlin/_core.py
```diff
@@ -0,0 +1,666 @@
...
+    # address prompt's per-lens model labels in sync with the files that
+    # were actually produced, for both orchestrator and standalone callers.
+    def _model_from(path: pathlib.Path, lens: str) -> str:
+        stem = path.stem  # review-code-<lens>-<model>
+        prefix = f"review-code-{lens}-"
+        return stem[len(prefix):] if stem.startswith(prefix) else stem
+
+    model_a = _model_from(review_files["holistic"], "holistic")
+    model_b = _model_from(review_files["detail"], "detail")
+    model_c = _model_from(review_files["scope"], "scope")
```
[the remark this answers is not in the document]
xbrianh: Fixed in 37792ed: _model_from() now validates the extracted model against MODEL_RE (same pattern used for -a/-b/-c CLI args) and dies with the offending filename if the suffix is empty or invalid.

skills/localgremlin/localreview.py
xbrianh: Fixed in 37792ed: branch on head1_exists so the initial-commit case now emits 'no commit history beyond HEAD and working tree is clean' instead of the misleading HEAD~1..HEAD message.

skills/localgremlin/_core.py
xbrianh: Went with the unlink approach in 37792ed: run_review_code_stage now unlinks all review-code-<lens>-*.md for each lens before spawning reviewers, so --resume-from review-code with different -a/-b/-c can no longer leave two files for the same lens. Documented the behavior in the docstring.

skills/localgremlin/_core.py
xbrianh: Promoted in 37792ed: run_review_code_stage docstring now explicitly states that plan_text="" omits the plan block entirely (the contract /localreview relies on) and that any non-empty plan_text takes the with-plan branch verbatim.

PR #42, opened 2026-04-24
Add --plan flag to skip the planning stage in both gremlins

### LINKED ISSUE #41: Add `--plan <source>` flag to `/localgremlin` and `/ghgremlin`
## Context

Both gremlins always run a `plan` stage as their first step, even when the user already has a plan in hand (from `/design`, an earlier `/ghplan`, a GitHub issue, or just a hand-written markdown doc). That wastes time and tokens, and it forces the user to re-derive what they already wrote down.

Today:
- `/localgremlin` runs the `plan` stage to produce `plan.md` in the session artifacts dir; `implement` reads it from there.
- `/ghgremlin` runs the `ghplan` skill to file a new GitHub issue, captures `ISSUE_NUM` / `ISSUE_URL` from the stream-json, then re-fetches the body via `gh issue view` at the implement stage; `commit-pr` instructs the model to include `Closes #$ISSUE_NUM` in the PR body.

We want to short-circuit all of that when the user already has the plan.

## Approach

Add a `--plan <source>` flag to both gremlins, parsed by each gremlin's own argument loop (not launch.sh). The flag and the positional instructions string are mutually exclusive. The launcher (`skills/_bg/launch.sh`) gains enough awareness of `--plan` to default the gremlin `description` to the first H1 of the plan content when the source is a file; for `/ghgremlin`'s issue-reference form the description is filled in by `ghgremlin.sh` once it has resolved the issue body, overwriting the launcher's fallback.

Why this shape:

- **Same artifact locations as today.** The plan content is staged into exactly the place the downstream stages already read from: `plan.md` in the session dir for `/localgremlin`, and (for `/ghgremlin`) either a resolved `ISSUE_NUM` fetched via `gh` or a local `plan.md` sidecar that `implement` falls back to when `ISSUE_NUM` is empty. Resume keeps working unchanged because the artifacts exist before `implement` runs.
- **No new auth surface for `/ghgremlin`.** Issue resolution uses the `gh` CLI that's already a prerequisite.
- **Flag parsing stays in each gremlin.** The gremlin scripts already own `--resume-from` and model flags; piggy-backing there keeps launch.sh simple. Launch.sh only needs a narrow peek at `--plan <path>` for description extraction, and that reuses its existing file-to-H1 logic (launch.sh:295-308).
- **Mutual exclusion is explicit.** Both gremlins validate early and exit with a clear error before any state directory is created.

### Rejected alternatives

- **Have launch.sh do all `--plan` handling.** Rejected because launch.sh would then need to know per-gremlin plan semantics (file vs. issue ref, where to stage it) and would drift from its current narrow job of setup + state + spawn.
- **Introduce a separate `/localplanfile` and `/ghplanissue` skill.** Rejected: strictly more surface area for the same behavior. The spec explicitly excludes further skill splits.
- **Write the plan to state.json rather than a sidecar file.** Rejected because `implement` already reads `plan.md` / issue bodies; reusing those paths preserves byte-for-byte equivalence with a normal run.

### Flag parsing and validation

- [ ] `skills/localgremlin/localgremlin.py`: extend argparse (or the manual arg loop) to accept `--plan <path>`. Validate path exists and file is non-empty. Error if `--plan` and a positional instructions string are both supplied (or if neither is). Persist `--plan <path>` through `pipeline_args` in state.json so `--resume` replays it.
- [ ] `skills/ghgremlin/ghgremlin.sh`: extend the `getopts`-style loop to accept `--plan <source>`. Classify the source:
  - A string matching a readable file path → file source.
  - `\d+` or `#\d+` → issue number against current repo remote.
  - `owner/repo#\d+` → cross-repo issue.
  - A full `https://github.com/...issues/\d+` URL → absolute issue URL.
  - Anything else → error.
  Mutually exclusive with the positional `INSTRUCTIONS`; error on both.
- [ ] Both: error message matches existing style (stderr + non-zero exit before any state is created).

### Description default (launch.sh)

- [ ] `skills/_bg/launch.sh`: in the description-resolution block (currently launch.sh:286-335), detect `--plan <arg>` inside `pipeline_args`. When `<arg>` is a readable file, reuse the existing H1-extraction path (launch.sh:300-302) against that file instead of the first positional. When `<arg>` is not a file (issue ref), leave the launcher's description as the fallback string — `ghgremlin.sh` will overwrite `.description` in state.json after it resolves the issue.
- [ ] Confirm `--description` explicit flag still wins over `--plan`-derived H1 (same precedence as today).

### `/localgremlin` plan-skip path

- [ ] `skills/localgremlin/localgremlin.py`: when `--plan <path>` is set, copy the file contents to `session_dir / "plan.md"` before the stage loop starts. Skip the `plan` stage entirely (same effect as `--resume-from implement`, but without the resume bookkeeping / rescue counter bump). Implement stage continues to read `plan.md` exactly as today (localgremlin.py:242).
- [ ] Ensure `--plan` + `--resume-from` is handled: if both are given, `--plan` re-stages `plan.md` (idempotent) and `--resume-from` controls the stage gate as usual. Resume from `implement` or later should work transparently even when the original launch was `--plan`-sourced, because `plan.md` already exists.

### `/ghgremlin` plan-skip path

- [ ] `skills/ghgremlin/ghgremlin.sh`: add a plan-source resolution block ahead of the stage loop.
  - **File source**: write the file contents to `$ARTIFACTS_DIR/plan.md`. Leave `ISSUE_NUM` and `ISSUE_URL` empty. Skip the `plan` stage.
  - **Issue source**: resolve `ISSUE_NUM` and `ISSUE_URL` via `gh issue view <ref> --json number,url,body,repository`. Validate the issue's repo matches the current remote unless the user gave `owner/repo#N` or a full URL. Write the body to `$ARTIFACTS_DIR/plan.md` as a cache (so resume doesn't re-fetch unnecessarily; not strictly required but cheap). Skip the `plan` stage. Populate state.json `.description` from first H1 of the issue body if `--description` wasn't supplied.
- [ ] `skills/ghgremlin/ghgremlin.sh` implement stage (currently ghgremlin.sh:243-266): branch on `ISSUE_NUM` set vs unset. When unset (file source), read plan from `$ARTIFACTS_DIR/plan.md` and inject it into the implement prompt where `ISSUE_BODY` used to go.
- [ ] `skills/ghgremlin/ghgremlin.sh` commit-pr stage (currently ghgremlin.sh:284-306): branch on `ISSUE_NUM` set vs unset.
  - Set: unchanged — PR body includes `Closes #${ISSUE_NUM}`, branch name derives from issue num.
  - Unset: PR body omits any `Closes` line; branch name derives from the `--description` or plan H1 slug instead of the issue number.
- [ ] Confirm `ghreview` / `ghaddress` stages don't assume `ISSUE_NUM` is set; patch any that do.

### State persistence and resume

- [ ] Verify `pipeline_args` in state.json captures `--plan <resolved-path>` (after normalizing to absolute path for local file sources) so `--resume` replays correctly.
- [ ] For `/ghgremlin`: persist `ISSUE_NUM` and `ISSUE_URL` (or the explicit absence thereof) in state.json so resume doesn't re-resolve the issue. Keys already exist in the schema (`issue_url`); add `issue_num` if not present.
- [ ] `/gremlins rescue` round-trip: a rescued gremlin reads `pipeline_args` from state.json and passes them back to the gremlin script. Confirm `--plan` + `--resume-from implement` yields the same `plan.md` contents as the original run (since the artifact is on disk).

### Documentation

- [ ] `skills/localgremlin/SKILL.md`: document `--plan <path>`, the mutual exclusion with positional instructions, and the error cases (missing file, empty file, both supplied, neither supplied).
- [ ] `skills/ghgremlin/SKILL.md`: document `--plan <source>` including file and issue-reference forms, the `Closes #N`-only-when-issue-source behavior, and error cases (missing file, empty file, unreachable issue, cross-repo issue via bare `#N`).
- [ ] `CLAUDE.md`: mention the new flag in the two `The ... gremlin` paragraphs if it changes the short description the user sees.

### Tests / manual verification

- [ ] Manual: `/localgremlin --plan ./some-plan.md` produces identical artifacts (module timestamps / session ids) to a normal run whose `plan.md` matches `some-plan.md`.
- [ ] Manual: `/ghgremlin --plan ./some-plan.md` opens a PR implementing it with no new issue filed and no `Closes #N` in the PR body.
- [ ] Manual: `/ghgremlin --plan 42` resolves issue #42 in the current repo, opens a PR with `Closes #42`, and does not file a new issue.
- [ ] Manual: `/ghgremlin --plan owner/other-repo#42` errors cleanly (cross-repo issue — decide at implementation time whether to allow this or reject; spec prefers allow for the explicit `owner/repo#N` form).
- [ ] Manual: passing `--plan foo.md "do X"` to either gremlin fails with a clear error before any state dir is created.
- [ ] Manual: `/gremlins rescue` of a `--plan`-launched gremlin at the `implement` stage works and never re-runs plan.
- [ ] Manual: `/gremlins` listing shows the first H1 of `plan.md` as description when `--plan` is used and `--description` isn't.

## Open questions

- **Cross-repo issue source for `/ghgremlin`.** The spec says `owner/repo#N` and full URLs should work. But the gremlin opens its PR against the *current* repo's remote, so an issue from a different repo can't be `Closes`-linked. Proposed resolution: accept `owner/repo#N` / full URLs as plan *content* only (use the body as the plan), but omit the `Closes` link — same behavior as the file source. Reject only if the `owner/repo` form points at a different repo and the user expected auto-close. Confirm this is acceptable before implementation, or tighten to "same-repo only, error otherwise."
- **Plan re-staging on rescue.** If the user edits the original `--plan` file between the initial launch and a rescue, should the rescue use the edited version or the originally-captured snapshot? Proposed: always use the snapshot in `$ARTIFACTS_DIR/plan.md` (captured at initial launch). This matches the existing invariant that rescue is deterministic with respect to what was recorded in state.

Ref: spec at `/tmp/design-gremlin-plan-flag.md`.

xbrianh:
## Summary

Adds `--plan <source>` to `/localgremlin` and `/ghgremlin` so users can hand
the gremlin a plan they already have (from `/design`, an earlier `/ghplan`,
an existing GitHub issue, or a hand-written markdown file) and skip the
redundant planning stage.

- `/localgremlin --plan <path>` — copy file into `session_dir/plan.md`,
  skip the plan agent, implement reads as-is.
- `/ghgremlin --plan <source>` — file / `42` / `#42` / `owner/repo#42` /
  full GitHub URL. No new issue is filed. `Closes #N` only when the
  resolved issue's repo matches the PR's target repo (so cross-repo
  issue-ref plans don't auto-close unrelated issues).
- Mutually exclusive with the positional `<instructions>`; errors fire
  in `launch.sh` before state dir creation.
- Rescue-safe: `plan.md` is captured at launch and treated as a snapshot;
  later edits or deletions of the original `--plan` source don't leak
  into resumed runs.

## Test plan

- [x] `bash -n` / `py_compile` — all three scripts parse cleanly.
- [x] `parse_args` for `localgremlin.py`: validates missing file, empty
      file, mutex, and neither-supplied cases.
- [x] `launch.sh` mutex: `--plan <path> "extra"` errors without creating
      a state dir.
- [ ] Manual: `/localgremlin --plan ./some-plan.md` produces artifacts
      identical to a normal run whose `plan.md` matches the source.
- [ ] Manual: `/ghgremlin --plan ./some-plan.md` opens a PR with no
      `Closes #N` and no new issue filed.
- [ ] Manual: `/ghgremlin --plan 42` resolves issue #42 in the current
      repo and opens a PR with `Closes #42`.
- [ ] Manual: `/ghgremlin --plan owner/other-repo#42` resolves the
      cross-repo issue body, opens a PR in the current repo, omits the
      `Closes` link.
- [ ] Manual: `/gremlins rescue` of a `--plan`-launched gremlin at the
      `implement` stage re-runs implement without re-running plan.

Closes #41

### DIFF
--- a/skills/_bg/launch.sh
+++ b/skills/_bg/launch.sh
@@ -200,9 +200,16 @@ if [[ -n "$RESUME_GR_ID" ]]; then
     # sidecar INSTRUCTIONS is just flag-echo garbage. Appending it as a
     # positional on resume would make the gremlin re-parse flag text as
     # prose — omit the trailing positional in that case.
+    # Note: this matches both the space-separated form (`--plan <value>`, our
+    # only caller shape today) and the `--plan=<value>` form, so a
+    # hand-crafted resume that packs them together still takes the no-trailing-
+    # positional branch.
     _has_plan=0
     for _pa in "${PIPELINE_ARGS[@]}"; do
-        [[ "$_pa" == "--plan" ]] && { _has_plan=1; break; }
+        if [[ "$_pa" == "--plan" || "$_pa" == --plan=* ]]; then
+            _has_plan=1
+            break
+        fi
     done
 
     (
@@ -317,6 +324,10 @@ done
 # both find the file. A relative --plan path would resolve relative to
 # $WORKDIR post-cd, not the original caller's cwd — breaking on first run.
 # Persisted pipeline_args picks up the normalized value via `set --` below.
+# Note: `cd "$(dirname …)" && pwd` follows symlinks, so the persisted path
+# is the resolved physical path. This is intentional — rescue is more
+# robust if the symlink is later moved or removed — but it means the
+# caller's original symlink path doesn't round-trip into state.json.
 _plan_abs=""
 if [[ -n "$_plan_arg" && -f "$_plan_arg" ]]; then
     _plan_abs=$(cd "$(dirname "$_plan_arg")" 2>/dev/null && printf '%s/%s' "$(pwd)" "$(basename "$_plan_arg")") || _plan_abs=""
@@ -325,6 +336,20 @@ if [[ -n "$_plan_arg" && -f "$_plan_arg" ]]; then
         # Rebuild $@ so the initial dispatch at the bottom of this script
         # uses the same normalized arg list as the persisted pipeline_args.
         set -- "${_args[@]}"
+    fi
+fi
+
+# Early validation for localgremlin --plan <path>: the arg is always a local
+# file path (localgremlin has no issue-ref form), so we can fail fast here
+# (before state dir creation) on missing/empty files instead of deferring to
+# localgremlin.py. ghgremlin's --plan accepts issue refs whose validity
+# depends on the network, so its validation stays inside ghgremlin.sh.
+if [[ "$KIND" == "localgremlin" && -n "$_plan_arg" ]]; then
+    if [[ ! -f "$_plan_arg" ]]; then
+        die "--plan: file not found: $_plan_arg"
+    fi
+    if [[ ! -s "$_plan_arg" ]]; then
+        die "--plan: file is empty: $_plan_arg"
     fi
 fi
 
@@ -403,10 +428,16 @@ SLUG=$(slugify "$SLUG_SOURCE")
 # launch.sh can't see the issue body, DESCRIPTION stays empty so ghgremlin.sh
 # can fill it in after resolving the issue (INSTR_RAW would be just the
 # flag echo "--plan 42" which is not a useful description).
+# The empty-description issue-ref branch is guarded to KIND=ghgremlin: only
+# ghgremlin has an issue-body fill-in step later in the pipeline, and its
+# --plan accepts non-file refs. localgremlin's --plan is always a file (and
+# the earlier early-validation block dies on a non-file value), so a
+# non-file _plan_arg reaching this block under localgremlin is structurally
+# impossible — but the guard makes the assumption explicit and future-proof.
 if [[ -z "$DESCRIPTION" ]]; then
     if [[ -n "$_plan_h1" ]]; then
         DESCRIPTION="${_plan_h1:0:60}"
-    elif [[ -n "$_plan_arg" ]]; then
+    elif [[ -n "$_plan_arg" && "$KIND" == "ghgremlin" ]]; then
         DESCRIPTION=""
     else
         DESCRIPTION="${INSTR_RAW:0:60}"
--- a/skills/_bg/launch.sh
+++ b/skills/_bg/launch.sh
@@ -40,12 +40,14 @@ slugify() {
 }
 
 DESCRIPTION=""
+DESCRIPTION_EXPLICIT=0
 RESUME_GR_ID=""
 while [[ $# -gt 0 ]]; do
     case "$1" in
         --description)
             [[ $# -ge 2 ]] || usage
             DESCRIPTION="$2"
+            DESCRIPTION_EXPLICIT=1
             shift 2
             ;;
         --resume)
@@ -193,10 +195,25 @@ if [[ -n "$RESUME_GR_ID" ]]; then
         done < <(jq -r '.[]' <<<"$PIPELINE_ARGS_JSON")
     fi
 
+    # When --plan is in pipeline_args, the gremlin was launched without a
+    # positional instructions string (the two are mutually exclusive), so the
+    # sidecar INSTRUCTIONS is just flag-echo garbage. Appending it as a
+    # positional on resume would make the gremlin re-parse flag text as
+    # prose — omit the trailing positional in that case.
+    _has_plan=0
+    for _pa in "${PIPELINE_ARGS[@]}"; do
+        [[ "$_pa" == "--plan" ]] && { _has_plan=1; break; }
+    done
+
     (
         cd "$WORKDIR"
-        nohup bash -c '"$PIPELINE" "$@"; EC=$?; "$HOME/.claude/skills/_bg/finish.sh" "$GR_ID" "$EC"' \
-            -- "${PIPELINE_ARGS[@]}" --resume-from "$STAGE" "$INSTRUCTIONS" </dev/null >>"$STATE_DIR/log" 2>&1 &
+        if [[ $_has_plan -eq 1 ]]; then
+            nohup bash -c '"$PIPELINE" "$@"; EC=$?; "$HOME/.claude/skills/_bg/finish.sh" "$GR_ID" "$EC"' \
+                -- "${PIPELINE_ARGS[@]}" --resume-from "$STAGE" </dev/null >>"$STATE_DIR/log" 2>&1 &
+        else
+            nohup bash -c '"$PIPELINE" "$@"; EC=$?; "$HOME/.claude/skills/_bg/finish.sh" "$GR_ID" "$EC"' \
+                -- "${PIPELINE_ARGS[@]}" --resume-from "$STAGE" "$INSTRUCTIONS" </dev/null >>"$STATE_DIR/log" 2>&1 &
+        fi
         echo $! >"$STATE_DIR/pid"
     )
 
@@ -279,41 +296,94 @@ done
 _first_positional=""
 (( _i < ${#_args[@]} )) && _first_positional="${_args[_i]}"
 
+# Find --plan <arg> in the pipeline args so both the slug-source and
+# description fallback paths can peek at the plan file's H1. The arg
+# may be a file path (local plan source) or an issue reference (ghgremlin
+# only) — only file-path values are used here; non-files fall through to
+# the raw-arg fallback, and ghgremlin.sh overwrites .description after
+# resolving the issue body.
+_plan_arg=""
+_plan_arg_idx=-1
+for (( _k=0; _k<${#_args[@]}; _k++ )); do
+    if [[ "${_args[_k]}" == "--plan" && $((_k+1)) -lt ${#_args[@]} ]]; then
+        _plan_arg="${_args[_k+1]}"
+        _plan_arg_idx=$((_k+1))
+        break
+    fi
+done
+
+# Resolve --plan file path to absolute so the initial dispatch (which
+# cd's to $WORKDIR) and any later rescue (which also cd's to $WORKDIR)
+# both find the file. A relative --plan path would resolve relative to
+# $WORKDIR post-cd, not the original caller's cwd — breaking on first run.
+# Persisted pipeline_args picks up the normalized value via `set --` below.
+_plan_abs=""
+if [[ -n "$_plan_arg" && -f "$_plan_arg" ]]; then
+    _plan_abs=$(cd "$(dirname "$_plan_arg")" 2>/dev/null && printf '%s/%s' "$(pwd)" "$(basename "$_plan_arg")") || _plan_abs=""
+    if [[ -n "$_plan_abs" && $_plan_arg_idx -ge 0 ]]; then
+        _args[$_plan_arg_idx]="$_plan_abs"
+        # Rebuild $@ so the initial dispatch at the bottom of this script
+        # uses the same normalized arg list as the persisted pipeline_args.
+        set -- "${_args[@]}"
+    fi
+fi
+
+# Extract first `# heading` from --plan file once for reuse in slug + description.
+_plan_h1=""
+if [[ -n "$_plan_arg" && -f "$_plan_arg" ]]; then
+    _plan_h1=$(head -n 50 "$_plan_arg" 2>/dev/null \
+               | sed -nE '/^#+[[:space:]]+.+/{s/^#+[[:space:]]+//p;q;}' \
+               || true)
+fi
+
+# Mutex: --plan and a positional instructions string cannot both be supplied.
+# Checked here (before any state dir creation) so the error path leaves no
+# litter behind. The gremlin scripts separately validate plan-source content
+# (file existence / issue-ref shape / non-empty body), which happens after
+# state dir creation.
+if [[ -n "$_plan_arg" && -n "$_first_positional" ]]; then
+    die "--plan and positional instructions are mutually exclusive"
+fi
+
 # Slug source resolution, in priority order. This runs *before* the
 # DESCRIPTION-from-INSTR_RAW fallback so the file-path branch isn't
 # shadowed by the fallback (which would feed the slugifier a path like
 # `/tmp/test-spec-slug.md` and produce `tmp-test-spec-slug-md`).
 #   1. Explicit --description (SKILL.md callers compose a clean ≤60-char phrase).
-#   2. If the first positional argument is a readable file, use its first
-#      `# heading` line, or fall back to its basename without extension.
-#   3. The raw instructions with leading flags stripped, first 80 chars.
-#   4. Literal "gremlin" last-resort (applied after slugify if result empty).
+#   2. --plan <file>: use the file's H1, or its basename without extension.
+#   3. First positional argument that is a readable file: same logic.
+#   4. --plan <non-file-arg>: use the raw arg (issue ref like "42" or "owner/repo#42").
+#   5. The raw instructions with leading flags stripped, first 80 chars.
+#   6. Literal "gremlin" last-resort (applied after slugify if result empty).
 SLUG_SOURCE=""
 if [[ -n "$DESCRIPTION" ]]; then
     SLUG_SOURCE="$DESCRIPTION"
-else
-    if [[ -n "$_first_positional" && -f "$_first_positional" ]]; then
-        # Quit on first `# …` line — BSD sed doesn't accept a line-range
-        # with a nested {…} block, so we use a plain pattern with `q` and
-        # cap the scan to 50 lines by piping through head (sed stops reading
-        # once `q` fires, so this is only a safety bound).
-        _title=$(head -n 50 "$_first_positional" 2>/dev/null \
-                 | sed -nE '/^#+[[:space:]]+.+/{s/^#+[[:space:]]+//p;q;}' \
-                 || true)
-        if [[ -n "$_title" ]]; then
-            SLUG_SOURCE="$_title"
-        else
-            _base="${_first_positional##*/}"
-            SLUG_SOURCE="${_base%.*}"
-        fi
+elif [[ -n "$_plan_arg" && -f "$_plan_arg" ]]; then
+    if [[ -n "$_plan_h1" ]]; then
+        SLUG_SOURCE="$_plan_h1"
     else
-        _rem=""
-        for (( _j=_i; _j<${#_args[@]}; _j++ )); do
-            _rem+="${_args[_j]} "
-        done
-        SLUG_SOURCE="${_rem% }"
-        SLUG_SOURCE="${SLUG_SOURCE:0:80}"
+        _base="${_plan_arg##*/}"
+        SLUG_SOURCE="${_base%.*}"
     fi
+elif [[ -n "$_first_positional" && -f "$_first_positional" ]]; then
+    _title=$(head -n 50 "$_first_positional" 2>/dev/null \
+             | sed -nE '/^#+[[:space:]]+.+/{s/^#+[[:space:]]+//p;q;}' \
+             || true)
+    if [[ -n "$_title" ]]; then
+        SLUG_SOURCE="$_title"
+    else
+        _base="${_first_positional##*/}"
+        SLUG_SOURCE="${_base%.*}"
+    fi
+elif [[ -n "$_plan_arg" ]]; then
+    SLUG_SOURCE="$_plan_arg"
+else
+    _rem=""
+    for (( _j=_i; _j<${#_args[@]}; _j++ )); do
+        _rem+="${_args[_j]} "
+    done
+    SLUG_SOURCE="${_rem% }"
+    SLUG_SOURCE="${SLUG_SOURCE:0:80}"
 fi
 
 # If the first positional arg is a readable file (e.g. a spec from /design),
@@ -327,11 +397,20 @@ fi
 SLUG=$(slugify "$SLUG_SOURCE")
 [[ -z "$SLUG" ]] && SLUG="gremlin"
 
-# Description fallback: explicit --description wins; otherwise fall back to
-# a truncated slice of the raw instructions so the status views always have
-# something to print.
+# Description fallback: explicit --description wins; otherwise prefer the
+# --plan file's H1 (when the plan arg is a local file), and finally fall back
+# to a truncated slice of the raw instructions. For a --plan issue-ref where
+# launch.sh can't see the issue body, DESCRIPTION stays empty so ghgremlin.sh
+# can fill it in after resolving the issue (INSTR_RAW would be just the
+# flag echo "--plan 42" which is not a useful description).
 if [[ -z "$DESCRIPTION" ]]; then
-    DESCRIPTION="${INSTR_RAW:0:60}"
+    if [[ -n "$_plan_h1" ]]; then
+        DESCRIPTION="${_plan_h1:0:60}"
+    elif [[ -n "$_plan_arg" ]]; then
+        DESCRIPTION=""
+    else
+        DESCRIPTION="${INSTR_RAW:0:60}"
+    fi
 fi
 
 RANDHEX=$(LC_ALL=C tr -dc 'a-f0-9' </dev/urandom 2>/dev/null | head -c 6 || true)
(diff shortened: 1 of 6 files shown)


xbrianh (commented):

skills/_bg/launch.sh
```diff
@@ -193,10 +195,25 @@ if [[ -n "$RESUME_GR_ID" ]]; then
...
     fi
 
+    # When --plan is in pipeline_args, the gremlin was launched without a
+    # positional instructions string (the two are mutually exclusive), so the
+    # sidecar INSTRUCTIONS is just flag-echo garbage. Appending it as a
+    # positional on resume would make the gremlin re-parse flag text as
+    # prose — omit the trailing positional in that case.
+    _has_plan=0
+    for _pa in "${PIPELINE_ARGS[@]}"; do
+        [[ "$_pa" == "--plan" ]] && { _has_plan=1; break; }
```
xbrianh: Matches on exact `--plan` only. That's consistent with how callers currently pass the flag (two separate argv elements), but if anyone ever forwards `--plan=<value>` form, this would silently miss it and the `else` branch below appends `$INSTRUCTIONS` as a positional to the resumed pipeline — which is the exact bug this detection is trying to prevent. Either add a comment pinning the "space-separated only" assumption, or broaden to `[[ "$_pa" == "--plan" || "$_pa" == --plan=* ]]`.

skills/_bg/launch.sh
```diff
@@ -279,41 +296,94 @@ done
...
+done
+
+# Resolve --plan file path to absolute so the initial dispatch (which
+# cd's to $WORKDIR) and any later rescue (which also cd's to $WORKDIR)
+# both find the file. A relative --plan path would resolve relative to
+# $WORKDIR post-cd, not the original caller's cwd — breaking on first run.
+# Persisted pipeline_args picks up the normalized value via `set --` below.
+_plan_abs=""
+if [[ -n "$_plan_arg" && -f "$_plan_arg" ]]; then
+    _plan_abs=$(cd "$(dirname "$_plan_arg")" 2>/dev/null && printf '%s/%s' "$(pwd)" "$(basename "$_plan_arg")") || _plan_abs=""
```
xbrianh, line 333: `cd "$(dirname …)" && pwd` canonicalizes through symlinks, so `~/symlink-to-plan.md` gets persisted to `pipeline_args` and `state.json` as the resolved physical path. That's probably desired (more rescue-safe if the symlink moves later), but it's not obvious from the surrounding code. If the intent is to preserve the logical path, use `pwd -L`; otherwise a one-line comment noting the resolve-symlinks behavior would help.

skills/_bg/launch.sh
```diff
@@ -327,11 +397,20 @@ fi
...
+# --plan file's H1 (when the plan arg is a local file), and finally fall back
+# to a truncated slice of the raw instructions. For a --plan issue-ref where
+# launch.sh can't see the issue body, DESCRIPTION stays empty so ghgremlin.sh
+# can fill it in after resolving the issue (INSTR_RAW would be just the
+# flag echo "--plan 42" which is not a useful description).
 if [[ -z "$DESCRIPTION" ]]; then
-    DESCRIPTION="${INSTR_RAW:0:60}"
+    if [[ -n "$_plan_h1" ]]; then
+        DESCRIPTION="${_plan_h1:0:60}"
+    elif [[ -n "$_plan_arg" ]]; then
```
xbrianh: This branch leaves `DESCRIPTION` empty for non-file `_plan_arg` so ghgremlin.sh can fill it in after resolving the issue body. That works for ghgremlin but runs for *any* `KIND` — if localgremlin ever accepts a non-file `--plan` form, this would leave the description permanently empty (localgremlin has no later fill-in step). Not broken today (localgremlin.py rejects non-file paths at `parse_args`), but a `[[ "$KIND" == ghgremlin ]]` guard or a comment pinning the assumption would harden it.

skills/ghgremlin/ghgremlin.sh
```diff
@@ -127,6 +140,94 @@ command -v jq >/dev/null     || die "jq not found"
...
+      #   https://github.com/owner/repo/issues/42[#fragment] → absolute URL
+      _issue_ref=""
+      _target_repo=""
+      if [[ "$PLAN_SOURCE" =~ ^#?([0-9]+)$ ]]; then
+        _issue_ref="${BASH_REMATCH[1]}"
+        _target_repo="$REPO"
+      elif [[ "$PLAN_SOURCE" =~ ^([A-Za-z0-9_.-]+/[A-Za-z0-9_.-]+)#([0-9]+)$ ]]; then
+        _target_repo="${BASH_REMATCH[1]}"
+        _issue_ref="${BASH_REMATCH[2]}"
+      elif [[ "$PLAN_SOURCE" =~ ^https://github\.com/([A-Za-z0-9_.-]+/[A-Za-z0-9_.-]+)/issues/([0-9]+)(#.*)?$ ]]; then
```
xbrianh, line 192: URL regex only matches `https://github.com/...`. GitHub Enterprise users (custom host) fall through to the "unrecognized shape" die. The `owner/repo#N` form still works for them, so not a blocker — but either mention the github.com-only constraint in SKILL.md or broaden to `^https?://[^/]+/([A-Za-z0-9_.-]+/[A-Za-z0-9_.-]+)/issues/([0-9]+)(#.*)?$` and capture the host.

skills/ghgremlin/ghgremlin.sh
```diff
@@ -127,6 +140,94 @@ command -v jq >/dev/null     || die "jq not found"
...
+      _resolved_num=$(jq -r '.number // ""' <<<"$_issue_json")
+      _resolved_nwo=$(jq -r '.repository.nameWithOwner // empty' <<<"$_issue_json")
+      # Only set ISSUE_URL/ISSUE_NUM (which drive the `Closes #N` link) when
+      # the resolved issue's repo matches the PR's target repo. Cross-repo
+      # issue sources get the body as plan content but no auto-close link.
+      if [[ "$_resolved_nwo" == "$REPO" ]]; then
+        ISSUE_URL="$_resolved_url"
+        ISSUE_NUM="$_resolved_num"
+      fi
+      printf '%s' "$ISSUE_BODY" > "$PLAN_MD"
```
xbrianh: `printf '%s'` strips the trailing newline, while the file-source branch at line 176 uses `cp` which preserves the source verbatim. The two code paths produce subtly different `plan.md` contents on disk for the same logical "plan body". Harmless for the LLM prompt, but inconsistent — `printf '%s\n'` (or `cat <<<"$ISSUE_BODY"`) would match `cp` semantics.

skills/ghgremlin/SKILL.md
```diff
@@ -59,6 +59,31 @@ Report the gremlin id, workdir, and log path that it prints. Make clear to the u
...
+- **Full URL `https://github.com/owner/repo/issues/42`** — same cross-repo
+  semantics.
+
+Rules:
+
+- Mutually exclusive with the positional `<instructions>`. Pass exactly one.
+- Files must be non-empty; issues must exist and have a non-empty body.
+- Errors (missing file, empty file, unreachable issue, unrecognized shape,
+  both forms supplied, neither supplied) are surfaced before the state
+  directory is created or cleaned up if the plan source turns out bad.
```
xbrianh: "surfaced before the state directory is created or cleaned up if the plan source turns out bad" reads ambiguously. Actual behavior: file-missing/empty and mutex errors fire **before** state dir creation (launch.sh); issue-ref errors (unreachable, unrecognized shape, empty body) fire **after** state dir creation in ghgremlin.sh and leave a failed state dir behind for `/gremlins rm`. Splitting into two bullets would make the clean-vs-dirty failure modes explicit.

skills/localgremlin/localgremlin.py
```diff
@@ -83,34 +83,56 @@ def changes_outside_git(sentinel: pathlib.Path, session_dir: pathlib.Path) -> bo
...
     # raising SystemExit. Wrapping it would bury the specific error behind
     # a second copy of the usage line.
     args = parser.parse_args(argv)
-    if not args.instructions:
-        die(usage)
-    for m in (args.plan, args.impl, args.address, args.holistic, args.detail, args.scope):
+    # launch.sh resume may pass an empty-string positional when a --plan
+    # gremlin is resumed; treat that as "no positional supplied" rather than
+    # a literal empty-string instruction.
+    args.instructions = [s for s in args.instructions if s]
```
xbrianh: Silently filtering empty-string positionals changes fresh-launch behavior too: `localgremlin.py --plan foo.md ""` used to trip the mutex (non-empty `instructions` list containing `""`), now succeeds because the filter strips the `""`. The comment ties the filter specifically to the resume path — either generalize the comment or narrow the filter with `if args.resume_from:` so the non-resume path's mutex stays strict.

skills/localgremlin/localgremlin.py
```diff
@@ -215,9 +237,21 @@ def main(argv: List[str]) -> int:
...
+    # agent entirely. An existing plan.md (resume path) is left alone — the
+    # snapshot captured at initial launch is authoritative, per the spec's
+    # rescue-determinism rule.
+    if args.plan_path:
+        if not plan_file.exists():
+            src = pathlib.Path(args.plan_path)
+            if not src.is_file() or src.stat().st_size == 0:
+                die(f"--plan: source file missing or empty at resume: {src}")
+            shutil.copyfile(src, plan_file)
+        print(f"==> [1/4] plan supplied via --plan -> {plan_file}", flush=True)
```
xbrianh: Minor log-wording nit: this fires on both fresh and resumed `--plan` runs, but on resume the copy at line 250 is skipped (plan.md already exists). Reads as if we just copied something we didn't. Consider distinguishing `"plan supplied via --plan (copied)"` vs `"plan reused from snapshot"` to match ghgremlin.sh's `"plan resumed from snapshot"` at line 169.

xbrianh (commented):
Overall a well-scoped and thoughtfully implemented feature. The mutex, resume handling, cross-repo semantics, and state.json plumbing are all considered carefully, and the comments explain the non-obvious branches well. A few notes below on behavior edges and documentation drift worth a second look; nothing blocking. The PR description still has several unchecked manual test items (`/localgremlin --plan`, `/ghgremlin --plan <file>`, `/ghgremlin --plan 42`, `/ghgremlin --plan owner/other-repo#42`, rescue of a `--plan` gremlin). Worth running at least one end-to-end of each form before landing — the bash-syntax checks and `parse_args` unit tests don't exercise the actual launch → implement → resume path.

skills/_bg/launch.sh
xbrianh: Matches on exact `--plan` only. That's consistent with how callers currently pass the flag (two separate argv elements), but if anyone ever forwards `--plan=<value>` form, this would silently miss it and the `else` branch below appends `$INSTRUCTIONS` as a positional to the resumed pipeline — which is the exact bug this detection is trying to prevent. Either add a comment pinning the "space-separated only" assumption, or broaden to `[[ "$_pa" == "--plan" || "$_pa" == --plan=* ]]`.

skills/_bg/launch.sh
xbrianh, line 333: `cd "$(dirname …)" && pwd` canonicalizes through symlinks, so `~/symlink-to-plan.md` gets persisted to `pipeline_args` and `state.json` as the resolved physical path. That's probably desired (more rescue-safe if the symlink moves later), but it's not obvious from the surrounding code. If the intent is to preserve the logical path, use `pwd -L`; otherwise a one-line comment noting the resolve-symlinks behavior would help.

skills/_bg/launch.sh
xbrianh: This branch leaves `DESCRIPTION` empty for non-file `_plan_arg` so ghgremlin.sh can fill it in after resolving the issue body. That works for ghgremlin but runs for *any* `KIND` — if localgremlin ever accepts a non-file `--plan` form, this would leave the description permanently empty (localgremlin has no later fill-in step). Not broken today (localgremlin.py rejects non-file paths at `parse_args`), but a `[[ "$KIND" == ghgremlin ]]` guard or a comment pinning the assumption would harden it.

skills/ghgremlin/ghgremlin.sh
xbrianh, line 192: URL regex only matches `https://github.com/...`. GitHub Enterprise users (custom host) fall through to the "unrecognized shape" die. The `owner/repo#N` form still works for them, so not a blocker — but either mention the github.com-only constraint in SKILL.md or broaden to `^https?://[^/]+/([A-Za-z0-9_.-]+/[A-Za-z0-9_.-]+)/issues/([0-9]+)(#.*)?$` and capture the host.

skills/ghgremlin/ghgremlin.sh
xbrianh: `printf '%s'` strips the trailing newline, while the file-source branch at line 176 uses `cp` which preserves the source verbatim. The two code paths produce subtly different `plan.md` contents on disk for the same logical "plan body". Harmless for the LLM prompt, but inconsistent — `printf '%s\n'` (or `cat <<<"$ISSUE_BODY"`) would match `cp` semantics.

skills/ghgremlin/SKILL.md
xbrianh: "surfaced before the state directory is created or cleaned up if the plan source turns out bad" reads ambiguously. Actual behavior: file-missing/empty and mutex errors fire **before** state dir creation (launch.sh); issue-ref errors (unreachable, unrecognized shape, empty body) fire **after** state dir creation in ghgremlin.sh and leave a failed state dir behind for `/gremlins rm`. Splitting into two bullets would make the clean-vs-dirty failure modes explicit.

skills/localgremlin/localgremlin.py
xbrianh: Silently filtering empty-string positionals changes fresh-launch behavior too: `localgremlin.py --plan foo.md ""` used to trip the mutex (non-empty `instructions` list containing `""`), now succeeds because the filter strips the `""`. The comment ties the filter specifically to the resume path — either generalize the comment or narrow the filter with `if args.resume_from:` so the non-resume path's mutex stays strict.

skills/localgremlin/localgremlin.py
xbrianh: Minor log-wording nit: this fires on both fresh and resumed `--plan` runs, but on resume the copy at line 250 is skipped (plan.md already exists). Reads as if we just copied something we didn't. Consider distinguishing `"plan supplied via --plan (copied)"` vs `"plan reused from snapshot"` to match ghgremlin.sh's `"plan resumed from snapshot"` at line 169.

xbrianh (commented):
Overall a well-scoped and thoughtfully implemented feature. The mutex, resume handling, cross-repo semantics, and state.json plumbing are all considered carefully, and the comments explain the non-obvious branches well. A few notes below on behavior edges and documentation drift worth a second look; nothing blocking. The PR description still has several unchecked manual test items (`/localgremlin --plan`, `/ghgremlin --plan <file>`, `/ghgremlin --plan 42`, `/ghgremlin --plan owner/other-repo#42`, rescue of a `--plan` gremlin). Worth running at least one end-to-end of each form before landing — the bash-syntax checks and `parse_args` unit tests don't exercise the actual launch → implement → resume path.

skills/localgremlin/localgremlin.py
```diff
@@ -215,9 +237,21 @@ def main(argv: List[str]) -> int:
...
+    elif start_idx <= VALID_RESUME_STAGES.index("plan"):
```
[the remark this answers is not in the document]
xbrianh: Good catch — fixed. The source file copy into `session_dir/plan.md` now happens in `main()` before the `--resume-from` precondition check, so `--plan <path> --resume-from implement` finds plan.md already in place. Commit 0fe2fb4.

skills/localgremlin/localgremlin.py
```diff
@@ -83,34 +83,56 @@ def changes_outside_git(sentinel: pathlib.Path, session_dir: pathlib.Path) -> bo
...
+        # Fresh launch: validate that the path is a readable, non-empty file.
+        # The snapshot check happens in main() against session_dir, so here
+        # we only do the fresh-launch validation. When resumed, main() skips
+        # re-copying, so a missing source file is tolerated there.
+        if not args.resume_from:
+            p = pathlib.Path(args.plan_path)
+            if not p.is_file():
+                die(f"--plan: file not found: {args.plan_path}")
+            if p.stat().st_size == 0:
+                die(f"--plan: file is empty: {args.plan_path}")
```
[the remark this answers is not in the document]
xbrianh: Fixed. Source-file validation moved from `parse_args()` into `main()` where it can check the `session_dir/plan.md` snapshot first: on resume the snapshot wins and the source file may legitimately be missing; only a fresh launch (no snapshot yet) validates the source path. Commit 0fe2fb4.

skills/localgremlin/localgremlin.py
```diff
@@ -215,9 +237,21 @@ def main(argv: List[str]) -> int:
...
+    # When --plan <path> is set, we short-circuit the plan stage: copy the
+    # user-supplied plan file into session_dir/plan.md and skip the planning
+    # agent entirely. An existing plan.md (resume path) is left alone — the
+    # snapshot captured at initial launch is authoritative, per the spec's
+    # rescue-determinism rule.
+    if args.plan_path:
+        if not plan_file.exists():
+            src = pathlib.Path(args.plan_path)
+            if not src.is_file() or src.stat().st_size == 0:
+                die(f"--plan: source file missing or empty at resume: {src}")
```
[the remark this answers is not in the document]
xbrianh: Reworded — the error now says `--plan: file not found` / `--plan: file is empty`, and since the copy moved into `main()` before the resume preconditions, the fresh-launch and resume paths share the same validation with the same clean message. Commit 0fe2fb4.

skills/localgremlin/SKILL.md
```diff
@@ -66,6 +66,19 @@ Report the gremlin id, workdir, and log path that it prints. Make clear to the u
...
+If the user already has an implementation plan, pass `--plan <path>` to skip
+the plan stage. The file's contents are copied into the gremlin's session as
+`plan.md` and the implement stage reads them as-is.
+
+- Mutually exclusive with the positional `<instructions>`. Pass exactly one.
+- The path must point to a readable, non-empty file.
+- The gremlin's description defaults to the first `# heading` in the plan
+  file unless `--description` is supplied explicitly.
+- Errors (file missing, file empty, both `--plan` and positional supplied,
+  neither supplied) are surfaced before the state directory is created.
```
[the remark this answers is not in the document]
xbrianh: Added early validation in `launch.sh` for `KIND=localgremlin` when `--plan <path>` is present: missing or empty file dies before state-dir creation, matching the doc's claim. (ghgremlin's `--plan` still accepts issue refs whose validity depends on the network, so those checks stay inside `ghgremlin.sh` — the ghgremlin SKILL.md now calls that split out explicitly.) Commit 0fe2fb4.

skills/ghgremlin/SKILL.md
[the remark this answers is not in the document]
xbrianh: Reworded to split the failure modes into two bullets: clean (mutex + missing/empty local file, caught in `launch.sh` before state-dir creation) vs dirty (issue-ref errors, caught in `ghgremlin.sh` after state-dir creation — use `/gremlins rm` to clean up). Commit 0fe2fb4.

skills/_bg/launch.sh
xbrianh: Broadened to match both `--plan <val>` and `--plan=<val>` (and added a comment pinning the two-form assumption). Commit 0fe2fb4.

skills/_bg/launch.sh
xbrianh: Added a comment noting the `cd $(dirname …) && pwd` symlink-canonicalization is intentional (rescue is more robust if the caller's symlink moves or gets removed) but not obvious. Left the behavior as-is. Commit 0fe2fb4.

skills/_bg/launch.sh
xbrianh: Added a `KIND == ghgremlin` guard on the empty-description issue-ref branch plus a comment explaining why the localgremlin path is structurally impossible today (early-validation block dies on non-file values). Future-proof. Commit 0fe2fb4.

skills/ghgremlin/ghgremlin.sh
xbrianh: Called out the github.com-only constraint in ghgremlin SKILL.md and steered GHE users to the `owner/repo#42` form (which works for any host `gh` is configured to reach). Left the regex as-is since broadening it would invite spurious URL matches. Commit 0fe2fb4.

skills/ghgremlin/ghgremlin.sh
xbrianh: Switched to `printf '%s\n'` to match the `cp` semantics of the file-source branch. Commit 0fe2fb4.

skills/ghgremlin/SKILL.md
xbrianh: Split into two bullets — clean (mutex + missing/empty local file, caught in launch.sh before state-dir creation) vs dirty (issue-ref errors, leave a failed state dir for `/gremlins rm`). Commit 0fe2fb4.

skills/localgremlin/localgremlin.py
xbrianh: Narrowed the empty-string filter with `if args.resume_from:` so the fresh-launch mutex (`--plan foo.md ""`) still fires on a literal empty-string positional. Commit 0fe2fb4.

skills/localgremlin/localgremlin.py
xbrianh: Split the log line: fresh-launch prints `plan supplied via --plan (copied) -> …`, resume prints `plan reused from snapshot -> …` (mirrors ghgremlin.sh's `plan resumed from snapshot` wording). Commit 0fe2fb4.

skills/_bg/launch.sh
xbrianh: Addressed in commit 0fe2fb4 (see duplicate thread for details).

skills/_bg/launch.sh
xbrianh: Addressed in commit 0fe2fb4 (see duplicate thread for details).

skills/_bg/launch.sh
xbrianh: Addressed in commit 0fe2fb4 (see duplicate thread for details).

skills/ghgremlin/ghgremlin.sh
xbrianh: Addressed in commit 0fe2fb4 (see duplicate thread for details).

skills/ghgremlin/ghgremlin.sh
xbrianh: Addressed in commit 0fe2fb4 (see duplicate thread for details).

skills/ghgremlin/SKILL.md
xbrianh: Addressed in commit 0fe2fb4 (see duplicate thread for details).

skills/localgremlin/localgremlin.py
xbrianh: Addressed in commit 0fe2fb4 (see duplicate thread for details).

skills/localgremlin/localgremlin.py
xbrianh: Addressed in commit 0fe2fb4 (see duplicate thread for details).

PR #44, opened 2026-04-24
Add /bossgremlin for chained serial gremlin workflows

xbrianh:
## Summary

- Adds `/bossgremlin` skill: a manager gremlin that chains multiple child gremlins (all local or all gh) serially without babysitting — give it a top-level spec and it runs handoff → launch → wait → land → repeat until the chain is done or bails
- The manager (`bossgremlin.py`) never reads plan/diff/review content; its entire decision surface is the handoff agent's `.state.json`, the `finished` marker, and exit codes from `land` and `rescue --headless`
- Per-handoff plan files (`handoff-001.md`, `handoff-001-child.md`, `handoff-001.state.json`, …) accumulate in the boss's state dir alongside a `boss_state.json` that records child outcomes and chain base ref — fully inspectable and resumable
- Boss shows in `/gremlins` as kind `boss`; children surface their owning boss via the existing `[boss:<id>]` column marker
- SIGTERM forwarded to current subprocess; during child-wait polling, explicitly calls `gremlins stop <child-id>` before exiting
- `launch.sh` and `gremlins.py` updated minimally: bossgremlin added to kind validation, `--plan` early-validation, `kind_short`, `GREMLIN_STAGES`, `GREMLIN_SCRIPTS`, and `do_land`

## Test plan

- [ ] Launch with a two-step local spec; confirm both children land and chain reports done
- [ ] Launch with a gh chain spec; confirm PRs created and merged serially
- [ ] Verify `/gremlins` shows boss as `boss` kind and children show `[boss:<id>]`
- [ ] Kill the boss mid-wait; confirm child is stopped; rescue boss and confirm chain resumes
- [ ] Force a child failure; confirm `rescue --headless` is called; exhaust rescue cap and confirm chain halts with recorded bail reason
- [ ] Confirm `launch.sh` rejects unknown kind (pre-existing behavior preserved)
- [ ] Confirm `gremlins land <boss-id>` prints the informative refusal message

### DIFF
--- a/skills/bossgremlin/bossgremlin.py
+++ b/skills/bossgremlin/bossgremlin.py
@@ -147,7 +147,7 @@ def save_boss_state(state_dir: str, boss_state: dict) -> None:
 
 
 def run_handoff(gr_id: str, state_dir: str, boss_state: dict,
-                project_root: str, model: str) -> tuple:
+                project_root: str, model: str, target_branch: str) -> tuple:
     """Run handoff agent. Returns (exit_state, signal dict).
 
     Updates boss_state in place (handoff_count, current_plan, handoff_records).
@@ -164,7 +164,7 @@ def run_handoff(gr_id: str, state_dir: str, boss_state: dict,
     current_plan = boss_state["current_plan"]
     base_ref = boss_state["chain_base_ref"]
 
-    log(f"handoff {n}: plan={current_plan}, base={base_ref[:12]}")
+    log(f"handoff {n}: plan={current_plan}, base={base_ref[:12]}, rev={target_branch or 'HEAD'}")
     cmd = [
         handoff_script,
         "--plan", current_plan,
@@ -173,6 +173,11 @@ def run_handoff(gr_id: str, state_dir: str, boss_state: dict,
         "--model", model,
         "--timeout", str(HANDOFF_TIMEOUT),
     ]
+    if target_branch:
+        # Run in the real repo root so git can resolve the target branch, and
+        # pass --rev so handoff inspects landed work on that branch rather than
+        # whatever HEAD the caller happens to have checked out.
+        cmd += ["--rev", target_branch]
     rc = run_proc(cmd, cwd=project_root)
     check_stop()
 
@@ -379,6 +384,7 @@ def main(argv):
                 boss_state=boss_state,
                 project_root=project_root,
                 model=args.model,
+                target_branch=boss_state.get("target_branch", ""),
             )
             save_boss_state(state_dir, boss_state)
             check_stop()
@@ -404,16 +410,57 @@ def main(argv):
             current_child_id = launch_child(gr_id, launch_kind, child_plan)
             boss_state["current_child_id"] = current_child_id
             save_boss_state(state_dir, boss_state)
+            # Stop the freshly launched child if a stop was requested during
+            # or just after launch (pre-wait window).
+            if _stop_requested:
+                log(f"stop requested — stopping newly launched child {current_child_id}")
+                gremlins = os.path.expanduser("~/.claude/skills/gremlins/gremlins.py")
+                if os.access(gremlins, os.X_OK):
+                    subprocess.run([gremlins, "stop", current_child_id], capture_output=True)
+                sys.exit(130)
             check_stop()
 
         else:
             # Resume path: already have a child in flight
             log(f"resuming with in-flight child: {current_child_id}")
             if child_is_closed(current_child_id):
-                log(f"child {current_child_id} already closed — treating as landed")
-                boss_state["current_child_id"] = None
-                save_boss_state(state_dir, boss_state)
-                continue
+                # `closed` is a UI hide flag, not a success/failure signal.
+                # Inspect the finished marker and exit_code to determine outcome.
+                child_wdir = os.path.join(STATE_ROOT, current_child_id)
+                finished_path = os.path.join(child_wdir, "finished")
+                state_path = os.path.join(child_wdir, "state.json")
+                if os.path.isfile(finished_path) and os.path.isfile(state_path):
+                    try:
+                        child_state = load_json(state_path)
+                        child_succeeded = child_state.get("exit_code") == 0
+                    except Exception:
+                        child_succeeded = False
+                else:
+                    child_succeeded = False
+
+                if child_succeeded:
+                    # Already finished successfully and closed — treat as landed
+                    log(f"child {current_child_id} already finished and closed — treating as landed")
+                    boss_state["children"].append({
+                        "id": current_child_id,
+                        "outcome": "landed",
+                    })
+                    boss_state["current_child_id"] = None
+                    save_boss_state(state_dir, boss_state)
+                    continue
+                else:
+                    # Closed but not successfully finished — operator may have
+                    # manually hidden a failed child.  Halt for operator action.
+                    log(
+                        f"child {current_child_id} is closed but did not finish successfully"
+                        f" — operator action required"
+                    )
+                    boss_state["current_child_id"] = None
+                    save_boss_state(state_dir, boss_state)
+                    die(
+                        f"chain halted: child {current_child_id} was manually closed without"
+                        f" successfully finishing — inspect and resume or reassign"
+                    )
 
         # Step 2: inner loop — wait → land → (rescue → wait → land)* → bail
         was_rescued = False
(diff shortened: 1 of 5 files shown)


xbrianh (commented):

skills/bossgremlin/bossgremlin.py
```diff
@@ -0,0 +1,466 @@
...
+    log(f"handoff {n}: plan={current_plan}, base={base_ref[:12]}")
+    cmd = [
+        handoff_script,
+        "--plan", current_plan,
+        "--out", out_path,
+        "--base", base_ref,
+        "--model", model,
+        "--timeout", str(HANDOFF_TIMEOUT),
+    ]
+    rc = run_proc(cmd, cwd=project_root)
```
xbrianh, line 181: **Handoff runs in the boss's frozen worktree — `git log base..HEAD` will always be empty after the first child lands.**

`project_root` is the boss's own git worktree, created by `launch.sh` at launch time and pinned to the launch-time commit. When `run_handoff` calls `run_proc(cmd, cwd=project_root)`, handoff.py runs `git log chain_base_ref..HEAD` in that worktree. Because `HEAD` in the boss's worktree never advances (the boss itself makes no commits), the diff is always empty — the handoff agent sees no evidence that any children have landed work. On the second and later handoffs it will make its decision blind to all prior changes.

For a `local` chain the fix is to `git fetch` + `git merge --ff-only origin/<target_branch>` (or the equivalent fast-forward) in the boss's worktree after each successful land, so HEAD tracks the landed work. Alternatively, pass a `--work-dir <real-target-branch-checkout>` flag to handoff.py so it runs its git commands against the right tree.

skills/bossgremlin/bossgremlin.py
```diff
@@ -0,0 +1,466 @@
...
+                set_stage(gr_id, "landing")
+                if land_child(current_child_id):
+                    outcome = "rescued-then-landed" if was_rescued else "landed"
+                    log(f"child {current_child_id} {outcome}")
+                    boss_state["children"].append({"id": current_child_id, "outcome": outcome})
+                    boss_state["current_child_id"] = None
+                    save_boss_state(state_dir, boss_state)
+                    break  # inner loop done; outer loop continues to next handoff
+                else:
+                    log(f"landing failed for {current_child_id}")
```
xbrianh: **Land failure falls through to rescue, but rescue can't fix a merge conflict or push rejection.**

When `success` is `True` (the pipeline finished with exit code 0) but `land_child` returns `False`, execution falls through to `set_stage(gr_id, "rescuing")` and `rescue_child`. `rescue --headless` relaunches the pipeline from its last failed stage — but if the pipeline itself succeeded and only the `land` step failed (e.g. a merge conflict, a branch protection rule rejection, or a squash conflict with a concurrent landed change), re-running the pipeline won't clear the blocker. The boss will rescue in a loop until gremlins.py exhausts its rescue cap and refuses, at which point the chain halts with a confusing bail message.

Suggested fix: treat a land failure after a successful pipeline as a hard bail:
```python
else:
    log(f"landing failed for {current_child_id} — cannot auto-resolve, halting chain")
    boss_state["children"].append({"id": current_child_id, "outcome": "land-failed"})
    boss_state["current_child_id"] = current_child_id
    save_boss_state(state_dir, boss_state)
    die(f"chain halted: land failed for child {current_child_id} (pipeline was clean)")
```

skills/bossgremlin/bossgremlin.py
```diff
@@ -0,0 +1,466 @@
...
+            check_stop()
+            current_child_id = launch_child(gr_id, launch_kind, child_plan)
+            boss_state["current_child_id"] = current_child_id
+            save_boss_state(state_dir, boss_state)
+            check_stop()
+
+        else:
+            # Resume path: already have a child in flight
+            log(f"resuming with in-flight child: {current_child_id}")
+            if child_is_closed(current_child_id):
```
xbrianh, line 426: **Closed-child resume path doesn't append the child to `boss_state["children"]`.**

When the boss resumes and finds `current_child_id` is already closed, it resets `current_child_id = None` and `continue`s — but never calls `boss_state["children"].append(...)`. The `children` list in `boss_state.json` will be missing this entry, so the count reported in the resume log line (`completed children: N`) will be off, and any tooling that inspects the children list won't see it.

```python
if child_is_closed(current_child_id):
    log(f"child {current_child_id} already closed — treating as landed")
    boss_state["children"].append({"id": current_child_id, "outcome": "pre-closed"})
    boss_state["current_child_id"] = None
    save_boss_state(state_dir, boss_state)
    continue
```

xbrianh (commented):
Overall the state machine is clean and the boss/child separation is well-reasoned. Three issues worth addressing before this ships: the handoff agent runs in the boss's frozen worktree and will always see an empty `git log base..HEAD` diff after the first child lands (the most consequential); a land failure silently falls through to rescue even though re-running the pipeline can't fix a merge conflict; and the resume path for an already-closed child omits the child from the `children` tracking list.

skills/bossgremlin/bossgremlin.py
xbrianh: Fixed. Added a `--rev` flag to `handoff.py` so callers can pass an explicit ref instead of relying on HEAD. `run_handoff` now passes `--rev <target_branch>` (from `boss_state["target_branch"]`) so the agent inspects the tip of the branch where children are landing rather than whatever the real repo's HEAD happens to point at.

skills/bossgremlin/bossgremlin.py
xbrianh: Fixed. The land failure path now halts the chain immediately with an operator-action message rather than falling through to rescue. A pipeline success + land failure means the blocker (merge conflict, branch protection, etc.) can't be cleared by re-running the pipeline, so rescue would just loop. The boss dies with the child's outcome recorded as `land-failed` and `current_child_id` cleared, so a resume after the operator fixes the blocker starts cleanly from handoff.

skills/bossgremlin/bossgremlin.py
xbrianh: Fixed. The closed-child resume path now checks `finished` + `exit_code` instead of treating `closed` as a success signal. If finished+exit_code==0, it appends to `children` (with outcome `landed`) and continues. If closed without a successful finish, it clears `current_child_id` and halts with an operator-action message.

skills/gremlins/gremlins.py
```diff
@@ -194,6 +194,8 @@ def kind_short(kind: str) -> str:
         return "local"
     if kind == "ghgremlin":
         return "gh"
+    if kind == "bossgremlin":
+        return "boss"
     return kind or ""
```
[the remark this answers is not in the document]
xbrianh: Fixed. `--kind` now accepts `boss` as a valid choice (`choices=["local", "gh", "boss"]`).

skills/bossgremlin/bossgremlin.py
```diff
@@ -0,0 +1,466 @@
...
+                log(f"child {current_child_id} already closed — treating as landed")
+                boss_state["current_child_id"] = None
+                save_boss_state(state_dir, boss_state)
+                continue
```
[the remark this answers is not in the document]
xbrianh: Addressed together with the children-list fix: closed is no longer treated as success. The path now reads `finished` + `exit_code` and either records a landing or halts for operator action.

skills/bossgremlin/bossgremlin.py
```diff
@@ -0,0 +1,466 @@
...
+            # next-plan: launch the next child
+            child_plan = sig.get("child_plan")
+            if not child_plan or not os.path.isfile(child_plan):
+                die(f"handoff returned next-plan but child_plan not found: {child_plan!r}")
+
+            check_stop()
+            current_child_id = launch_child(gr_id, launch_kind, child_plan)
+            boss_state["current_child_id"] = current_child_id
+            save_boss_state(state_dir, boss_state)
+            check_stop()
```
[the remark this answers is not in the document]
xbrianh: Fixed. After saving `current_child_id` to boss_state, there is now an explicit check for `_stop_requested` that calls `gremlins stop <child_id>` before exiting, covering the pre-wait window.

skills/bossgremlin/bossgremlin.py
```diff
@@ -0,0 +1,466 @@
...
+            # Pipeline or land failure → rescue
+            set_stage(gr_id, "rescuing")
+            if not rescue_child(current_child_id):
+                bail_reason = get_child_bail_reason(current_child_id)
+                log(f"rescue refused for {current_child_id} ({bail_reason or 'no bail_reason'})")
+                boss_state["children"].append({
+                    "id": current_child_id,
+                    "outcome": f"bailed:{bail_reason}" if bail_reason else "bailed",
+                })
+                boss_state["current_child_id"] = current_child_id
```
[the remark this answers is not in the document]
xbrianh: Fixed. The rescue-refusal path now sets `boss_state["current_child_id"] = None` before calling `die()`, so a subsequent resume sees no in-flight child and starts cleanly from handoff.

skills/gremlins/gremlins.py
```diff
@@ -1390,6 +1394,9 @@ def do_land(target: str, force: bool = False) -> bool:
         return _land_local(gr_id, sf, wdir, state)
     elif kind == "ghgremlin":
         return _land_gh(gr_id, sf, wdir, state, force=force)
+    elif kind == "bossgremlin":
+        print(f"bossgremlin chains complete automatically — use '/gremlins close {gr_id}' to hide it")
```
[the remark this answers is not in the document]
xbrianh: Fixed. Message now reads: `use 'gremlins close <id>' (or '/gremlins close <id>') to hide it`.

PR #47, opened 2026-04-24
Improve child gremlin identity in /gremlins output

xbrianh:
## Summary

- Child gremlin plans now start with a meaningful `# <title>` heading, so their ID slug and description are derived from the actual task rather than the generic section header "Context"
- The ID column is never truncated — the boss marker that previously crowded it is moved to a dedicated `BOSS` column
- New `BOSS` column shows the parent gremlin's ID for child gremlins, empty for top-level gremlins

## Test plan

- [ ] Run a boss gremlin chain and verify child gremlin IDs and descriptions reflect the task title from the handoff-generated child plan
- [ ] Confirm the `BOSS` column appears in `/gremlins` output and shows the boss ID for child gremlins
- [ ] Confirm the `ID` column is never truncated regardless of slug length

### DIFF
--- a/skills/gremlins/gremlins.py
+++ b/skills/gremlins/gremlins.py
@@ -30,7 +30,7 @@ STATE_ROOT = os.path.join(
     "claude-gremlins",
 )
 
-FMT = "%-5s  %-47s  %-22s  %-28s  %-5s  %s"
+FMT = "%-5s  %-47s  %-22s  %-28s  %-5s  %-20s  %s"
 
 # Headless rescue caps. The attempt cap is shared across interactive and
 # headless rescues — both check `rescue_count`, but interactive only warns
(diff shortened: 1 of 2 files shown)


skills/gremlins/gremlins.py
```diff
@@ -263,14 +263,13 @@ def build_row(gr_id, sf, wdir, state, live):
     age = humanize_age(started_at)
     sid = display_id(gr_id)
     parent_id = state.get("parent_id") or ""
-    if parent_id:
-        boss_marker = f" [boss:{display_id(parent_id)}]"
-        sid = (sid + boss_marker)[:47]
+    boss_disp = display_id(parent_id) if parent_id else ""
```
xbrianh: New-format boss IDs (slug + 6-char hex) can be up to ~47 chars, which overflows the `%-20s` BOSS column and shifts DESCRIPTION out of alignment. Cap it to match the column width:

```python
boss_disp = display_id(parent_id)[:20] if parent_id else ""
```

xbrianh (commented):
Solid changes overall — the handoff template fix is straightforward and the BOSS column is a clean replacement for the truncating inline marker. One alignment issue: `boss_disp` is not capped to match the `%-20s` column width, so new-format boss IDs (slug + hex, up to ~47 chars) will push the DESCRIPTION column rightward. Worth a one-line cap.

skills/gremlins/gremlins.py
xbrianh: Good catch — capped to `[:20]` to match the column width.

PR #52, opened 2026-04-24
Add --model flag to ghgremlin.sh

xbrianh:
Passes `--model` through to all inner `claude` invocations so callers can override the model without editing the script.

- Title-generation `claude -p` (inside `--plan` file-source branch) now inherits `--model` for full model consistency.
- `MODEL` is persisted to `state.json` so `--resume-from` invocations automatically restore the original model.

### DIFF
--- a/skills/ghgremlin/ghgremlin.sh
+++ b/skills/ghgremlin/ghgremlin.sh
@@ -52,7 +52,8 @@ progress_tee() {
 REF=""
 RESUME_FROM=""
 PLAN_SOURCE=""
-USAGE='usage: ghgremlin.sh [-r <ref>] [--resume-from <stage>] [--plan <path|issue-ref>] "<instructions>"'
+MODEL=""
+USAGE='usage: ghgremlin.sh [-r <ref>] [--resume-from <stage>] [--plan <path|issue-ref>] [--model <model>] "<instructions>"'
 while [[ $# -gt 0 ]]; do
   case "$1" in
     -r)
@@ -64,6 +65,9 @@ while [[ $# -gt 0 ]]; do
     --plan)
       [[ $# -ge 2 ]] || die "$USAGE"
       PLAN_SOURCE="$2"; shift 2 ;;
+    --model)
+      [[ $# -ge 2 ]] || die "$USAGE"
+      MODEL="$2"; shift 2 ;;
     --) shift; break ;;
     -*) die "unknown flag: $1" ;;
     *) break ;;
@@ -275,6 +279,7 @@ $ISSUE_BODY") || die "--plan: title-generation agent failed"
 fi
 
 CLAUDE_FLAGS=(--permission-mode bypassPermissions --output-format stream-json --verbose)
+[[ -n "$MODEL" ]] && CLAUDE_FLAGS+=(--model "$MODEL")
 # Stages run sequentially; the wait-copilot sleep 20 loop should not be extended beyond ~5 min intervals or the Anthropic prompt cache TTL expires and cache benefits between the review and address stages are lost.
 
 # Extract a URL from a Bash-tool_result event matching a regex, preferring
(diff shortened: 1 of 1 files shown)


xbrianh (commented):
Two scope issues:

**Skipped plan item:** The sole task in the plan — updating the child plan template in `skills/handoff/handoff.py` `build_prompt()` to prepend an H1 title — is entirely absent from the diff. `handoff.py` is not touched at all.

**Hitchhiker:** The diff adds a `--model` flag to `skills/ghgremlin/ghgremlin.sh`. This has no connection to the plan, which is narrowly scoped to a template string change in `handoff.py`.

The diff appears to belong to a different task (the recent `--model` flag commit visible in git history). The intended change has not landed.

skills/ghgremlin/ghgremlin.sh
```diff
@@ -275,6 +279,7 @@ $ISSUE_BODY") || die "--plan: title-generation agent failed"
 fi
 
 CLAUDE_FLAGS=(--permission-mode bypassPermissions --output-format stream-json --verbose)
+[[ -n "$MODEL" ]] && CLAUDE_FLAGS+=(--model "$MODEL")
```
xbrianh, line 289: The title-generation `claude -p` (~line 207, inside the `--plan` file-source branch) builds its own flag list (`--permission-mode bypassPermissions --output-format text`) rather than using `CLAUDE_FLAGS`, so it won't pick up `--model`. Probably fine for a one-shot title call, but worth noting if callers expect full model consistency across the entire run.

xbrianh (commented):
Clean, minimal change — the flag is correctly parsed and wired into `CLAUDE_FLAGS` so it propagates uniformly to all pipeline stages. Two minor notes:

1. The title-generation `claude -p` at ~line 207 (inside the `--plan` file-source branch) assembles its own flags directly rather than using `CLAUDE_FLAGS`, so it won't inherit `--model`. Probably fine for a cheap one-shot call, but worth knowing if strict model consistency across the full run is expected.

2. `MODEL` is not persisted to `state.json`, so a `--resume-from` invocation must re-supply `--model` to keep consistent behavior — same pattern as `-r` / `REF`, but resume callers should be aware.

skills/ghgremlin/ghgremlin.sh
xbrianh: Fixed — the title-gen call now builds its own flag array and conditionally appends `--model` when set, matching the approach used for `CLAUDE_FLAGS`. Also persisted `MODEL` to `state.json` and added a restore-on-resume block so `--resume-from` invocations pick up the original model automatically.

skills/ghgremlin/ghgremlin.sh
```diff
@@ -52,7 +52,8 @@ progress_tee() {
 REF=""
 RESUME_FROM=""
 PLAN_SOURCE=""
-USAGE='usage: ghgremlin.sh [-r <ref>] [--resume-from <stage>] [--plan <path|issue-ref>] "<instructions>"'
+MODEL=""
+USAGE='usage: ghgremlin.sh [-r <ref>] [--resume-from <stage>] [--plan <path|issue-ref>] [--model <model>] "<instructions>"'
```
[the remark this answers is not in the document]
xbrianh: Agreed — removed `Closes #51` from the PR description. Issue #51 (H1 title in handoff.py child plan template) was already fixed in #47 (commit ca00314). This PR only adds the `--model` flag to ghgremlin.sh.

xbrianh:
Re the scope concerns: the H1 title fix for `handoff.py` was already landed in #47 (commit ca00314), so it's present on main and on this branch — just not visible in this PR's diff. Removed `Closes #51` from the description since that fix belongs to #47.

The two gaps you flagged in the review comment are now addressed: the title-gen `claude -p` call inherits `--model`, and `MODEL` is persisted to `state.json` and restored automatically on `--resume-from`.

PR #57, opened 2026-04-24
Restrict rescue agent to state.json-only fixes

### LINKED ISSUE #55: Restrict rescue agent to state.json-only fixes; remove worktree access
# Mechanically restrict rescue agent to state.json-only fixes

## Context

The rescue agent in `gremlins.py` currently runs with `cwd=workdir` (the full gremlin worktree) and receives the original task `instructions` and `artifact_paths` (including `plan.md`) in its prompt. This gives it enough context to implement new features rather than just fix pipeline failures — which is exactly what happened: a rescue for an "unknown flag: --model" crash resulted in the agent implementing `--model` support in `ghgremlin.sh` instead of just stripping the bad flag from `state.json`.

## Approach

Remove the ability for the rescue agent to act on the worktree at all. Its only permitted fix surface is `state.json`. If fixing requires anything beyond patching `state.json`, the failure is `unsalvageable` for headless rescue.

Concretely:

1. Strip `instructions` and `artifact_paths` from the rescue prompt (`build_rescue_prompt`). The rescue agent does not need to know what the gremlin was implementing.
2. Run both headless and interactive Phase A with `cwd` set to a temporary scratch directory (not the worktree). The agent has no filesystem path to the worktree and cannot navigate to it.
3. Copy only the log tail into the scratch dir as a readable file (so the agent can read it naturally if needed, not just as inline text).
4. Pass `state.json`'s absolute path explicitly in the prompt as the one and only file the agent may edit.
5. Update the prompt instructions: the only permitted fix is editing `state.json` (e.g. removing bad `pipeline_args`). Anything requiring worktree access → declare `unsalvageable`.
6. Clean up the scratch dir after Phase A completes (success or failure).

## Tasks

- [ ] Task 1: In `build_rescue_prompt`, remove the `instructions` field and `artifacts_section` from the prompt text. Remove the worktree path from the "fixable" description.
- [ ] Task 2: Update the headless and interactive prompt body to state: "Your only permitted action is editing `state.json` at `{state_file_path}`. If the fix requires anything else, declare `unsalvageable`."
- [ ] Task 3: In `_run_headless_phase_a` and the interactive `subprocess.run` call, change `cwd` to a temporary scratch directory. Write the log tail to a file in that scratch dir so the agent can reference it. Clean up the scratch dir after Phase A exits.
- [ ] Task 4: Pass the absolute path to `state.json` into `build_rescue_prompt` and include it in the prompt.

## Open questions

- Interactive rescue currently lets a human watch Phase A live and Ctrl-C if it goes wrong. Restricting `cwd` there too is safe since the human is the last line of defense — the restriction is still an improvement even with a human watching.

xbrianh:
## Summary

- Remove `instructions` and `artifact_paths` from the rescue prompt so the agent has no context about what the gremlin was implementing
- Phase A now runs with `cwd` set to a temporary scratch directory (not the worktree), so the agent has no filesystem path to navigate to
- `state.json` absolute path is passed explicitly as the only file the agent may edit; anything requiring worktree access is declared `unsalvageable`
- Log tail is written to `scratch_dir/gremlin.log`; scratch dir is cleaned up via `try/finally` after Phase A exits

## Test plan

- [ ] Run `/gremlins rescue` on a dead gremlin; confirm Phase A prompt no longer contains instructions or artifact paths
- [ ] Confirm the agent's working directory is a temp scratch dir, not the worktree
- [ ] Confirm the scratch dir is removed after Phase A completes (success or failure)
- [ ] Simulate a `pipeline_args` failure; confirm headless rescue patches `state.json` and declares `fixed`
- [ ] Simulate a failure requiring worktree changes; confirm agent declares `unsalvageable`

Closes #55

### DIFF
--- a/skills/gremlins/gremlins.py
+++ b/skills/gremlins/gremlins.py
@@ -417,7 +417,7 @@ def do_stop(target: str) -> bool:
     return True
 
 
-def build_rescue_prompt(state, wdir, log_tail, state_file_path, log_file_path,
+def build_rescue_prompt(state, log_tail, state_file_path, log_file_path,
                         *, headless=False, marker_path=None):
     kind = state.get("kind", "localgremlin")
     stage = state.get("stage") or "unknown"
@@ -438,7 +438,7 @@ Stage order for {kind}: {' → '.join(stages)}
 
 ## Failure log (last ~200 lines)
 
-(Full log also available at: {log_file_path})
+(Log tail also written to: {log_file_path})
 
 ```
 {log_tail_safe}
@@ -482,8 +482,8 @@ Constraints for headless mode:
 ## What to do
 
 1. Diagnose the failure from the log above.
-2. Your **only permitted fix** is editing `state.json` at `{state_file_path}` (e.g. removing bad `pipeline_args` entries). If fixing the failure requires anything beyond patching `state.json`, say so clearly in your final message and leave everything untouched; the operator watching Phase A will decide whether to proceed.
-3. STOP. Do NOT re-run the failed stage or any remaining stages yourself — after you exit, a background resume will relaunch the gremlin pipeline starting at {stage} and complete the rest automatically.
+2. Your **only permitted fix** is editing `state.json` at `{state_file_path}` (e.g. removing bad `pipeline_args` entries). If fixing the failure requires anything beyond patching `state.json`, say so clearly in your final message and leave everything untouched; the operator must Ctrl-C during Phase A to prevent the automatic resume.
+3. STOP. Do NOT re-run the failed stage or any remaining stages yourself — after you exit successfully, a background resume will relaunch the gremlin pipeline starting at {stage} and complete the rest automatically.
 
 Your working directory is a scratch space. The only file outside it you may touch is `state.json` at `{state_file_path}`.
 """
@@ -710,21 +710,24 @@ def do_rescue(target: str, headless: bool = False) -> bool:
     # Create a scratch dir so Phase A runs isolated from the worktree.
     # The log tail is written there as a readable file; state.json is passed
     # by absolute path as the only file the agent may edit.
-    scratch_dir = tempfile.mkdtemp(prefix="gremlin-rescue-")
-    scratch_log = os.path.join(scratch_dir, "gremlin.log")
+    # mkdtemp is inside the try so any subsequent exception triggers cleanup.
+    scratch_dir = None
     try:
-        with open(scratch_log, "w", encoding="utf-8") as fh:
-            fh.write(log_tail)
-    except OSError:
-        scratch_log = log_path  # fallback: reference the original log path
+        scratch_dir = tempfile.mkdtemp(prefix="gremlin-rescue-")
+        scratch_log = os.path.join(scratch_dir, "gremlin.log")
+        try:
+            with open(scratch_log, "w", encoding="utf-8") as fh:
+                fh.write(log_tail)
+        except OSError:
+            scratch_log = log_path  # fallback: reference the original log path
 
-    prompt = build_rescue_prompt(state, wdir, log_tail, sf, scratch_log,
-                                 headless=headless, marker_path=marker_path)
+        prompt = build_rescue_prompt(state, log_tail, sf, scratch_log,
+                                     headless=headless, marker_path=marker_path)
 
-    print(f"Rescuing gremlin {gr_id} (stage: {stage}, liveness: {live})")
-    print(f"Working directory: {workdir}")
+        print(f"Rescuing gremlin {gr_id} (stage: {stage}, liveness: {live})")
+        print(f"Gremlin workdir: {workdir}")
+        print(f"Agent scratch dir: {scratch_dir}")
 
-    try:
         if headless:
             print(
                 f"Phase A (headless): running diagnosis agent "
@@ -775,7 +778,8 @@ def do_rescue(target: str, headless: bool = False) -> bool:
                 print(f"Inspect the log at {log_path} and worktree at {workdir} for details.")
                 return False
     finally:
-        shutil.rmtree(scratch_dir, ignore_errors=True)
+        if scratch_dir is not None:
+            shutil.rmtree(scratch_dir, ignore_errors=True)
 
     # Phase B: hand off to launch.sh --resume so the remaining stages run in the
     # background under the same GR_ID. launch.sh patches state.json, clears the
--- a/skills/gremlins/gremlins.py
+++ b/skills/gremlins/gremlins.py
@@ -17,6 +17,7 @@ import shutil
 import signal
 import subprocess
 import sys
+import tempfile
 import time
 
 # ---------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


xbrianh (commented):

skills/gremlins/gremlins.py
```diff
@@ -416,46 +417,28 @@ def do_stop(target: str) -> bool:
     return True
 
 
-def build_rescue_prompt(state, wdir, log_tail, artifact_paths,
+def build_rescue_prompt(state, wdir, log_tail, state_file_path, log_file_path,
```
xbrianh: `wdir` is accepted but never referenced in the new function body — it can be removed from the signature (and the call site at line 721 updated accordingly). Leaving a dead parameter makes the API dishonest about what information is actually used.

skills/gremlins/gremlins.py
```diff
@@ -733,61 +707,75 @@ def do_rescue(target: str, headless: bool = False) -> bool:
...
+        with open(scratch_log, "w", encoding="utf-8") as fh:
+            fh.write(log_tail)
+    except OSError:
+        scratch_log = log_path  # fallback: reference the original log path
+
+    prompt = build_rescue_prompt(state, wdir, log_tail, sf, scratch_log,
                                  headless=headless, marker_path=marker_path)
 
     print(f"Rescuing gremlin {gr_id} (stage: {stage}, liveness: {live})")
     print(f"Working directory: {workdir}")
```
xbrianh: `print(f"Working directory: {workdir}")` now prints the gremlin's original worktree path, but Phase A actually runs in `scratch_dir`. Someone triaging output will look in the wrong place. Consider replacing with two lines:
```python
print(f"Gremlin workdir: {workdir}")
print(f"Agent scratch dir: {scratch_dir}")
```

skills/gremlins/gremlins.py
```diff
@@ -733,61 +707,75 @@ def do_rescue(target: str, headless: bool = False) -> bool:
         ts = datetime.datetime.utcnow().strftime("%Y%m%dT%H%M%SZ")
         marker_path = os.path.join(artifacts_dir, f"rescue-{ts}.done")
 
-    prompt = build_rescue_prompt(state, wdir, log_tail, artifact_paths,
+    # Create a scratch dir so Phase A runs isolated from the worktree.
+    # The log tail is written there as a readable file; state.json is passed
+    # by absolute path as the only file the agent may edit.
+    scratch_dir = tempfile.mkdtemp(prefix="gremlin-rescue-")
```
xbrianh: `scratch_dir` is created before the `try/finally` that cleans it up. Any exception thrown between `mkdtemp` and line 727 (`try:`) — e.g. an unexpected error inside `build_rescue_prompt` — will leak the temp directory. Move `scratch_dir = tempfile.mkdtemp(...)` to just inside the `try` block, or wrap the entire body from `mkdtemp` onwards in the `try`.

xbrianh (commented):
Clean, well-scoped change that correctly tightens the rescue agent's blast radius. Two actionable issues and one minor one worth cleaning up.

skills/gremlins/gremlins.py
xbrianh: Fixed — removed `wdir` from the `build_rescue_prompt` signature and updated the call site.

skills/gremlins/gremlins.py
xbrianh: Fixed — replaced the single `Working directory` line with `Gremlin workdir: {workdir}` and `Agent scratch dir: {scratch_dir}` as suggested.

skills/gremlins/gremlins.py
xbrianh: Fixed — `scratch_dir = None` is set before the `try`, then `mkdtemp` is the first statement inside the `try` block. The `finally` guards against None so it still handles the mkdtemp-itself-failing case cleanly.

skills/gremlins/gremlins.py
```diff
@@ -416,46 +417,28 @@ def do_stop(target: str) -> bool:
...
-
-## Gremlin reference
-
-Gremlin script: {gremlin_script}
 Stage order for {kind}: {' → '.join(stages)}
 
-{artifacts_section}## Failure log (last ~200 lines)
+## Failure log (last ~200 lines)
+
+(Full log also available at: {log_file_path})
```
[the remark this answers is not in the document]
xbrianh: Fixed — changed to 'Log tail also written to: {log_file_path}' to accurately reflect that the file contains only the tail.

skills/gremlins/gremlins.py
```diff
@@ -733,61 +707,75 @@ def do_rescue(target: str, headless: bool = False) -> bool:
...
+    # by absolute path as the only file the agent may edit.
+    scratch_dir = tempfile.mkdtemp(prefix="gremlin-rescue-")
+    scratch_log = os.path.join(scratch_dir, "gremlin.log")
+    try:
+        with open(scratch_log, "w", encoding="utf-8") as fh:
+            fh.write(log_tail)
+    except OSError:
+        scratch_log = log_path  # fallback: reference the original log path
+
+    prompt = build_rescue_prompt(state, wdir, log_tail, sf, scratch_log,
```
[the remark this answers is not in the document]
xbrianh: Fixed — same as the owner's comment on this thread: `mkdtemp` is now the first statement inside the `try` block.

skills/gremlins/gremlins.py
```diff
@@ -492,19 +475,17 @@ def build_rescue_prompt(state, wdir, log_tail, artifact_paths,
...
-3. Write a brief rescue note to `{rescue_note_path}` describing what failed and what you fixed.
-4. STOP. Do NOT re-run the failed stage or any remaining stages yourself — after you exit, a background resume will relaunch the gremlin pipeline starting at {stage} and complete the rest automatically. If you conclude the failure is unsalvageable, say so clearly in your final message and in the rescue note, and leave the worktree untouched; the operator watches Phase A output and will Ctrl-C / decline resume if they agree.
+2. Your **only permitted fix** is editing `state.json` at `{state_file_path}` (e.g. removing bad `pipeline_args` entries). If fixing the failure requires anything beyond patching `state.json`, say so clearly in your final message and leave everything untouched; the operator watching Phase A will decide whether to proceed.
+3. STOP. Do NOT re-run the failed stage or any remaining stages yourself — after you exit, a background resume will relaunch the gremlin pipeline starting at {stage} and complete the rest automatically.
```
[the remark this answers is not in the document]
xbrianh: Fixed — updated the interactive prompt to say 'the operator must Ctrl-C during Phase A to prevent the automatic resume', which matches the actual behavior.

skills/gremlins/gremlins.py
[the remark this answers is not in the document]
xbrianh: Fixed — the print now shows both 'Gremlin workdir' and 'Agent scratch dir' on separate lines.

skills/gremlins/gremlins.py
```diff
@@ -733,61 +707,75 @@ def do_rescue(target: str, headless: bool = False) -> bool:
...
+    scratch_dir = tempfile.mkdtemp(prefix="gremlin-rescue-")
+    scratch_log = os.path.join(scratch_dir, "gremlin.log")
+    try:
+        with open(scratch_log, "w", encoding="utf-8") as fh:
+            fh.write(log_tail)
+    except OSError:
+        scratch_log = log_path  # fallback: reference the original log path
+
+    prompt = build_rescue_prompt(state, wdir, log_tail, sf, scratch_log,
                                  headless=headless, marker_path=marker_path)
```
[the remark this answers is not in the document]
xbrianh: Noted — this is a valid architectural observation, but it's out of scope for this PR. The goal here was reducing the agent's context (no instructions, no artifact paths, scratch cwd), not mechanically preventing all filesystem access. A sanitized-copy approach would be a meaningful follow-on change.

PR #62, opened 2026-04-25
Fix handoff git context for local and gh chains

### LINKED ISSUE #59: handoff: collects git context from the wrong worktree (local chains see no progress; gh chains see stale main)
## Symptom

The handoff agent (`skills/handoff/handoff.py`) reads stale or wrong git context, causing it to either re-emit child plans for already-landed work or to misjudge chain progress entirely. Child gremlins launched from the resulting plans then die at the implement guard (`implementation step produced no changes; refusing to open empty PR`).

## Root cause: handoff git-context source differs from where work actually lands

`bossgremlin.run_handoff` (`skills/bossgremlin/bossgremlin.py:149-215`) always invokes handoff with:

```
cwd       = boss.state.project_root   # the user's repo
--rev     = target_branch             # e.g. "main", captured from project_root HEAD at boss start
--base    = chain_base_ref            # SHA of project_root HEAD at boss start
```

Handoff then computes `git diff <merge_base(rev,base)>..<rev>` against that local ref. But where chain commits actually land depends on chain kind:

| Chain kind | Where children's work lands | Where handoff currently looks |
|---|---|---|
| **local** | Boss's workdir HEAD (detached) — `_land_local` squashes child branch into `cwd=child.project_root`, and child.project_root is set by launch.sh to the boss's workdir because the boss runs with cwd=workdir and launch.sh resolves PROJECT_ROOT via `git rev-parse --show-toplevel`. | `refs/heads/main` in the user's repo — **never advances** during a local chain. |
| **gh**   | `origin/main`, mirrored locally via `_fast_forward_main` (`skills/gremlins/gremlins.py:864`). | `refs/heads/main` in the user's repo — correct as long as `_fast_forward_main` succeeded. |

Consequences:

- **Local chain:** every handoff sees zero diff since chain start, regardless of how many children landed. It either declares `chain-done` on the first call or keeps emitting the same `next-plan` over and over. The boss's actual progress is invisible.
- **GH chain:** mostly works, but is fragile. `_fast_forward_main` failures are warnings only, and anything that lands on `origin/main` outside `gremlins.py land` (manual GH merges, parallel tooling, another machine) leaves local `main` stale. Concrete instance from this repo: PR #47 was merged via the GitHub UI on the `brian-fixes` branch, `_fast_forward_main` never ran, and a subsequent handoff produced `/tmp/fix-handoff-child-plan-h1.md` (mtime *after* the merge) asking for the exact change #47 already contained. Two child gh gremlins (#54, #56) launched from that plan died at the implement guard.

## Suggested fix

Confine the fix to `bossgremlin.run_handoff` — handoff.py needs no changes. Its existing `--rev` / `--base` flags accept any ref expression (`HEAD`, `origin/main`, a SHA), and `cwd` is already controlled by the boss via `subprocess.run(..., cwd=...)`. Per-chain-kind dispatch:

- **Local chains:** invoke handoff with `cwd=boss.workdir` and omit `--rev` (defaults to `HEAD` in `collect_git_context`). Keep `--base <chain_base_ref>`. The diff becomes `chain_base..boss_workdir_HEAD`, which actually reflects the squashed-in children.
- **GH chains:** before invoking handoff, run `git fetch origin <target_branch>` in `cwd=project_root`. Then invoke handoff with `cwd=project_root, --rev origin/<target_branch>, --base <chain_base_ref>`. Using the remote-tracking ref directly bypasses any `_fast_forward_main` skips. Optionally surface `_fast_forward_main` failures as errors so the boss halts loudly when a fast-forward goes wrong (separate change in `skills/gremlins/gremlins.py`).

Cosmetic note: in a detached worktree, `git rev-parse --abbrev-ref HEAD` returns the literal `"HEAD"`, so handoff's prompt will show `Branch: HEAD` for local chains. Functional but ugly; an optional one-line tweak in `collect_git_context` could relabel it as `(detached at <sha[:12]>)`.

## Out of scope

Orthogonal bug: ghgremlin 96cf38 was launched with `/tmp/fix-handoff-child-plan-h1.md` (the H1-fix plan) but produced PR #52 (`Add --model flag to ghgremlin.sh`) closing #51. The implementation diverged from the supplied plan — worth filing separately if reproducible.

xbrianh:
## Summary
- Local chains: invoke handoff with `cwd=boss.workdir` and no `--rev` so the diff reflects children squashed into the boss's own workdir HEAD (the user's `refs/heads/main` never advances during a local chain).
- GH chains: `git fetch origin <target_branch>` in `project_root`, then invoke handoff with `cwd=project_root` and `--rev origin/<target_branch>` so handoff sees PRs that landed on the remote even when `_fast_forward_main` was skipped (manual UI merges, another machine).
- Cosmetic: relabel a literal `"HEAD"` (detached worktree) as `(detached at <sha[:12]>)` in `handoff.collect_git_context` so the prompt no longer reads `Branch: HEAD`.

Confined to `bossgremlin.run_handoff` and a one-line tweak in `handoff.collect_git_context`; the existing `--rev` / `--base` flags and `cwd` plumbing are unchanged.

Closes #59

## Test plan
- [ ] Run a local chain with multiple steps; confirm handoff sees each landed step's diff (no repeated/stale `next-plan`, no premature `chain-done`).
- [ ] Run a gh chain where a PR is merged via the GitHub UI (skipping `_fast_forward_main`); confirm the next handoff sees the merged work and does not re-emit it.
- [ ] Verify a hard fail when `git fetch origin <target_branch>` errors on a gh chain.
- [ ] Verify the prompt shows `Branch: (detached at …)` rather than `Branch: HEAD` in a local chain.

### DIFF
--- a/skills/bossgremlin/bossgremlin.py
+++ b/skills/bossgremlin/bossgremlin.py
@@ -32,6 +32,7 @@ STATE_ROOT = os.path.join(
 
 POLL_INTERVAL = 5  # seconds between finished-marker polls
 HANDOFF_TIMEOUT = int(os.environ.get("BOSSGREMLIN_HANDOFF_TIMEOUT", "3600"))
+HANDOFF_FETCH_TIMEOUT = int(os.environ.get("BOSSGREMLIN_HANDOFF_FETCH_TIMEOUT", "60"))
 
 _current_proc = None
 _stop_requested = False
@@ -182,11 +183,17 @@ def run_handoff(gr_id: str, state_dir: str, boss_state: dict,
             die("gh chain has no target branch — cannot resolve remote ref for handoff")
         # Refresh the remote-tracking ref so we see PRs that landed on the
         # remote (including merges done via the GitHub UI or another machine,
-        # which never trigger _fast_forward_main locally).
-        fetch = subprocess.run(
-            ["git", "fetch", "origin", target_branch],
-            capture_output=True, text=True, cwd=project_root,
-        )
+        # which never trigger _fast_forward_main locally). Bound the fetch so
+        # an unreachable origin can't stall the chain indefinitely between
+        # handoffs.
+        try:
+            fetch = subprocess.run(
+                ["git", "fetch", "origin", target_branch],
+                capture_output=True, text=True, cwd=project_root,
+                timeout=HANDOFF_FETCH_TIMEOUT,
+            )
+        except subprocess.TimeoutExpired:
+            die(f"git fetch origin {target_branch} timed out after {HANDOFF_FETCH_TIMEOUT}s")
         if fetch.returncode != 0:
             die(f"git fetch origin {target_branch} failed: {fetch.stderr.strip()}")
         handoff_cwd = project_root
--- a/skills/bossgremlin/bossgremlin.py
+++ b/skills/bossgremlin/bossgremlin.py
@@ -147,11 +147,16 @@ def save_boss_state(state_dir: str, boss_state: dict) -> None:
 
 
 def run_handoff(gr_id: str, state_dir: str, boss_state: dict,
-                project_root: str, model: str, target_branch: str) -> tuple:
+                project_root: str, boss_workdir: str, model: str) -> tuple:
     """Run handoff agent. Returns (exit_state, signal dict).
 
     Updates boss_state in place (handoff_count, current_plan, handoff_records).
     Calls die() on infrastructure failure.
+
+    Children of a local chain squash-land into the boss's own workdir HEAD,
+    while children of a gh chain push to origin/<target_branch>. Pick the
+    right cwd/rev so handoff sees the actually-landed work, not a stale ref
+    in the user's repo that may never advance during the chain.
     """
     set_stage(gr_id, "handoff")
     handoff_script = os.path.expanduser("~/.claude/skills/handoff/handoff.py")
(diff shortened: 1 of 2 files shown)


xbrianh (commented):

skills/bossgremlin/bossgremlin.py
```diff
@@ -163,22 +168,44 @@ def run_handoff(gr_id: str, state_dir: str, boss_state: dict,
     signal_path = os.path.join(state_dir, f"handoff-{n:03d}.state.json")
     current_plan = boss_state["current_plan"]
     base_ref = boss_state["chain_base_ref"]
+    chain_kind = boss_state.get("chain_kind")
+    target_branch = boss_state.get("target_branch", "")
```
xbrianh, line 173: Pre-existing semantic worth surfacing: `target_branch` is captured at chain start as the user's current branch (`get_current_branch(project_root)` in `main()`), but ghgremlin children create their PRs *from the default branch* (see `ghgremlin.sh:463` — `"Create a new branch from the default branch..."`). So `origin/<target_branch>` only matches "where PRs landed" when the user happens to be on the default branch when launching the chain.

If they launch a gh chain from a feature branch:
- best case, `origin/<feature>` exists but is stale and never gets PR merges, so handoff sees no diff and emits `chain-done` prematurely;
- worst case, the branch doesn't exist on origin at all and `git fetch origin <feature>` hard-fails on the first handoff.

The old code had the same blind spot (it passed `target_branch` straight into `--rev`), so this PR doesn't regress anything — but since the PR is reshaping the gh-chain story, it'd be worth either documenting the assumption explicitly ("gh chains must be launched from the branch the PRs will merge into") or resolving the default branch via `gh repo view --json defaultBranchRef` once at chain start. Not a blocker for the reported bug.

skills/bossgremlin/bossgremlin.py
```diff
@@ -163,22 +168,44 @@ def run_handoff(gr_id: str, state_dir: str, boss_state: dict,
...
+        handoff_cwd = boss_workdir
+        rev_label = "HEAD"
+    elif chain_kind == "gh":
+        if not target_branch:
+            die("gh chain has no target branch — cannot resolve remote ref for handoff")
+        # Refresh the remote-tracking ref so we see PRs that landed on the
+        # remote (including merges done via the GitHub UI or another machine,
+        # which never trigger _fast_forward_main locally).
+        fetch = subprocess.run(
+            ["git", "fetch", "origin", target_branch],
```
xbrianh: Nit: this fetch has no timeout. If `origin` is unreachable or slow, the boss hangs indefinitely between handoffs (the SIGTERM path in `_sigterm_handler` does forward to the current proc, so a stop request will still escape, but the chain's effective stall time is bounded only by the operator). The agent invocation a few lines down uses `HANDOFF_TIMEOUT`; you could give the fetch a much smaller bound (e.g. 60s) to cap the worst case without affecting normal runs.

xbrianh (commented):
Targets both reported issues cleanly: local chains run handoff against `boss.workdir` HEAD (where children squash-land), and gh chains fetch then inspect `origin/<target_branch>` so manual UI merges become visible. The detached-HEAD label fix in `collect_git_context` is a nice quality-of-life cleanup. Two non-blocking observations below about the assumptions baked into `target_branch` and the unbounded fetch.

skills/bossgremlin/bossgremlin.py
xbrianh: Added a 60s bound (override via `BOSSGREMLIN_HANDOFF_FETCH_TIMEOUT`) in 1063cb9. `subprocess.TimeoutExpired` routes through `die()` with a clear message, matching the existing fetch-error handling.

skills/bossgremlin/bossgremlin.py
xbrianh: Acknowledged — leaving as a follow-up since the reviewer flagged it non-blocking and the underlying assumption (`target_branch` = the user's current branch at chain start) predates this PR. The clean fix is to resolve the default branch via `gh repo view --json defaultBranchRef` once at chain start and persist it in `boss_state`, separately from the user's working branch; that's a small but distinct change worth tracking in its own issue rather than expanding scope here.

PR #65, opened 2026-04-25
Clear completed tasks from /handoff rolling plan

### LINKED ISSUE #64: Drop completed tasks from `/handoff` rolling plan output
# `/handoff` — clear finished work from the rolling plan

## Context

`/handoff` is the per-step decision agent in the `/bossgremlin` chained workflow. On each invocation it produces two plan documents:

1. A **rolling plan** (`plan-NNN.md`) — the document that the boss feeds back into the next `/handoff` invocation.
2. A **child plan** (`plan-NNN-child.md`) — the document handed to the next gremlin to implement.

Today the rolling plan keeps completed items, marking them `[x]`. As the chain progresses, the rolling plan grows: every invocation of `/handoff` re-reads the full history of finished work. That information is no longer load-bearing — finished work is recorded in git, and the chain of versioned plan files (`plan.md` → `plan-001.md` → `plan-002.md` → …) already serves as an audit trail. Carrying it forward in the rolling plan only adds noise and risks pulling the next step's framing toward already-done work.

## Goals

- The rolling plan emitted by `/handoff` represents **only remaining work** — completed items are gone, not annotated.
- Future `/handoff` invocations and any downstream readers of the rolling plan see a slim document focused on what's left.
- The audit trail is preserved entirely through the existing per-invocation versioned files; no information is destroyed at the chain level.
- Child plans remain self-contained — a fresh gremlin given only the child plan can implement it without ever seeing the overarching goal of the chain.

## Non-goals

- Changing how `/handoff` decides exit state (`next-plan` / `chain-done` / `bail`).
- Changing the file-naming or signal-file contract.
- Changing the child plan structure (already self-contained today).
- Re-deriving an aggregate audit log from the per-version files. The chain of files *is* the audit trail; nothing new is needed.
- Changing how the original plan / spec is created (no change to `/ghplan`, `/design`, or upstream planners).

## Users

The `/bossgremlin` chain orchestrator and any human inspecting the rolling plan between chain steps.

## Behavior

When `/handoff` runs and determines exit state:

- **`next-plan`** — the rolling plan output contains only the tasks that have **not** yet been implemented. Completed tasks are removed entirely; no `[x]` markers, no struck-through entries, no "completed" appendix. The child plan is unchanged (already self-contained, scoped to the next chunk).

- **`chain-done`** — every task has landed. The rolling plan output is effectively empty of tasks. It should be a small document that makes it obvious the chain is finished (e.g., a brief note that all work is complete) rather than an empty file or one with a stale task list.

- **`bail`** — completed tasks are still removed from the rolling plan. The bail reason is recorded prominently. (Diagnostic context comes from the bail reason itself plus the prior versioned plan files, not from carrying `[x]` items forward.)

### What happens around the tasks

The rolling plan's surrounding sections (`## Context`, `## Approach`, `## Open questions`, etc.) should be pruned to match the remaining tasks:

- If a section's reason for existing was a now-completed task, drop it.
- Otherwise, keep or trim it so the rolling plan stays a coherent description of the *remaining* work.

The overarching goal of the chain belongs upstream (in the original spec or planner output), not in the rolling plan that flows step-to-step. `/handoff` should not propagate the original goal forward into rolling plans or into child plans.

### Edge cases

- **First invocation** — the input plan is the original spec/plan with all tasks unfinished. Nothing to clear; the rolling plan and the spec look essentially the same modulo any restructuring `/handoff` already does.
- **Partially completed task** — if a task is only partly landed, it stays in the rolling plan (possibly rewritten to reflect what remains). It is not dropped.
- **No tasks completed since last invocation** — rolling plan is unchanged in substance from the input; this is fine.
- **Plan with no `## Tasks` section** — out of scope; current behavior preserved.

## Constraints

- Backward compatible with `/bossgremlin` — same CLI flags, same exit codes, same signal-file shape.
- Inner agent prompt change only; no Python orchestration changes required unless something downstream depends on `[x]` markers in the rolling plan (verify during implementation).

## Acceptance criteria

- After a chain step where Task A landed and Task B did not, the resulting rolling plan contains Task B but not Task A — no `[x]`, no mention.
- A `chain-done` rolling plan contains no leftover task list.
- A `bail` rolling plan still drops completed tasks; the bail reason is at the top.
- Child plan structure and contents are unaffected by this change (manual diff against a current child plan shows no regression).
- Re-running an existing `/bossgremlin` chain end-to-end with the modified `/handoff` produces a sequence of progressively shorter rolling plans, each containing only outstanding work, with the chain still terminating on `chain-done`.

## Resolved decisions

- **`chain-done` rolling plan**: minimal. A short marker indicating the chain is complete; no leftover context or task list. The `.state.json` signal file carries the structured outcome.
- **`## Open questions` carry-forward**: carry forward unresolved entries across iterations; drop entries tied to completed tasks.

xbrianh:
## Summary
- Rewrites the `/handoff` inner-agent prompt so the rolling plan emitted between chain steps contains only remaining work — no `[x]` markers, no struck-through entries, no "completed" appendix. The chain of versioned plan files plus git history is the audit trail.
- Per exit state: `next-plan` prunes surrounding sections to match remaining tasks and carries forward unresolved `## Open questions`; `chain-done` produces a minimal completion marker; `bail` follows the same pruning rules as `next-plan` with a bail-reason banner added at the top.
- Tells the agent not to propagate the overarching chain goal into either the rolling plan or the child plan; child plans remain self-contained for their chunk.
- Updates the `handoff-NNN.md` artifact description in `skills/bossgremlin/SKILL.md` to reflect the new "remaining work only" semantics.

Scope note: this PR only changes the **handoff rolling plan** convention. Per-gremlin `plan.md` files still use `[x]` to mark completed tasks, and `gremlins.py`'s commit-message synthesis still parses those markers from `## Tasks` — that behavior is unchanged. No Python orchestration changes are needed: `bossgremlin.py` only feeds the rolling plan back into the next handoff invocation and never parses `[x]` itself.

## Test plan
- [ ] Run `/bossgremlin` end-to-end on a multi-step spec; confirm successive `handoff-NNN.md` files shrink as tasks land and contain no `[x]` entries.
- [ ] Verify a `chain-done` rolling plan is a brief completion marker with no leftover task list.
- [ ] Verify a `bail` rolling plan starts with the bail reason banner and contains only still-remaining tasks, with surrounding sections pruned.
- [ ] Diff a child plan against one produced by the previous version; confirm structure and contents are unchanged.

Closes #64

### DIFF
--- a/skills/handoff/handoff.py
+++ b/skills/handoff/handoff.py
@@ -149,7 +149,7 @@ Git diff since chain start:
      - Under `## Open questions`, carry forward unresolved entries; drop entries tied to completed tasks.
      - If a task is only partly landed, keep it (rewritten if needed to reflect what remains).
    - **`chain-done`**: minimal output. A short note that the chain is complete is enough — no leftover task list, no carried-over context. The signal file carries the structured outcome.
-   - **`bail`**: record the bail reason prominently at the top, then list the still-remaining tasks. Completed tasks are still dropped.
+   - **`bail`**: same pruning rules as `next-plan` (only remaining tasks, surrounding sections trimmed accordingly, unresolved `## Open questions` carried forward), with a bail-reason banner added prominently at the top.
 
 5. If exit state is **`next-plan`**, write a **child plan** to: `{child_plan_path}`
    - Use the standard localgremlin plan structure exactly:
--- a/skills/handoff/handoff.py
+++ b/skills/handoff/handoff.py
@@ -141,10 +141,15 @@ Git diff since chain start:
    - **`next-plan`**: some tasks remain unimplemented; the next gremlin should tackle them.
    - **`bail`**: something prevents safe continuation (broken state, incoherent plan, security issue, etc.).
 
-4. Write an **updated plan document** to: `{out_path}`
-   - Free-form markdown. A human reading it must be able to see what is done and what remains.
-   - Mark completed tasks (e.g. `[x]`) and leave remaining tasks as `[ ]`.
-   - If `bail`, record the bail reason prominently at the top.
+4. Write an **updated plan document** (the "rolling plan") to: `{out_path}`
+
+   The rolling plan describes only **remaining** work. Completed tasks are removed entirely — no `[x]` markers, no struck-through entries, no "completed" appendix. The chain of versioned plan files plus git history is the audit trail; the rolling plan does not repeat it. Do not propagate the overarching goal of the chain forward into the rolling plan — that lives upstream, in the original spec.
+
+   - **`next-plan`**: include only the tasks that are not yet implemented (still `[ ]`). Prune the surrounding sections (`## Context`, `## Approach`, `## Open questions`, etc.) to match: drop sections whose reason for existing was a now-completed task; keep or trim the rest so the document stays a coherent description of the remaining work.
+     - Under `## Open questions`, carry forward unresolved entries; drop entries tied to completed tasks.
+     - If a task is only partly landed, keep it (rewritten if needed to reflect what remains).
+   - **`chain-done`**: minimal output. A short note that the chain is complete is enough — no leftover task list, no carried-over context. The signal file carries the structured outcome.
+   - **`bail`**: record the bail reason prominently at the top, then list the still-remaining tasks. Completed tasks are still dropped.
 
 5. If exit state is **`next-plan`**, write a **child plan** to: `{child_plan_path}`
    - Use the standard localgremlin plan structure exactly:
@@ -165,7 +170,7 @@ Git diff since chain start:
      ## Open questions
      <risks or open questions, or "(none)" if there are none>
      ```
-   - The child plan must be self-contained — a fresh gremlin with only this file must know exactly what to implement.
+   - The child plan must be self-contained — a fresh gremlin with only this file must know exactly what to implement. Do not propagate the overarching goal of the chain into the child plan; scope it to the next chunk.
 
 6. Write the **signal marker** to: `{signal_path}`
    - Valid JSON, exactly this structure:
--- a/skills/bossgremlin/SKILL.md
+++ b/skills/bossgremlin/SKILL.md
@@ -39,7 +39,7 @@ Flags:
 ## Where artifacts go
 
 - `~/.local/state/claude-gremlins/<boss-id>/boss_state.json` — ordered list of child ids with outcomes, per-handoff records (timestamps, plan paths, exit states), chain base ref.
-- `~/.local/state/claude-gremlins/<boss-id>/handoff-001.md`, `handoff-002.md`, … — updated plan documents produced by each handoff invocation, showing which tasks are done.
+- `~/.local/state/claude-gremlins/<boss-id>/handoff-001.md`, `handoff-002.md`, … — rolling plan documents produced by each handoff invocation, each containing only the work still remaining at that point. The sequence of files is the audit trail of progression.
 - `~/.local/state/claude-gremlins/<boss-id>/handoff-001-child.md`, … — child plans passed to each child gremlin.
 - `~/.local/state/claude-gremlins/<boss-id>/handoff-001.state.json`, … — handoff signal files recording exit_state and reason.
 - `~/.local/state/claude-gremlins/<boss-id>/log` — boss lifecycle events (handoff invoked, child started, child landed, rescue attempts). Does not contain plan/diff/review content.


xbrianh (commented):

skills/handoff/handoff.py
```diff
@@ -141,10 +141,15 @@ def build_prompt(
...
+
+   - **`next-plan`**: include only the tasks that are not yet implemented (still `[ ]`). Prune the surrounding sections (`## Context`, `## Approach`, `## Open questions`, etc.) to match: drop sections whose reason for existing was a now-completed task; keep or trim the rest so the document stays a coherent description of the remaining work.
+     - Under `## Open questions`, carry forward unresolved entries; drop entries tied to completed tasks.
+     - If a task is only partly landed, keep it (rewritten if needed to reflect what remains).
+   - **`chain-done`**: minimal output. A short note that the chain is complete is enough — no leftover task list, no carried-over context. The signal file carries the structured outcome.
+   - **`bail`**: record the bail reason prominently at the top, then list the still-remaining tasks. Completed tasks are still dropped.
```
xbrianh: The `bail` case is underspecified relative to `next-plan`. For `next-plan` you spell out how `## Context`, `## Approach`, and `## Open questions` should be pruned; for `bail` you only say "record the bail reason prominently at the top, then list the still-remaining tasks. Completed tasks are still dropped" — leaving the agent to guess what to do with the other sections. It's likely the intent is the same pruning rules apply, just with a bail-reason banner at the top, but I'd say so explicitly rather than rely on the agent inferring it. Otherwise different runs may produce structurally different bail documents.

skills/handoff/handoff.py
```diff
@@ -141,10 +141,15 @@ def build_prompt(
    - **`next-plan`**: some tasks remain unimplemented; the next gremlin should tackle them.
    - **`bail`**: something prevents safe continuation (broken state, incoherent plan, security issue, etc.).
 
-4. Write an **updated plan document** to: `{out_path}`
-   - Free-form markdown. A human reading it must be able to see what is done and what remains.
-   - Mark completed tasks (e.g. `[x]`) and leave remaining tasks as `[ ]`.
-   - If `bail`, record the bail reason prominently at the top.
+4. Write an **updated plan document** (the "rolling plan") to: `{out_path}`
+
+   The rolling plan describes only **remaining** work. Completed tasks are removed entirely — no `[x]` markers, no struck-through entries, no "completed" appendix. The chain of versioned plan files plus git history is the audit trail; the rolling plan does not repeat it. Do not propagate the overarching goal of the chain forward into the rolling plan — that lives upstream, in the original spec.
```
xbrianh, line 146: Worth flagging the operational consequence of "that lives upstream, in the original spec": `bossgremlin.py` replaces `boss_state["current_plan"]` with each new rolling plan (line 246) and only the value of `current_plan` is forwarded as `--plan` to subsequent handoff invocations (line 208). `boss_state["spec_path"]` is recorded at init but never re-read. So under the new rule, the **overarching goal is only ever seen by the first handoff agent** — every subsequent handoff sees a goal-less remaining-task list and has to scope the next chunk from that alone. This may be perfectly fine in practice (task dependencies and ordering are usually self-evident from the remaining list), but if the goal context turns out to matter for chunking decisions on long chains, consider either (a) re-reading `boss_state["spec_path"]` and passing it as a separate read-only "north star" input to handoff, or (b) carrying a one-line goal header forward in the rolling plan as an exception to the "only remaining work" rule. No change required for this PR — just a note.

xbrianh (commented):
Prompt-only change with a small SKILL.md update. The PR description's claim that no Python orchestration changes are needed checks out: `[x]` markers appear nowhere outside the prompt body itself, and `bossgremlin.py` only feeds the rolling plan back as the next handoff's `--plan` (line 246) without parsing it. The new "remaining work only" semantics are coherent and the per-exit-state guidance is clear. Two observations below — both minor, neither a blocker.

skills/handoff/handoff.py
```diff
@@ -141,10 +141,15 @@ def build_prompt(
...
+
+   - **`next-plan`**: include only the tasks that are not yet implemented (still `[ ]`). Prune the surrounding sections (`## Context`, `## Approach`, `## Open questions`, etc.) to match: drop sections whose reason for existing was a now-completed task; keep or trim the rest so the document stays a coherent description of the remaining work.
+     - Under `## Open questions`, carry forward unresolved entries; drop entries tied to completed tasks.
+     - If a task is only partly landed, keep it (rewritten if needed to reflect what remains).
+   - **`chain-done`**: minimal output. A short note that the chain is complete is enough — no leftover task list, no carried-over context. The signal file carries the structured outcome.
```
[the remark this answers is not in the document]
xbrianh: Good catch — the PR description was overstated. Updated to clarify scope: this PR only changes the **handoff rolling plan** convention. Per-gremlin `plan.md` files still use `[x]` to mark completed tasks, and `gremlins.py`'s commit-message synthesis (the regex match on `- [x] ...` under `## Tasks`) still parses those markers — that behavior is unchanged.

skills/handoff/handoff.py
xbrianh: Agreed — fixed in 955523d. The bail bullet now says explicitly that the same pruning rules as `next-plan` apply (only remaining tasks, surrounding sections trimmed accordingly, unresolved `## Open questions` carried forward), with a bail-reason banner added at the top.

skills/handoff/handoff.py
xbrianh: Acknowledged — leaving as-is for this PR per your note. Filing the operational consequence as a follow-up consideration: if chunking quality on long chains turns out to suffer once handoff #2+ stops seeing the spec, options (a) re-reading `spec_path` as a separate read-only input or (b) carrying a one-line goal header in the rolling plan are both reasonable next steps. Not changing scope here.

PR #67, opened 2026-04-25
Anchor gh bossgremlin chains to the repo default branch

### LINKED ISSUE #63: bossgremlin gh-chain: target_branch assumes user launched from default branch
Surfaced in [PR #62 review comment](https://github.com/xbrianh/claude/pull/62#discussion_r3142365385).

`bossgremlin` captures `target_branch` at chain start via `get_current_branch(project_root)` in `main()`. For gh chains, `handoff` then inspects `origin/<target_branch>` to see what PRs have landed. But ghgremlin children create their PRs *from the default branch* (see `skills/ghgremlin/ghgremlin.sh:463` — "Create a new branch from the default branch..."), so `origin/<target_branch>` only matches "where PRs landed" when the user happens to be on the default branch when launching the chain.

If a user launches a gh chain from a feature branch:

- **Best case**: `origin/<feature>` exists but is stale and never receives PR merges. `handoff` sees no diff and emits `chain-done` prematurely.
- **Worst case**: the branch does not exist on origin at all, and `git fetch origin <feature>` hard-fails on the first handoff.

This is pre-existing — the old code passed `target_branch` straight into `--rev` and had the same blind spot — so PR #62 does not regress anything. Filing as a follow-up.

## Resolution

A gh bossgremlin should not care what branch the user happens to be on when launching the chain. It should always inspect the repo's **default branch** (typically `origin/main`), since that is where ghgremlin children land their PRs.

Implementation sketch:

- At chain start, resolve the default branch via `gh repo view --json defaultBranchRef` (once, cached in boss state).
- Use that — not the user's current branch — as `target_branch` / `chain_base_ref` for all subsequent handoff invocations in a gh chain.
- The user's current branch becomes irrelevant to a gh chain.

This makes gh chains launchable from any working state without requiring the user to first check out `main`.

xbrianh:
## Summary

For gh chains, `bossgremlin` previously captured `target_branch` from `git rev-parse --abbrev-ref HEAD` at chain start. But ghgremlin children create their PRs from the default branch, so `origin/<target_branch>` only matched "where PRs land" when the user happened to be on the default branch at launch time. From a feature branch this either silently emitted `chain-done` (if `origin/<feature>` was stale) or hard-failed at the first handoff fetch (if the feature branch wasn't on origin at all).

Fix:

- For gh chains, resolve the default branch via `gh repo view --json defaultBranchRef` at chain start (cached in `boss_state`).
- Use that as `target_branch`, and set `chain_base_ref` to the SHA of `origin/<default-branch>` (also fetched at chain start) so handoff diffs are exactly the PRs landed during the chain.
- Local chains keep the existing HEAD/current-branch anchoring.

The user's current branch is now irrelevant to a gh chain — they can launch from anywhere.

Closes #63

## Test plan

- [ ] Launch a gh boss chain from a feature branch and verify it correctly inspects `origin/main` (or the configured default) at each handoff.
- [ ] Launch a gh boss chain from `main` and verify behavior is unchanged.
- [ ] Launch a local boss chain and verify HEAD-based anchoring is unchanged.
- [ ] Verify `boss_state.json` records `target_branch` as the default branch name and `chain_base_ref` as the `origin/<default>` SHA for gh chains.

### DIFF
--- a/skills/bossgremlin/bossgremlin.py
+++ b/skills/bossgremlin/bossgremlin.py
@@ -122,6 +122,42 @@ def get_current_branch(project_root: str) -> str:
     return "" if branch == "HEAD" else branch
 
 
+def get_default_branch(project_root: str) -> str:
+    """Resolve the repo's default branch via gh CLI. Calls die() on failure."""
+    r = subprocess.run(
+        ["gh", "repo", "view", "--json", "defaultBranchRef",
+         "-q", ".defaultBranchRef.name"],
+        capture_output=True, text=True, cwd=project_root,
+    )
+    if r.returncode != 0:
+        die(f"gh repo view failed in {project_root}: {r.stderr.strip()}")
+    name = r.stdout.strip()
+    if not name:
+        die(f"gh repo view returned empty default branch in {project_root}")
+    return name
+
+
+def get_remote_branch_sha(project_root: str, branch: str) -> str:
+    """Fetch origin/<branch> and return its SHA. Calls die() on failure."""
+    try:
+        fetch = subprocess.run(
+            ["git", "fetch", "origin", branch],
+            capture_output=True, text=True, cwd=project_root,
+            timeout=HANDOFF_FETCH_TIMEOUT,
+        )
+    except subprocess.TimeoutExpired:
+        die(f"git fetch origin {branch} timed out after {HANDOFF_FETCH_TIMEOUT}s at chain start")
+    if fetch.returncode != 0:
+        die(f"git fetch origin {branch} failed at chain start: {fetch.stderr.strip()}")
+    r = subprocess.run(
+        ["git", "rev-parse", f"origin/{branch}"],
+        capture_output=True, text=True, cwd=project_root,
+    )
+    if r.returncode != 0:
+        die(f"git rev-parse origin/{branch} failed: {r.stderr.strip()}")
+    return r.stdout.strip()
+
+
 def init_boss_state(spec_path: str, chain_kind: str, chain_base_ref: str,
                     target_branch: str, state_dir: str) -> dict:
     boss_state = {
@@ -392,9 +428,17 @@ def main(argv):
     boss_state_file = os.path.join(state_dir, "boss_state.json")
     if not os.path.isfile(boss_state_file):
         log(f"chain start: kind={chain_kind}, spec={spec_path}")
-        chain_base_ref = get_head_ref(project_root)
-        target_branch = get_current_branch(project_root)
-        log(f"base ref: {chain_base_ref[:12]}, target branch: {target_branch or '(detached)'}")
+        if chain_kind == "gh":
+            # gh children open PRs from the repo's default branch and land
+            # there, regardless of where the user happens to be. Anchor the
+            # chain to origin/<default-branch> so handoff diffs land cleanly.
+            target_branch = get_default_branch(project_root)
+            chain_base_ref = get_remote_branch_sha(project_root, target_branch)
+            log(f"base ref: {chain_base_ref[:12]} (origin/{target_branch}), target branch: {target_branch}")
+        else:
+            chain_base_ref = get_head_ref(project_root)
+            target_branch = get_current_branch(project_root)
+            log(f"base ref: {chain_base_ref[:12]}, target branch: {target_branch or '(detached)'}")
         boss_state = init_boss_state(
             spec_path=spec_path,
             chain_kind=chain_kind,
--- a/skills/bossgremlin/bossgremlin.py
+++ b/skills/bossgremlin/bossgremlin.py
@@ -32,7 +32,11 @@ STATE_ROOT = os.path.join(
 
 POLL_INTERVAL = 5  # seconds between finished-marker polls
 HANDOFF_TIMEOUT = int(os.environ.get("BOSSGREMLIN_HANDOFF_TIMEOUT", "3600"))
+# Bounds every interaction with `origin` (chain-start fetch of the default
+# branch and per-handoff fetch of the target branch). Named for the historical
+# per-handoff use; kept under the same env var for backwards compatibility.
 HANDOFF_FETCH_TIMEOUT = int(os.environ.get("BOSSGREMLIN_HANDOFF_FETCH_TIMEOUT", "60"))
+GH_VIEW_TIMEOUT = 30  # seconds; bounds `gh repo view` at chain start
 
 _current_proc = None
 _stop_requested = False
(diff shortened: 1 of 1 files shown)


xbrianh (commented):

skills/bossgremlin/bossgremlin.py
```diff
@@ -122,6 +122,42 @@ def get_current_branch(project_root: str) -> str:
...
+        die(f"git fetch origin {branch} timed out after {HANDOFF_FETCH_TIMEOUT}s at chain start")
+    if fetch.returncode != 0:
+        die(f"git fetch origin {branch} failed at chain start: {fetch.stderr.strip()}")
+    r = subprocess.run(
+        ["git", "rev-parse", f"origin/{branch}"],
+        capture_output=True, text=True, cwd=project_root,
+    )
+    if r.returncode != 0:
+        die(f"git rev-parse origin/{branch} failed: {r.stderr.strip()}")
+    return r.stdout.strip()
```
xbrianh, line 179: Minor: this chain-start `git fetch origin <branch>` will be re-done immediately by `run_handoff` for the very first handoff (it always fetches `origin/<target_branch>` before invoking handoff). Two back-to-back fetches of the same ref is harmless — just noting the duplication in case you'd rather use `git ls-remote origin <branch>` here to grab the SHA without populating the local tracking ref. Equally fine to leave as-is for symmetry with the per-handoff fetch.

skills/bossgremlin/bossgremlin.py
```diff
@@ -122,6 +122,42 @@ def get_current_branch(project_root: str) -> str:
...
+    return name
+
+
+def get_remote_branch_sha(project_root: str, branch: str) -> str:
+    """Fetch origin/<branch> and return its SHA. Calls die() on failure."""
+    try:
+        fetch = subprocess.run(
+            ["git", "fetch", "origin", branch],
+            capture_output=True, text=True, cwd=project_root,
+            timeout=HANDOFF_FETCH_TIMEOUT,
```
xbrianh, line 162: Naming nit: `HANDOFF_FETCH_TIMEOUT` (and its env var `BOSSGREMLIN_HANDOFF_FETCH_TIMEOUT`) now also gates a chain-start fetch, not just per-handoff fetches. The semantics are the same ("how long are we willing to wait on `origin`"), so reuse is reasonable, but the name slightly overstates scope. Worth a one-line comment near the constant or a rename to something like `ORIGIN_FETCH_TIMEOUT` if you want the name to match. Not worth blocking on.

skills/bossgremlin/bossgremlin.py
```diff
@@ -392,9 +428,17 @@ def main(argv):
...
-        chain_base_ref = get_head_ref(project_root)
-        target_branch = get_current_branch(project_root)
-        log(f"base ref: {chain_base_ref[:12]}, target branch: {target_branch or '(detached)'}")
+        if chain_kind == "gh":
+            # gh children open PRs from the repo's default branch and land
+            # there, regardless of where the user happens to be. Anchor the
+            # chain to origin/<default-branch> so handoff diffs land cleanly.
+            target_branch = get_default_branch(project_root)
+            chain_base_ref = get_remote_branch_sha(project_root, target_branch)
+            log(f"base ref: {chain_base_ref[:12]} (origin/{target_branch}), target branch: {target_branch}")
```
xbrianh, line 449: The PR description test plan still has all four checkboxes unchecked. Before landing, worth at minimum confirming items 1 and 4 (gh chain launched from a feature branch correctly inspects `origin/main`, and `boss_state.json` records the expected `target_branch` / `chain_base_ref`) — those are the two that exercise the new code paths and aren't covered elsewhere.

xbrianh (commented):
Correct fix for the right bug: gh-chain children open their PRs from the repo's default branch, so anchoring `chain_base_ref` / `target_branch` to `origin/<default-branch>` (and persisting both in `boss_state`) is the right invariant. The branching in `main()` and the per-chain-kind dispatch already in `run_handoff` line up cleanly. Two minor, non-blocking notes inline; no blocker findings.

skills/bossgremlin/bossgremlin.py
xbrianh: Good call — I've folded both fetches (chain-start and per-handoff) into a single `fetch_origin_branch()` helper, so the duplication concern (#3142398109) is resolved. Kept the chain-start `git fetch` rather than `git ls-remote` so the local tracking ref is populated symmetrically with the per-handoff fetch.

skills/bossgremlin/bossgremlin.py
xbrianh: Added a comment near the constant noting it now also bounds the chain-start fetch. Kept the existing name and env var (`BOSSGREMLIN_HANDOFF_FETCH_TIMEOUT`) for back-compat rather than renaming.

skills/bossgremlin/bossgremlin.py
xbrianh: Acknowledged — I'll exercise items 1 and 4 manually before landing (gh chain from a feature branch, and inspect the resulting `boss_state.json`). No code change needed for this comment.

skills/bossgremlin/bossgremlin.py
```diff
@@ -122,6 +122,42 @@ def get_current_branch(project_root: str) -> str:
...
+    try:
+        fetch = subprocess.run(
+            ["git", "fetch", "origin", branch],
+            capture_output=True, text=True, cwd=project_root,
+            timeout=HANDOFF_FETCH_TIMEOUT,
+        )
+    except subprocess.TimeoutExpired:
+        die(f"git fetch origin {branch} timed out after {HANDOFF_FETCH_TIMEOUT}s at chain start")
+    if fetch.returncode != 0:
+        die(f"git fetch origin {branch} failed at chain start: {fetch.stderr.strip()}")
```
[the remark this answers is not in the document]
xbrianh: Done — extracted `fetch_origin_branch(project_root, branch, *, context=...)` that owns the bounded fetch + die-on-failure pattern. Both `get_remote_branch_sha` (chain start) and `run_handoff` (per-handoff) now go through it, with `context` interpolated into the error messages.

skills/bossgremlin/bossgremlin.py
```diff
@@ -122,6 +122,42 @@ def get_current_branch(project_root: str) -> str:
...
+        ["gh", "repo", "view", "--json", "defaultBranchRef",
+         "-q", ".defaultBranchRef.name"],
+        capture_output=True, text=True, cwd=project_root,
+    )
+    if r.returncode != 0:
+        die(f"gh repo view failed in {project_root}: {r.stderr.strip()}")
+    name = r.stdout.strip()
+    if not name:
+        die(f"gh repo view returned empty default branch in {project_root}")
+    return name
```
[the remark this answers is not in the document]
xbrianh: Done — `get_default_branch` now passes `timeout=GH_VIEW_TIMEOUT` (30s) and catches both `FileNotFoundError` (gh not installed) and `subprocess.TimeoutExpired`, dying with a clear message in each case.

skills/bossgremlin/bossgremlin.py
```diff
@@ -122,6 +122,42 @@ def get_current_branch(project_root: str) -> str:
...
+    except subprocess.TimeoutExpired:
+        die(f"git fetch origin {branch} timed out after {HANDOFF_FETCH_TIMEOUT}s at chain start")
+    if fetch.returncode != 0:
+        die(f"git fetch origin {branch} failed at chain start: {fetch.stderr.strip()}")
+    r = subprocess.run(
+        ["git", "rev-parse", f"origin/{branch}"],
+        capture_output=True, text=True, cwd=project_root,
+    )
+    if r.returncode != 0:
+        die(f"git rev-parse origin/{branch} failed: {r.stderr.strip()}")
```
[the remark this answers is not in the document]
xbrianh: Done — the new `fetch_origin_branch` helper invokes `git fetch origin refs/heads/<branch>:refs/remotes/origin/<branch>` instead of the bare branch name, and `get_remote_branch_sha` resolves the corresponding `refs/remotes/origin/<branch>`. A leading `-` in the branch name can no longer be parsed by `git fetch` as an option.

PR #68, opened 2026-04-25
Pass overarching spec as north star to every handoff

### LINKED ISSUE #66: boss chain: handoff #2+ no longer sees overarching goal
Surfaced as a follow-up consideration in [PR #65 review thread](https://github.com/xbrianh/claude/pull/65#discussion_r3142380900) (and [reply](https://github.com/xbrianh/claude/pull/65#discussion_r3142382904)). Not changing scope on PR #65; filing the operational consequence here.

## Background

After PR #65, the rolling plan that flows step-to-step in a boss chain contains only remaining work — the overarching goal of the chain does **not** propagate forward.

`bossgremlin.py` mechanics:

- `boss_state["current_plan"]` is replaced with each new rolling plan (line 246).
- Only the value of `current_plan` is forwarded as `--plan` to subsequent handoff invocations (line 208).
- `boss_state["spec_path"]` is recorded at init but never re-read.

So under the new rule, **the overarching goal is only ever seen by the first handoff agent.** Every subsequent handoff sees a goal-less remaining-task list and has to scope the next chunk from that alone.

## Why this might or might not matter

In practice, task dependencies and ordering are usually self-evident from the remaining list, so this may be fine. But if chunking quality on long chains turns out to suffer once handoff #2+ stops seeing the spec, two reasonable options:

1. **Re-read `boss_state["spec_path"]`** and pass it as a separate read-only "north star" input to every handoff invocation.
2. **Carry a one-line goal header forward** in the rolling plan as an explicit exception to the "only remaining work" rule.

## When to act

No action needed unless we observe degraded chunking on long chains. This issue is a placeholder so the option set is recorded if the symptom appears.

xbrianh:
## Summary
- After PR #65, the rolling plan only carries remaining work, so any handoff after the first lost sight of the chain's original goal.
- `bossgremlin.run_handoff()` now forwards `boss_state["spec_path"]` as a new optional `--spec` flag on every `handoff.py` invocation.
- `handoff.py` reads the spec and renders an "Overarching goal (north star)" section above the rolling input plan, framed as read-only context that must not be echoed into the updated plan. The flag stays optional so the standalone `/handoff` use case keeps working without a separate spec.

## Test plan
- [ ] Run a `bossgremlin` chain with two or more handoffs and confirm the second-and-later handoff prompts include the original spec under the north-star section.
- [ ] Run `/handoff --plan <path>` (no `--spec`) and confirm the prompt renders without the north-star section.
- [ ] Confirm the rolling plan output does not duplicate the spec content.

Closes #66

### DIFF
--- a/skills/bossgremlin/bossgremlin.py
+++ b/skills/bossgremlin/bossgremlin.py
@@ -168,6 +168,7 @@ def run_handoff(gr_id: str, state_dir: str, boss_state: dict,
     out_path = os.path.join(state_dir, f"handoff-{n:03d}.md")
     signal_path = os.path.join(state_dir, f"handoff-{n:03d}.state.json")
     current_plan = boss_state["current_plan"]
+    spec_path = boss_state.get("spec_path", "")
     base_ref = boss_state["chain_base_ref"]
     chain_kind = boss_state.get("chain_kind")
     target_branch = boss_state.get("target_branch", "")
@@ -202,10 +203,12 @@ def run_handoff(gr_id: str, state_dir: str, boss_state: dict,
     else:
         die(f"unknown chain_kind: {chain_kind!r}")
 
-    log(f"handoff {n}: plan={current_plan}, base={base_ref[:12]}, rev={rev_label}, cwd={handoff_cwd}")
+    log(f"handoff {n}: plan={current_plan}, spec={spec_path}, base={base_ref[:12]}, rev={rev_label}, cwd={handoff_cwd}")
+    spec_args = ["--spec", spec_path] if spec_path else []
     cmd = [
         handoff_script,
         "--plan", current_plan,
+        *spec_args,
         "--out", out_path,
         "--base", base_ref,
         "--model", model,
--- a/skills/bossgremlin/bossgremlin.py
+++ b/skills/bossgremlin/bossgremlin.py
@@ -168,7 +168,7 @@ def run_handoff(gr_id: str, state_dir: str, boss_state: dict,
     out_path = os.path.join(state_dir, f"handoff-{n:03d}.md")
     signal_path = os.path.join(state_dir, f"handoff-{n:03d}.state.json")
     current_plan = boss_state["current_plan"]
-    spec_path = boss_state.get("spec_path", "")
+    spec_path = boss_state["spec_path"]
     base_ref = boss_state["chain_base_ref"]
     chain_kind = boss_state.get("chain_kind")
     target_branch = boss_state.get("target_branch", "")
@@ -203,8 +203,16 @@ def run_handoff(gr_id: str, state_dir: str, boss_state: dict,
     else:
         die(f"unknown chain_kind: {chain_kind!r}")
 
-    log(f"handoff {n}: plan={current_plan}, spec={spec_path}, base={base_ref[:12]}, rev={rev_label}, cwd={handoff_cwd}")
-    spec_args = ["--spec", spec_path] if spec_path else []
+    # Only forward --spec once the rolling plan has diverged from the spec.
+    # On handoff #1, init_boss_state seeds current_plan = spec_path, so
+    # passing --spec would render the same document twice in the prompt
+    # (once as north-star, once as the input plan). Skipping --spec there
+    # avoids the duplication; the north-star section first appears on the
+    # handoff that runs against an actual rolling plan (#2+).
+    forward_spec = bool(spec_path) and spec_path != current_plan
+    spec_log = spec_path if forward_spec else "(none)"
+    log(f"handoff {n}: plan={current_plan}, spec={spec_log}, base={base_ref[:12]}, rev={rev_label}, cwd={handoff_cwd}")
+    spec_args = ["--spec", spec_path] if forward_spec else []
     cmd = [
         handoff_script,
         "--plan", current_plan,
--- a/skills/handoff/handoff.py
+++ b/skills/handoff/handoff.py
@@ -104,14 +104,27 @@ def build_prompt(
     out_path: pathlib.Path,
     child_plan_path: pathlib.Path,
     signal_path: pathlib.Path,
+    spec_text: Optional[str] = None,
 ) -> str:
     diff_body = git_diff[:50000] if git_diff else "(empty — no changes yet)"
     diff_trunc = f"\n(diff truncated to 50000 chars; {len(git_diff)} chars total)" if len(git_diff) > 50000 else ""
     log_body = git_log if git_log else "(no commits yet — branch just started)"
 
+    spec_section = ""
+    if spec_text is not None:
+        spec_section = f"""## Overarching goal (north star)
+
+This is the original chain spec. It does not change between handoffs and is read-only context for understanding what the chain as a whole is working toward. Use it to judge whether the rolling input plan below is on track and to scope the next step coherently. Do not echo it into the updated plan — it stays in `--spec`.
+
+~~~~
+{spec_text}
+~~~~
+
+"""
+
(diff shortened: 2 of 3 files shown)


xbrianh (commented):

skills/bossgremlin/bossgremlin.py
```diff
@@ -202,10 +203,12 @@ def run_handoff(gr_id: str, state_dir: str, boss_state: dict,
     else:
         die(f"unknown chain_kind: {chain_kind!r}")
 
-    log(f"handoff {n}: plan={current_plan}, base={base_ref[:12]}, rev={rev_label}, cwd={handoff_cwd}")
+    log(f"handoff {n}: plan={current_plan}, spec={spec_path}, base={base_ref[:12]}, rev={rev_label}, cwd={handoff_cwd}")
+    spec_args = ["--spec", spec_path] if spec_path else []
```
xbrianh: On handoff #1, `current_plan == spec_path` (set by `init_boss_state` at line 132), so the agent gets the same file content rendered twice — once under "Overarching goal (north star)" and once under "## Input plan" — and the prompt's instruction "Do not echo it into the updated plan — it stays in `--spec`" is in tension with the normal task of emitting remaining tasks from the input plan into the rolling plan.

The agent will probably handle it (input-plan tasks are clearly the source for the rolling plan), but it's a chunk of wasted context and a bit of a foot-gun for a misread. Two cheap fixes:

- Gate `spec_args` on `spec_path and spec_path != current_plan`, so the north-star section only appears once divergence has occurred (handoff #2+).
- Or, in `init_boss_state`, leave `current_plan` unset / equal to a sentinel until the first rolling plan is written, and have handoff.py treat "no input plan yet" as the signal to seed from spec.

First option is the smaller change.

skills/handoff/handoff.py
```diff
@@ -217,6 +232,22 @@ def main(argv: List[str]) -> int:
     except OSError as exc:
         die(f"failed to read --plan {plan_path}: {exc}")
 
+    spec_text: Optional[str] = None
+    if args.spec is not None:
+        spec_path = pathlib.Path(args.spec).resolve()
+        if not spec_path.exists():
```
xbrianh, line 246: If the spec file is moved, renamed, or deleted between chain start and handoff time, this `die()` halts the entire boss chain via handoff exit code 1, even though the rolling plan and landed diff alone are still enough to make a coherent next-step decision (which is exactly what the standalone `/handoff` invocation does). Worth deciding whether that's the right tradeoff:

- Hard fail (current): treats the spec as a load-bearing chain anchor; user gets a clear error and resolves it. Costs a chain halt for what may be an incidental file move.
- Graceful fallback: log a warning, render without the north-star section, continue. Matches standalone `/handoff` semantics and keeps the chain alive at the cost of losing north-star context partway through.

No strong opinion either way, but the silent halt-the-whole-chain consequence isn't obvious from the code and deserves a deliberate call.

skills/bossgremlin/bossgremlin.py
```diff
@@ -168,6 +168,7 @@ def run_handoff(gr_id: str, state_dir: str, boss_state: dict,
     out_path = os.path.join(state_dir, f"handoff-{n:03d}.md")
     signal_path = os.path.join(state_dir, f"handoff-{n:03d}.state.json")
     current_plan = boss_state["current_plan"]
+    spec_path = boss_state.get("spec_path", "")
```
xbrianh: Minor: `init_boss_state` always sets `spec_path` (line 128), so this `.get("spec_path", "")` default only triggers on a `boss_state.json` written before this PR landed. If you intend to support resuming pre-PR chains, fine; if not, `boss_state["spec_path"]` is clearer about the invariant. Either way, the empty-string log output `spec=, base=...` at line 206 looks a bit off — consider `spec_path or '(none)'` for the log line.

xbrianh (commented):
Threads the chain spec cleanly through bossgremlin → handoff.py and surfaces it to the agent prompt; the code is straightforward and matches the existing `--plan` validation pattern. Two concerns worth a closer look: (1) on handoff #1, `current_plan == spec_path`, so the agent receives the same document twice with mildly conflicting framing about what to keep in the rolling plan; (2) a moved or deleted spec file mid-chain hard-fails handoff.py and halts the whole boss chain — graceful degradation may be preferable. One minor note about a defensive default that init_boss_state already guarantees.

skills/bossgremlin/bossgremlin.py
xbrianh: Fixed in 4155556. Took the smaller option: gated `spec_args` in `run_handoff` on `spec_path and spec_path != current_plan`, so the north-star section is suppressed on handoff #1 (where `current_plan == spec_path`) and only appears once divergence has occurred. The log line shows `spec=(none)` rather than the path when `--spec` is suppressed, so it's visible from the run log which handoff this kicks in on.

skills/handoff/handoff.py
xbrianh: Fixed in 4155556. Went with graceful fallback: a missing, non-file, empty, non-UTF-8, or unreadable `--spec` now logs a warning to stderr and renders the prompt without the north-star section instead of `die()`-ing. This matches standalone `/handoff` semantics (which renders without north-star when no `--spec` is passed at all) and keeps the boss chain alive on incidental file moves. The rolling plan + landed diff are still the load-bearing inputs for the decision; the spec is best-effort context.

skills/bossgremlin/bossgremlin.py
xbrianh: Fixed in 4155556. Switched to `boss_state["spec_path"]` to make the invariant explicit (no need to support pre-PR `boss_state.json` files). Log line now uses `spec_log = spec_path if forward_spec else "(none)"` so it reads `spec=(none)` when `--spec` is suppressed instead of `spec=`.

skills/bossgremlin/bossgremlin.py
```diff
@@ -202,10 +203,12 @@ def run_handoff(gr_id: str, state_dir: str, boss_state: dict,
     else:
         die(f"unknown chain_kind: {chain_kind!r}")
 
-    log(f"handoff {n}: plan={current_plan}, base={base_ref[:12]}, rev={rev_label}, cwd={handoff_cwd}")
+    log(f"handoff {n}: plan={current_plan}, spec={spec_path}, base={base_ref[:12]}, rev={rev_label}, cwd={handoff_cwd}")
+    spec_args = ["--spec", spec_path] if spec_path else []
     cmd = [
         handoff_script,
         "--plan", current_plan,
+        *spec_args,
```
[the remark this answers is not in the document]
xbrianh: Fixed in 4155556 — same change addresses xbrianh's parallel comment above. `bossgremlin.run_handoff` now only forwards `--spec` when `spec_path and spec_path != current_plan`, suppressing the duplication on handoff #1.

skills/handoff/handoff.py
```diff
@@ -104,14 +104,27 @@ def build_prompt(
...
 
+    spec_section = ""
+    if spec_text is not None:
+        spec_section = f"""## Overarching goal (north star)
+
+This is the original chain spec. It does not change between handoffs and is read-only context for understanding what the chain as a whole is working toward. Use it to judge whether the rolling input plan below is on track and to scope the next step coherently. Do not echo it into the updated plan — it stays in `--spec`.
+
+~~~~
+{spec_text}
+~~~~
```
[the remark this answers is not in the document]
xbrianh: Fixed in 4155556 — applied the suggested truncation. `build_prompt` now caps `spec_body` at 50000 chars with a `(spec truncated to 50000 chars; N chars total)` trailer, matching the existing `git_diff` handling.

PR #70, opened 2026-04-25
ghaddress: file issues for out-of-scope review comments

### LINKED ISSUE #69: /ghaddress: file issues for out-of-scope review comments instead of dropping them
## Context

Today, when `/ghaddress` decides a review comment is "out of scope" (OOS), the comment is replied to on the PR thread but no durable record is created elsewhere. If the OOS comment was actually flagging a real defect, the work is silently lost — especially painful in autonomous gremlin runs where no human is around to remember "we deferred that."

We want `/ghaddress` to file a GitHub issue for OOS comments that look like real defects, while still dismissing pure noise (style nits, factually-wrong comments, reviewer-marked nits) without creating issue-tracker churn.

Reference: design doc at `/tmp/design-ghaddress-file-oos-issues.md` (kept locally; not in repo).

## Approach

This is a behavior-only change to the `/ghaddress` skill prompt — `skills/ghaddress/SKILL.md`. No new scripts, no new wrappers. `/ghgremlin` invokes `/ghaddress` as its address stage, so updating the skill prompt covers both standalone use and gremlin use without touching `ghgremlin.sh`.

The change adds a second decision step inside the OOS branch of the existing per-comment loop: after marking a comment OOS, the agent classifies it as "real defect" vs "noise," and conditionally calls `gh issue create` before posting the PR thread reply. The reply text differs based on which branch was taken (`Filed as #N` vs a short dismissal reason).

We chose this over a separate post-pass / standalone "OOS triage" skill because:
- The agent already has the comment in context when it decides OOS — re-loading it later is wasteful and error-prone.
- One control flow keeps the PR-thread reply tightly coupled to the issue-creation outcome (so we never reply `Filed as #N` when the create failed, or vice versa).
- No new scaffolding: the skill prompt is the only surface that changes.

Failure handling for `gh issue create` follows the design doc: log prominently, do not fail the address stage. We'll instruct the agent to fall back to a dismissal-style reply that names the intended issue title, so the comment thread isn't silently dropped.

## Tasks

- [ ] Update `skills/ghaddress/SKILL.md` Process section to split the OOS branch into two decisions: (1) is this OOS? (existing) and (2) if OOS, is it a real defect or noise?
- [ ] Document the "real defect vs noise" filter inline in the skill prompt, mirroring the design doc's bullet lists (file when: bug/regression/security/correctness/perf/invariant violation; don't file when: reviewer wrong, style nit, reviewer-marked non-blocking, already addressed elsewhere). Include the "ambiguous → file it" tie-breaker.
- [ ] Specify the `gh issue create` invocation: title = short distillation of the comment (no special prefix); body quotes or summarizes the comment, links back to the originating PR (`Ref #<pr>` and a permalink to the comment), and stands on its own without requiring readers to chase the PR.
- [ ] Specify the PR-thread reply text in both branches: `Filed as #N` (with link) when an issue was filed; brief dismissal reason when noise/wrong.
- [ ] Specify failure handling for `gh issue create`: log prominently in the run output, fall back to a dismissal reply that names what the issue *would* have said, do not fail the address stage. Do not write a bail marker for this — issue-creation failure is not a reason to stop addressing the rest of the PR.
- [ ] Verify the updated skill prompt still works under `/ghgremlin`'s address stage (it invokes `/ghaddress` directly, so no `ghgremlin.sh` change should be required — confirm by reading the address-stage call site).
- [ ] Run `scripts/sync.sh push` to mirror the SKILL.md change to `~/.claude/skills/ghaddress/SKILL.md`.

## Open questions

None blocking. Issue title style is settled (short distillation, no prefix). Cross-PR de-duplication is explicitly out of scope per the design doc.

xbrianh:
## Summary
- Splits the `/ghaddress` per-comment loop so OOS comments go through a second decision: real defect vs noise.
- Real defects (or ambiguous cases, by tie-breaker) get `gh issue create` first, then the PR thread reply becomes `Filed as #N`. Pure noise gets a brief dismissal reply, no issue.
- Issue-creation failures log prominently and fall back to a dismissal reply that names the intended issue title -- they do not bail the address stage. Behavior change is in `skills/ghaddress/SKILL.md` only; `/ghgremlin` invokes `/ghaddress` directly so no wrapper change is needed.

## Test plan
- [ ] Run `/ghaddress` on a PR with a mix of comments that should be (a) addressed, (b) filed as issues, (c) dismissed as noise; confirm one issue per OOS-defect comment with a self-contained body, `Ref #<pr>`, and a permalink, and confirm the matching `Filed as #N` reply lands on the originating thread.
- [ ] Run `/ghaddress` on a PR with only style/nit OOS comments; confirm no issues are filed and the dismissal replies state a brief reason.
- [ ] Simulate a `gh issue create` failure (e.g., revoke issue scope on the token); confirm the address stage finishes, the failure is logged loudly, and the comment thread gets a dismissal reply naming the intended issue title rather than being silently dropped.
- [ ] Run an end-to-end `/ghgremlin` whose review surfaces at least one OOS-defect comment; confirm the gremlin completes its address stage and the issue + reply pair land on the PR.

Closes #69

### DIFF
--- a/skills/ghaddress/SKILL.md
+++ b/skills/ghaddress/SKILL.md
@@ -30,7 +30,7 @@ $ARGUMENTS[1:]
 ## Process
 
 1. Read all review comments above carefully.
-2. For each actionable comment:
+2. For each comment, decide whether it is in-scope to address in this PR or out-of-scope (OOS). For each in-scope comment:
    a. Understand what the reviewer is asking for.
    b. Read the relevant code to understand the current state.
    c. Make the fix or change requested.
@@ -38,11 +38,74 @@ $ARGUMENTS[1:]
    a. `git add` the changed files (by name, not `-A`).
    b. `git commit` with a message summarizing what was addressed.
    c. `git push` to the PR branch.
-4. Only after the push succeeds, reply to each comment thread:
-   - For review comments, use `gh api repos/{owner}/{repo}/pulls/<number>/comments/{comment_id}/replies -f body="<reply>"`.
-   - For questions or acknowledgements (no code change needed), reply briefly.
-   - Skip comments that have already been resolved.
-5. Summarize what was done.
+4. Only after the push succeeds, reply to each comment thread. Skip comments that have already been resolved.
+   - For in-scope comments you addressed: reply briefly acknowledging the fix.
+   - For questions or acknowledgements that need no code change: reply briefly.
+   - For OOS comments: run the OOS triage in the next section before replying.
+   - Post replies to review comments with `gh api repos/{owner}/{repo}/pulls/<number>/comments/{comment_id}/replies -f body="<reply>"`.
+5. Summarize what was done, including any issues filed for OOS comments and any `gh issue create` failures.
+
+## Out-of-scope triage (file issues for real defects)
+
+For each comment you marked OOS, decide whether it looks like a real defect or noise. The goal is that real defects survive the OOS decision as filed issues, while pure noise doesn't pollute the issue tracker.
+
+**File a new issue when the OOS comment flags any of:**
+
+- A bug or regression
+- A security or data-correctness concern
+- A performance pathology
+- A hidden invariant being violated
+- Anything else that, if true, would warrant a fix in some future PR
+
+**Don't file an issue when:**
+
+- The reviewer's claim is wrong — they misread the code, missed context, or are factually incorrect.
+- The comment is a subjective style nit, naming bikeshed, or "consider extracting…" preference with no defect claim.
+- The reviewer themselves marked the comment as non-blocking, nit, or fyi.
+- The comment is already addressed elsewhere (e.g., another comment in the same review covers the same point).
+
+**Tie-breaker:** If you're genuinely unsure whether the comment is a real defect or the reviewer is wrong, **file the issue**. Over-filing is cheap; losing a real bug is not. A human triaging the issue can close it as invalid.
+
+### Filing the issue
+
+Use `gh issue create` with:
+
+- **Title**: a short distillation of the comment (no special prefix). Aim for a sentence fragment that names the defect, not the comment ID.
+- **Body**: must stand on its own — a reader should not need to chase the PR to understand the issue. Include:
+  - A short summary of the defect in your own words.
+  - The reviewer's comment, quoted or summarized, so the original framing is preserved.
+  - `Ref #<pr-number>` so GitHub cross-links the PR.
+  - A permalink to the originating review comment (the comment's `html_url` from the API response).
+
+Example invocation (pass body via heredoc to preserve formatting):
+
+```
+gh issue create --title "<short distillation>" --body "$(cat <<'EOF'
+<summary>
+
+Reviewer comment:
+> <quoted or summarized comment>
+
+Ref #<pr-number>
+<permalink to review comment>
+EOF
+)"
+```
+
+Capture the issue number/URL from the command's output for the reply.
+
+### Replying on the PR thread
+
+- **Issue filed**: reply `Filed as #N` (or `Filed as <issue-url>`) so the reviewer and any later reader can find it.
+- **No issue (noise / reviewer wrong / already addressed)**: reply with a brief dismissal reason — enough that a human skimming the PR understands why no further action was taken.
+
+### When `gh issue create` fails
+
+Issue-creation failure is not a reason to stop addressing the rest of the PR. Do **not** write a bail marker.
+
+- Log the failure prominently in the run output (a clear `ERROR: failed to file issue for comment <id>: <error>` line, and include it in the final summary).
+- Fall back to a dismissal-style reply on the PR thread that names what the issue *would* have said (intended title and a one-line summary), so the comment is not silently dropped.
+- Continue with the remaining comments.
 
 ## Bail markers (only when running under a gremlin)
 
--- a/skills/ghaddress/SKILL.md
+++ b/skills/ghaddress/SKILL.md
@@ -119,7 +119,7 @@ Use the helper:
   ~/.claude/skills/_bg/set-bail.sh "$GR_ID" secrets "<one-line reason>"
   ```
 
-- For any other reason you decline to proceed (ambiguous reviewer ask, conflicting comments, scope outside this PR, etc.):
+- For any other reason you decline to proceed (ambiguous reviewer ask, conflicting comments, etc.):
 
   ```
   ~/.claude/skills/_bg/set-bail.sh "$GR_ID" other "<one-line reason>"
--- a/skills/ghaddress/SKILL.md
+++ b/skills/ghaddress/SKILL.md
@@ -34,11 +34,15 @@ $ARGUMENTS[1:]
    a. Understand what the reviewer is asking for.
    b. Read the relevant code to understand the current state.
    c. Make the fix or change requested.
-3. After making all code changes, stage, commit, and push:
+
+   OOS comments are deferred to step 4 and the *Out-of-scope triage* section below — do not act on them in this step.
+3. After making all in-scope code changes, stage, commit, and push:
    a. `git add` the changed files (by name, not `-A`).
    b. `git commit` with a message summarizing what was addressed.
    c. `git push` to the PR branch.
-4. Only after the push succeeds, reply to each comment thread. Skip comments that have already been resolved.
+
+   If there were no in-scope changes (every comment was OOS or already-resolved), skip 3a–3c and proceed directly to step 4 — `git commit` would fail with "nothing to commit" and gate out the OOS triage, which is the case where the new behavior matters most.
+4. Reply to each comment thread (after the push succeeds, if step 3 ran). Skip comments that have already been resolved.
    - For in-scope comments you addressed: reply briefly acknowledging the fix.
    - For questions or acknowledgements that need no code change: reply briefly.
    - For OOS comments: run the OOS triage in the next section before replying.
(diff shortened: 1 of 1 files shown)


xbrianh (commented):

skills/ghaddress/SKILL.md
```diff
@@ -30,19 +30,82 @@ $ARGUMENTS[1:]
...
+Capture the issue number/URL from the command's output for the reply.
+
+### Replying on the PR thread
+
+- **Issue filed**: reply `Filed as #N` (or `Filed as <issue-url>`) so the reviewer and any later reader can find it.
+- **No issue (noise / reviewer wrong / already addressed)**: reply with a brief dismissal reason — enough that a human skimming the PR understands why no further action was taken.
+
+### When `gh issue create` fails
+
+Issue-creation failure is not a reason to stop addressing the rest of the PR. Do **not** write a bail marker.
```
xbrianh, line 109: **This new line contradicts the existing Bail markers section below.** The pre-existing bail-markers section (line 122 in the new file, just past this hunk) still lists `scope outside this PR` as an example reason to write an `other` bail marker. With this PR, OOS comments are explicitly *not* a bail reason — they go through the triage above (file issue or post a dismissal reply). A reader who follows the bail section literally — or a gremlin run where the bail marker short-circuits the rest of the run — could end up bailing on the very comments this section says to triage.

Suggest updating that list in the same PR: drop `scope outside this PR` from the bail-markers examples, and add a one-liner like "Note: out-of-scope comments are not a bail reason — see *Out-of-scope triage* above."

skills/ghaddress/SKILL.md
```diff
@@ -30,19 +30,82 @@ $ARGUMENTS[1:]
...
+
+- Log the failure prominently in the run output (a clear `ERROR: failed to file issue for comment <id>: <error>` line, and include it in the final summary).
+- Fall back to a dismissal-style reply on the PR thread that names what the issue *would* have said (intended title and a one-line summary), so the comment is not silently dropped.
```
xbrianh: **Fallback reply is indistinguishable from a noise dismissal.** Per line 100, a noise/wrong-reviewer/already-addressed dismissal also "reply with a brief dismissal reason." The failure fallback here produces the same shape — "names what the issue *would* have said" — so a reviewer reading the PR thread can't tell whether (a) the agent classified the comment as noise or (b) the agent classified it as a real defect, tried to file an issue, and the create failed. The latter is a much louder signal for a human triager.

Worth saying explicitly that the reply should mention filing was attempted and failed, e.g. "Intended to file as `<title>` but `gh issue create` failed: <error>; please file manually if this is a real defect." The loud `ERROR:` log line covers the run output, but the PR thread is the durable surface a human reviewer will see.

xbrianh (commented):
Docs-only change to `skills/ghaddress/SKILL.md` that splits the OOS branch into a real-defect-vs-noise triage and adds `gh issue create` + `Filed as #N` reply semantics. Logic and failure handling read cleanly and match issue #69. Two notable gaps: the pre-existing **Bail markers** section still lists "scope outside this PR" as a bail reason, which now contradicts the new OOS-triage path; and the `gh issue create` failure fallback writes a dismissal-style reply that's indistinguishable from a normal noise dismissal, so reviewers can't tell that a filing attempt failed.

skills/ghaddress/SKILL.md
```diff
@@ -30,19 +30,82 @@ $ARGUMENTS[1:]
 ## Process
 
 1. Read all review comments above carefully.
-2. For each actionable comment:
+2. For each comment, decide whether it is in-scope to address in this PR or out-of-scope (OOS). For each in-scope comment:
```
xbrianh, line 33: Minor: the loop body in step 2 only describes work for in-scope comments ("For each in-scope comment: a/b/c"). OOS comments are classified here but not acted on until step 4. That's fine in practice, but a reader skimming the Process list could miss that OOS comments deliberately have no step-2 action. Consider adding a one-liner like "OOS comments are deferred to step 4 and the *Out-of-scope triage* section" to make the deferral explicit.

skills/ghaddress/SKILL.md
xbrianh: The failure-fallback reply is shaped exactly like an intentional 'this is noise' dismissal — both reduce to 'reply with a brief reason' on the PR thread. A human triaging the PR later can't tell whether (a) the comment was deliberately dismissed or (b) `gh issue create` failed and the issue still needs to be filed by hand.

Suggest the fallback mark itself, e.g. open with `⚠️ Tried to file as "<intended title>" but `gh issue create` failed: <error>. Please file manually if this is a real defect.` That preserves the audit trail and gives a future triager a concrete next action without breaking the no-bail rule. The run-log `ERROR:` line on bullet 1 covers the operator, but a reader looking only at the PR thread has no way to find the dropped issue.

skills/ghaddress/SKILL.md
```diff
@@ -30,19 +30,82 @@ $ARGUMENTS[1:]
...
+- **Body**: must stand on its own — a reader should not need to chase the PR to understand the issue. Include:
+  - A short summary of the defect in your own words.
+  - The reviewer's comment, quoted or summarized, so the original framing is preserved.
+  - `Ref #<pr-number>` so GitHub cross-links the PR.
+  - A permalink to the originating review comment (the comment's `html_url` from the API response).
+
+Example invocation (pass body via heredoc to preserve formatting):
+
+```
+gh issue create --title "<short distillation>" --body "$(cat <<'EOF'
```
xbrianh: Two cross-repo gotchas worth pinning down in the example:

1. `gh issue create` without `--repo` files in the cwd repo. Under autonomous gremlins this usually matches the PR's repo, but the skill's frontmatter is `context: fork`, and a fork checkout configured with a downstream `origin` will silently file the issue in the wrong tracker. Suggest deriving owner/repo from the same `gh pr view` data already fetched (or `gh pr view --json baseRepository`) and passing `--repo <owner>/<repo>` explicitly in the example.

2. `Ref #<pr-number>` only autolinks within the same repo. If the agent ever files cross-repo (intentionally or by mistake), use `Ref <owner>/<repo>#<pr-number>`.

Both small, but the failure mode for #1 is silent — the issue is created, just in the wrong place — so it's hard to detect after the fact.

skills/ghaddress/SKILL.md
```diff
@@ -30,19 +30,82 @@ $ARGUMENTS[1:]
...
 3. After making all code changes, stage, commit, and push:
    a. `git add` the changed files (by name, not `-A`).
    b. `git commit` with a message summarizing what was addressed.
    c. `git push` to the PR branch.
-4. Only after the push succeeds, reply to each comment thread:
-   - For review comments, use `gh api repos/{owner}/{repo}/pulls/<number>/comments/{comment_id}/replies -f body="<reply>"`.
-   - For questions or acknowledgements (no code change needed), reply briefly.
-   - Skip comments that have already been resolved.
-5. Summarize what was done.
+4. Only after the push succeeds, reply to each comment thread. Skip comments that have already been resolved.
```
xbrianh: Latent edge case introduced by the in-scope/OOS split: if every reviewer comment is OOS, step 2 produces no code changes, step 3b's `git commit` fails with 'nothing to commit', and this 'only after the push succeeds' gate prevents the OOS triage from running at all — exactly the scenario where the new issue-filing behavior is most valuable.

Suggest a one-line escape hatch in step 3, e.g. 'If there were no in-scope changes, skip 3a–3c and proceed to step 4', or condition the commit on `git diff --cached --quiet` returning non-zero. Without it, an all-OOS review would silently drop every defect that should have been filed.

xbrianh (commented):
Solid direction — the OOS triage rules are well-articulated, the bias toward over-filing is the right tie-breaker, and not bailing on `gh issue create` failures matches how the rest of the gremlin pipeline handles non-essential side effects. Three things worth tightening before merge:

1. The dismissal-style fallback when `gh issue create` fails leaves no audit trail distinguishing it from an intentional 'this is noise' dismissal.
2. The example `gh issue create` invocation has no `--repo`, which can silently mis-file in fork/multi-remote setups (recall the skill's frontmatter is `context: fork`).
3. Step 3 unconditionally commits and pushes, but an all-OOS PR produces no code changes — `git commit` will fail and gate step 4's OOS triage out entirely, exactly the case where the new behavior matters most.

Inline comments below.

skills/ghaddress/SKILL.md
xbrianh: Done — added a 'Note: out-of-scope comments are not a bail reason — see *Out-of-scope triage* above' line at the bottom of the bail-markers list, and called out that gh issue create failures aren't a bail reason either. The 'scope outside this PR' phrase was already dropped in b8c023c.

skills/ghaddress/SKILL.md
xbrianh: Done — fallback reply now opens with `Tried to file as "<intended title>" but \`gh issue create\` failed: <error>. Please file manually if this is a real defect.` and explicitly says not to phrase it as a generic dismissal. The PR thread now carries the audit trail, not just the run log.

skills/ghaddress/SKILL.md
xbrianh: Done — added a one-liner at the end of step 2: 'OOS comments are deferred to step 4 and the *Out-of-scope triage* section below — do not act on them in this step.'

skills/ghaddress/SKILL.md
```diff
@@ -30,19 +30,82 @@ $ARGUMENTS[1:]
...
+   - For in-scope comments you addressed: reply briefly acknowledging the fix.
+   - For questions or acknowledgements that need no code change: reply briefly.
+   - For OOS comments: run the OOS triage in the next section before replying.
+   - Post replies to review comments with `gh api repos/{owner}/{repo}/pulls/<number>/comments/{comment_id}/replies -f body="<reply>"`.
```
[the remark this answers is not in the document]
xbrianh: The documented path is actually correct. Per [GitHub's REST docs](https://docs.github.com/en/rest/pulls/comments#create-a-reply-for-a-review-comment), the endpoint is `POST /repos/{owner}/{repo}/pulls/{pull_number}/comments/{comment_id}/replies` — it does take the pull number. (`gh api repos/{owner}/{repo}/pulls/comments/{comment_id}/replies` would 404.) Leaving as-is.

skills/ghaddress/SKILL.md
```diff
@@ -30,19 +30,82 @@ $ARGUMENTS[1:]
...
+### Filing the issue
+
+Use `gh issue create` with:
+
+- **Title**: a short distillation of the comment (no special prefix). Aim for a sentence fragment that names the defect, not the comment ID.
+- **Body**: must stand on its own — a reader should not need to chase the PR to understand the issue. Include:
+  - A short summary of the defect in your own words.
+  - The reviewer's comment, quoted or summarized, so the original framing is preserved.
+  - `Ref #<pr-number>` so GitHub cross-links the PR.
+  - A permalink to the originating review comment (the comment's `html_url` from the API response).
```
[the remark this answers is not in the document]
xbrianh: Done — added an explicit secret-redaction step before quoting reviewer comments: scan for credentials/keys/tokens/internal URLs/customer data, redact with `[redacted]`, or if redaction isn't safe/clean, bail with `set-bail.sh "$GR_ID" secrets ...` instead of filing. Issues are public and permanent, so erring on the side of bailing.

skills/ghaddress/SKILL.md
xbrianh: Done — fallback reply now opens with `Tried to file as "<intended title>" but \`gh issue create\` failed: <error>. Please file manually if this is a real defect.` Self-marking, distinguishable from a noise dismissal, gives a future triager a concrete next action without breaking the no-bail rule.

skills/ghaddress/SKILL.md
xbrianh: Done on both:
1. Example now uses `gh issue create --repo <owner>/<repo> ...`, with a note to derive owner/repo from the same `gh pr view` data (`--json baseRepository`) and a callout that the misfile-into-fork failure mode is silent.
2. Documented `Ref <owner>/<repo>#<pr-number>` for the cross-repo case alongside the same-repo `Ref #<n>`.

skills/ghaddress/SKILL.md
xbrianh: Done — step 3 now has an escape hatch: 'If there were no in-scope changes (every comment was OOS or already-resolved), skip 3a–3c and proceed directly to step 4 — `git commit` would fail with "nothing to commit" and gate out the OOS triage, which is the case where the new behavior matters most.' Also softened step 4's gate to 'after the push succeeds, if step 3 ran'.

PR #83, opened 2026-04-26
Phase 0: pipeline/ design doc

### LINKED ISSUE #81: Phase 0: Inventory and design doc for `pipeline/` package layout
# Phase 0: Inventory and pipeline/ design doc

## Context

Produce a design-doc-only PR that documents the proposed `pipeline/` package layout, defines the `ClaudeClient` interface, and specifies the updated `launch.sh --resume` contract. No code moves in this phase — only `pipeline/DESIGN.md` is created. This doc becomes the shared contract that Phases 1, 3, and 4 implement against.

## Approach

Read the existing code in `skills/localgremlin/localgremlin.py`, `skills/localgremlin/_core.py`, and `skills/ghgremlin/ghgremlin.sh` to inventory the shared concepts before writing anything. Then write `pipeline/DESIGN.md` (creating the `pipeline/` directory for it). Do not touch any other file.

Shared concepts to inventory and document:
- Stage runner (sequencing, resume, signal handling, child-process tracking from `_core._reap_all`)
- `claude -p` wrapper + stream-json logger
- State-dir / artifact layout
- `set_stage` / `emit_bail` bookkeeping
- Plan/spec loading and snapshotting
- Resume-precondition checks
- Handoff invocation contract
- Branch/worktree management
- "Land" semantics

Proposed module layout to document in `pipeline/DESIGN.md`:
- `pipeline/cli.py` — `python -m pipeline.cli {local,gh,boss}` dispatch
- `pipeline/clients/claude.py` — `ClaudeClient` protocol + real subprocess implementation
- `pipeline/clients/fake.py` — recording test double
- `pipeline/runner.py` — generic stage runner
- `pipeline/state.py` — session-dir resolution, `set_stage` / `emit_bail` wrappers
- `pipeline/git.py` — `in_git_repo`, `git_head`, dirty-tree checks, branch/worktree helpers
- `pipeline/stages/{plan,implement,review_code,address_code,commit_pr,ghreview,ghaddress,wait_copilot}.py`
- `pipeline/orchestrators/{local,gh,boss}.py`
- `pipeline/prompts/` — stage prompts currently embedded in scripts
- `pipeline/pyproject.toml` — project metadata + pytest config
- `pipeline/tests/`

`ClaudeClient` interface to document: minimum surface based on current `_core.run_claude` and `ghgremlin.sh` usage: `run(model, prompt, *, label, raw_path, output_format="stream-json", flags=...) -> None` raising on non-zero exit. Real implementation owns child-process tracking and `_reap_all` on signal; fake records calls and replays canned stream-json from a fixture directory.

`launch.sh --resume` contract to document: how `_bg/launch.sh` will resolve `python -m pipeline.cli {local,gh,boss}` instead of `skills/<name>/<name>.{py,sh}` once each orchestrator migrates.

## Tasks

- [ ] Task 1: Read `skills/localgremlin/localgremlin.py`, `skills/localgremlin/_core.py`, and `skills/ghgremlin/ghgremlin.sh` to inventory shared concepts across the three gremlins.
- [ ] Task 2: Create directory `pipeline/` at the repo root.
- [ ] Task 3: Write `pipeline/DESIGN.md` documenting the proposed module layout (as listed above), the `ClaudeClient` interface (protocol + real + fake), and the updated `launch.sh --resume` contract.
- [ ] Task 4: Verify no other file in the repo was changed (no `skills/` changes, no `scripts/sync.sh` change, no `CLAUDE.md` change).

## Open questions

- `pipeline/` at repo root vs. under `scripts/`. Default: repo root.

xbrianh:
## Summary
- Add `pipeline/DESIGN.md` documenting the proposed shared-pipeline package layout, the `ClaudeClient` interface (protocol + real subprocess + fake recording double), and the updated `launch.sh --resume` contract.
- Inventories shared concepts across `localgremlin`, `ghgremlin`, and `bossgremlin` (stage runner, `claude -p` wrapper + stream-json logger, state-dir layout, `set_stage`/`emit_bail`, plan/spec loading, resume preconditions, handoff, branch/worktree, land semantics) with file:line citations into the existing scripts.
- No code moves: `pipeline/DESIGN.md` is the only new file. It becomes the shared contract for Phases 1, 3, and 4 to implement against.

## Test plan
- [ ] Confirm only `pipeline/DESIGN.md` is added (no `skills/` / `scripts/sync.sh` / `CLAUDE.md` changes).
- [ ] Review module layout and decide on the three open questions at the bottom of the doc (`pipeline/` location, single vs. self-contained `pyproject.toml`, parsing helpers on client vs. free functions).
- [ ] Confirm the `launch.sh --resume` migration plan (prefer `python -m pipeline.cli <kind>`, fall back to existing paths).

Closes #81

### DIFF
--- a/pipeline/DESIGN.md
+++ b/pipeline/DESIGN.md
@@ -1,0 +1,490 @@
+# `pipeline/` design doc
+
+Status: design-only (Phase 0). No code under `pipeline/` other than this file
+exists yet. This document is the shared contract for Phases 1, 3, and 4 to
+implement against.
+
+## Why this exists
+
+Today the three gremlin orchestrators (`localgremlin`, `ghgremlin`,
+`bossgremlin`) live as separate scripts under `skills/<kind>/`. Two are
+Python (`localgremlin.py` + `_core.py`, plus the foreground `localreview` /
+`localaddress` CLIs), and one is bash (`ghgremlin.sh`). They share a
+non-trivial amount of conceptual surface area — stage sequencing, resume,
+`claude -p` invocation, stream-json parsing, state-dir bookkeeping, plan
+loading, child-process tracking, signal handling — but the shared bits are
+either duplicated across the two languages or leak through the launcher
+(`skills/_bg/launch.sh`).
+
+The `pipeline/` package consolidates the shared mechanism into one Python
+package with a single CLI entry point (`python -m pipeline.cli {local,gh,boss}`),
+behind a swappable `ClaudeClient` interface so the orchestrators are testable
+without spawning real `claude -p` subprocesses.
+
+## Inventory of shared concepts
+
+The concepts below are present (in some form) in all three gremlins today
+and need to be reified by the new package:
+
+### Stage runner
+
+- Each gremlin defines an ordered list of stages and runs them
+  sequentially. Today this is:
+  - localgremlin: `plan → implement → review-code → address-code` (4 stages,
+    `VALID_RESUME_STAGES` in `localgremlin.py:82`).
+  - ghgremlin: `plan → implement → commit-pr → request-copilot → ghreview →
+    wait-copilot → ghaddress` (7 stages, `STAGES` in `ghgremlin.sh:98`).
+  - bossgremlin: chains child gremlins serially, deciding next-step via
+    `/handoff` between each.
+- `--resume-from <stage>` skips any stage whose index in the list is below
+  the target. Each gremlin re-validates its own preconditions before
+  declaring the resume safe (see "Resume-precondition checks" below).
+- Stage transitions emit a `set_stage` call so `/gremlins` and the
+  session-summary hook can see progress. `review-code` additionally emits
+  a sub-stage dict (`{holistic, detail, scope}` flipping running→done as
+  each reviewer thread finishes). See `_core.py:417`.
+- Child-process tracking: long-running stages (notably `review-code`'s
+  triple-reviewer fan-out) spawn multiple `claude -p` subprocesses in
+  parallel. On SIGINT/SIGTERM we must terminate every live child. Today
+  that's `_core._reap_all` (`_core.py:68`) backed by a module-level
+  `_children: List[Popen]` under an RLock; bash uses `trap 'kill -- -$$'`.
+
+### `claude -p` wrapper + stream-json logger
+
+- All three gremlins spawn `claude -p <prompt>` with
+  `--permission-mode bypassPermissions --output-format stream-json
+  --verbose`, optionally `--model <model>`, and parse the resulting
+  newline-delimited JSON.
+- Python: `_core.run_claude` (`_core.py:263`) tees stdout to a `.jsonl` raw
+  trace under the session dir and emits a human-readable progress trace to
+  stderr via `log_stream` / `_emit_event` (`_core.py:179`).
+- Bash: `progress_tee` in `ghgremlin.sh:19` does the same with `jq`.
+- ghgremlin additionally extracts `session_id` from the stream-json `init`
+  event (`extract_session_id`, `ghgremlin.sh:358`) so the `commit-pr` stage
+  can `--resume <session-id>` the same agent session that did the
+  implementation. It also extracts URLs from gh-tool tool_results
+  (`extract_gh_url`, `ghgremlin.sh:305`) — preferring the most recent
+  `gh issue create` / `gh pr create` call's tool_result over a final-text
+  scrape, which is fragile.
+- Trim semantics: progress lines are truncated to 200 chars; malformed
+  JSON lines from a crashing `claude -p` are skipped silently (the bash
+  side does this with `jq … || true`; the Python side does it with a
+  try/except around `json.loads` in `log_stream`).
+
+### State dir + artifact layout
+
+- Real gremlins (under `_bg/launch.sh`) get
+  `${XDG_STATE_HOME:-$HOME/.local/state}/claude-gremlins/<GR_ID>/` with:
+  - `state.json` — the canonical bookkeeping doc (id, kind, workdir,
+    branch, status, stage, sub_stage, pid, exit_code, bail_class,
+    pipeline_args, parent_id, etc.).
+  - `artifacts/` — plan.md, review-code-*.md, stream-*.jsonl traces,
+    spec.md (boss only), ghplan-out.jsonl (gh only).
+  - `instructions.txt` — full untruncated instructions sidecar (state.json
+    only stores a 200-char display summary).
+  - `log` — appended stdout/stderr from the gremlin process itself.
+  - `pid` — captured PID for liveness / signal delivery.
+  - `finished` / `summarized` — terminal markers cleared on resume.
+- Direct invocations (no `GR_ID`) drop everything under
+  `$STATE_ROOT/direct/<ts>-<rand>/artifacts/` so they're visually separated
+  and can be pruned on a simple age-based heuristic
+  (`resolve_session_dir`, `_core.py:452`).
+
+### `set_stage` / `emit_bail` bookkeeping
+
+- `set_stage(stage, sub_stage=None)`: shells out to
+  `~/.claude/skills/_bg/set-stage.sh` to atomically patch
+  `state.json.stage` (and `.sub_stage`). No-op without `GR_ID` or when the
+  helper is missing/non-executable. Never raises — bookkeeping must not
+  break the running gremlin.
+- `emit_bail(bail_class, bail_detail)`: shells out to
+  `~/.claude/skills/_bg/set-bail.sh`. Recorded so `/gremlins rescue
+  --headless` can decide whether automated recovery is allowed. Bail
+  classes seen in the wild: `secrets`, `security`, `reviewer_requested_changes`
+  (refused outright), and `other` (the catch-all that allows rescue).
+- bash `check_bail` (`ghgremlin.sh:157`) reads `bail_class` after each
+  stage and dies the pipeline if non-empty.
+
+### Plan / spec loading and snapshotting
+
+- `localgremlin --plan <path>`: source path is copied into
+  `session_dir/plan.md` on fresh launch; on resume the snapshot is
+  authoritative (rescue-determinism — the source may have been edited or
+  deleted in the meantime). See `localgremlin.py:159`.
+- `ghgremlin --plan <path|issue-ref>`: a file path is posted as a new
+  GitHub issue (auto-titled by a one-shot `claude -p` call) and the
+  contents copied to `artifacts/plan.md`; an issue ref
+  (`42` / `#42` / `owner/repo#42` / a full `https://github.com/.../issues/N`
+  URL) is fetched via `gh issue view`. Cross-repo issue refs deliberately
+  skip the `Closes #N` link to avoid auto-closing an issue in another
+  repo. See `ghgremlin.sh:193`.
+- `bossgremlin`: `--plan <spec-path>` is required; the spec is copied to
+  `state-dir/artifacts/spec.md` by launch.sh up front
+  (`launch.sh:472`) and is the north star handed to every child via
+  `/handoff`.
+
+### Resume-precondition checks
+
+Each orchestrator validates the in-tree state before agreeing to resume at
+a non-zero stage:
+
+- localgremlin (`localgremlin.py:191`):
+  - `implement`+: `plan.md` exists and is non-empty.
+  - `review-code`+: in git, either dirty tree or any commit reachable from
+    HEAD; outside git, the worktree contains at least one non-metadata
+    file.
+  - `address-code`: all three review-code-*.md exist and are non-empty.
+- ghgremlin (`ghgremlin.sh`):
+  - `commit-pr` is structurally unresumable (it needs `IMPL_SESSION` from
+    the same in-process run), so a request to resume there silently
+    rewinds to `implement` (`ghgremlin.sh:107`).
+  - Stages from `implement` onward without `--plan` need `issue_url` in
+    state.json; stages from `wait-copilot`+ need `pr_url`.
+- bossgremlin: resume strategy is "rescue the failing child gremlin in
+  place, then the boss continues" rather than fast-forwarding the boss
+  itself.
+
+### Handoff invocation contract
+
+bossgremlin calls `~/.claude/skills/handoff/handoff.py --plan <path>
+[--out <path>] [--base <ref>] [--model <model>] [--timeout <secs>]`
+between children. The handoff agent reads the current plan and the diff
+landed on the branch and writes either:
+- a `next-plan` document (keep going, here's the next child plan), or
+- a `chain-done` marker (we're done), or
+- a `bail` marker (something is wrong, stop).
+
+### Branch / worktree management
+
+- localgremlin in a git repo: `launch.sh` creates a named branch
+  `bg/localgremlin/<GR_ID>` so commits remain reachable after `finish.sh`.
+- ghgremlin: detached worktree, since stage 2b creates and pushes its own
+  `issue-N-<slug>` branch.
+- Outside a git repo: a copy of the project root via `cp -a`.
+- Resume reuses the existing workdir + branch from state.json
+  (`launch.sh:124`).
+
+### "Land" semantics
+
+`/gremlins land <id>` lands a finished gremlin onto its target branch:
+- localgremlin: squash-merge the `bg/localgremlin/<GR_ID>` branch into the
+  parent and clean up the worktree.
+- ghgremlin: merge the PR via `gh pr merge` and clean up.
+
+This is `/gremlins`-side logic, not pipeline-side, but `pipeline/` will
+expose enough state (branch name, PR URL, exit code) for `/gremlins` to
+keep doing its job unchanged.
+
+## Proposed module layout
+
+All paths relative to the repo root.
+
+```
+pipeline/
+├── DESIGN.md                       # this file
+├── pyproject.toml                  # project metadata + pytest config
+├── cli.py                          # `python -m pipeline.cli {local,gh,boss}` dispatch
+├── runner.py                       # generic stage runner
+├── state.py                        # session-dir resolution, set_stage / emit_bail wrappers
+├── git.py                          # in_git_repo, git_head, dirty-tree checks, branch/worktree helpers
+├── clients/
+│   ├── __init__.py
+│   ├── claude.py                   # ClaudeClient protocol + real subprocess implementation
+│   └── fake.py                     # recording test double
+├── stages/
+│   ├── __init__.py
+│   ├── plan.py                     # local + gh plan resolution (file-source, issue-ref, /ghplan)
+│   ├── implement.py
+│   ├── review_code.py              # triple-lens parallel reviewer fan-out
+│   ├── address_code.py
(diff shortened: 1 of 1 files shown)


xbrianh (commented):

pipeline/DESIGN.md
```diff
@@ -0,0 +1,490 @@
...
+  declaring the resume safe (see "Resume-precondition checks" below).
+- Stage transitions emit a `set_stage` call so `/gremlins` and the
+  session-summary hook can see progress. `review-code` additionally emits
+  a sub-stage dict (`{holistic, detail, scope}` flipping running→done as
+  each reviewer thread finishes). See `_core.py:417`.
+- Child-process tracking: long-running stages (notably `review-code`'s
+  triple-reviewer fan-out) spawn multiple `claude -p` subprocesses in
+  parallel. On SIGINT/SIGTERM we must terminate every live child. Today
+  that's `_core._reap_all` (`_core.py:68`) backed by a module-level
+  `_children: List[Popen]` under an RLock; bash uses `trap 'kill -- -$$'`.
```
xbrianh: The parenthetical `bash uses \`trap 'kill -- -$$'\`` doesn't match current `ghgremlin.sh` — it has no `trap` at all (`grep -n trap skills/ghgremlin/ghgremlin.sh` → no matches). The `_core.py:47` comment is referring to the *historical* bash predecessor, not today's ghgremlin.

This matters for the parity claim a few sections down ("Implementations MUST track child processes and reap them on SIGINT/SIGTERM"): ghgremlin's bash today does no reaping, so the unified pipeline *gains* that behavior for ghgremlin rather than preserving it. Suggest restating: "Python tracks via `_core._reap_all`; bash today does no reaping (children are cleaned up only when the launcher's process tree is killed) — the unified pipeline brings ghgremlin to parity with localgremlin."

pipeline/DESIGN.md
```diff
@@ -0,0 +1,490 @@
...
+  - `review-code`+: in git, either dirty tree or any commit reachable from
+    HEAD; outside git, the worktree contains at least one non-metadata
+    file.
+  - `address-code`: all three review-code-*.md exist and are non-empty.
+- ghgremlin (`ghgremlin.sh`):
+  - `commit-pr` is structurally unresumable (it needs `IMPL_SESSION` from
+    the same in-process run), so a request to resume there silently
+    rewinds to `implement` (`ghgremlin.sh:107`).
+  - Stages from `implement` onward without `--plan` need `issue_url` in
+    state.json; stages from `wait-copilot`+ need `pr_url`.
```
xbrianh: Off-by-one on the resume-precondition stage. The PR_URL resolution block in `ghgremlin.sh:480-487` runs unconditionally before the `request-copilot` stage and dies if `.pr_url` is missing on resume — so `request-copilot` (and every stage after it) needs `pr_url`, not just `wait-copilot`+. The accurate statement is: "stages from `request-copilot` onward need `pr_url`."

xbrianh (commented):
Solid Phase-0 doc — the inventory matches the code well, the module split reads sensibly, and the migration plan correctly threads through `launch.sh`'s pipeline-resolver. Most of my comments are precision fixes on behavioral claims (a couple of which point at code that isn't quite what's written) and a few design questions worth resolving before Phase 1 starts coding against the contract: the test seam scoped to `claude -p` only, the `ClaudeClient.run` signature inconsistencies, and one of the open questions that the doc itself already argues toward an answer for.

pipeline/DESIGN.md
```diff
@@ -0,0 +1,490 @@
...
+- bossgremlin: resume strategy is "rescue the failing child gremlin in
+  place, then the boss continues" rather than fast-forwarding the boss
+  itself.
+
+### Handoff invocation contract
+
+bossgremlin calls `~/.claude/skills/handoff/handoff.py --plan <path>
+[--out <path>] [--base <ref>] [--model <model>] [--timeout <secs>]`
```
xbrianh: The handoff invocation contract is missing two flags that bossgremlin actually forwards today (`bossgremlin.py:264-273`, `handoff.py:201-212`):

- `--spec <path>` — the immutable north-star spec, distinct from the rolling `--plan`. The doc's `bossgremlin` plan-loading section already names this distinction ("the spec is copied to `state-dir/artifacts/spec.md` … and is the north star handed to every child"), so the handoff contract should mirror it.
- `--rev <ref>` — the rev to compare the diff against (e.g. `origin/<target_branch>` for gh chains so handoff sees PRs that landed remotely).

Both are load-bearing for chain semantics; worth listing here so Phase 4 isn't surprised.

pipeline/DESIGN.md
```diff
@@ -0,0 +1,490 @@
...
+│   ├── gh.py                       # ghgremlin pipeline
+│   └── boss.py                     # bossgremlin pipeline (chains children + handoff)
+├── prompts/                        # stage prompts currently embedded in scripts
+│   ├── __init__.py
+│   ├── local_plan.md
+│   ├── implement.md
+│   ├── address_code.md
+│   ├── gh_pr.md
+│   ├── gh_pr_no_issue.md
+│   └── scope_review.md
```
xbrianh: The `pipeline/prompts/` list omits the three review-lens files that already live as standalone `.md` at `skills/localgremlin/lens-{holistic,detail,scope}-code.md` and are read by both `_core.py` (localgremlin's review-code) and `ghgremlin.sh:504` (the parallel scope reviewer reaches across to `lens-scope-code.md`). The migration needs to bring those along — either co-located in `pipeline/prompts/lens-*.md`, or kept where they are with a documented load path. Worth listing explicitly so the file inventory stays honest.

pipeline/DESIGN.md
```diff
@@ -0,0 +1,490 @@
...
+- `has_commits() -> bool`
+- `worktree_has_files(skip: Iterable[Path]) -> bool` (the non-git
+  fallback used by localgremlin's resume-precondition check)
+
+#### `pipeline/stages/`
+
+One module per stage. Each module exports a single `run(ctx) -> Result`
+function plus the prompt template (loaded from `pipeline/prompts/`).
+Stages depend only on `ctx.client: ClaudeClient`, never on
+`subprocess.Popen` directly — this is what lets fake clients drive them.
```
xbrianh: "Stages depend only on `ctx.client: ClaudeClient`, never on `subprocess.Popen` directly" undersells what stages actually shell out to. `gh issue view`, `gh issue create`, `gh pr create`, `gh pr edit`, `gh pr review`, `gh pr diff`, `gh pr merge`, plus `git worktree`, `git rev-parse`, dirty-tree checks, etc. Faking only `ClaudeClient` lets unit tests make `claude -p` deterministic, but stages will still spawn real `gh`/`git` against whatever happens to be on PATH and in cwd.

Worth deciding now: do `gh`/`git` calls go through their own `pipeline.gh` / `pipeline.git` modules with a fake variant for tests, or are stages tested only at the integration layer? The current text implies full unit-testability, which only holds if there's a seam for the gh/git side too.

pipeline/DESIGN.md
```diff
@@ -0,0 +1,490 @@
...
+        output_format: str = "stream-json",
+        flags: Sequence[str] = (),
+        resume_session: str | None = None,
+    ) -> None:
+        """Spawn `claude -p`, stream stdout to raw_path (also tee to stderr
+        as a human-readable progress trace), and raise on non-zero exit.
+
+        Implementations MUST track child processes and reap them on
+        SIGINT/SIGTERM (parity with _core._reap_all)."""
+        ...
```
xbrianh: Two type/default inconsistencies between the protocol and the real implementation it'll wrap:

1. `model: str` (mandatory) vs. real-impl behavior "`--model <model>` is appended only when `model` is truthy." Either the empty string is a meaningful sentinel (then type should be `str | None` or `str = ""` with a doc note) or the empty case is bug-prone and `model` should always be required. The protocol shouldn't lie about which one.

2. `flags: Sequence[str] = ()` in the protocol vs. real-impl "defaults `flags` to `(--permission-mode bypassPermissions --output-format stream-json --verbose)`." An empty default means "no flags," but the real client unconditionally injects standard flags — so callers passing `flags=()` would get the *opposite* of what the signature implies. Either move the standard flags into the protocol's default, or rename the parameter (e.g. `extra_flags: Sequence[str] = ()`) so the standard set is owned implicitly by the implementation.

Resolving these in the design doc, not the implementation, keeps the contract honest for fakes.

pipeline/DESIGN.md
```diff
@@ -0,0 +1,490 @@
...
+- Installs `SIGINT` / `SIGTERM` handlers on first construction (idempotent
+  guard so multiple clients in one process don't stomp the handler chain).
+- Streams stdout via the same `log_stream` logic as today (8 KiB buffered
+  reader so `readline()` doesn't degrade to one `os.read()` per byte —
+  this is a measured perf win on long implement-stage traces, see
+  `_core.py:271`).
+- Defaults `flags` to `("--permission-mode", "bypassPermissions",
+  "--output-format", "stream-json", "--verbose")`. `output_format=` is
+  separate (and default `"stream-json"`) so future callers can switch to
+  `"text"` for one-shot non-streaming calls (e.g. ghgremlin's issue-title
```
xbrianh: `output_format` separated from `flags` is reasonable, but the interaction with the standard flag set is underspecified. The default flags include `--output-format stream-json --verbose`. If a caller passes `output_format="text"`, what happens?

- Override `--output-format` in the default set and keep `--verbose`?
- Strip both `--output-format` and `--verbose` (since `--verbose` is paired with stream-json)?
- Build flags fresh from `output_format` and skip the default set entirely?

Three plausible behaviors, all defensible. Pick one and spell it out — otherwise Phase 1 will pick by accident and Phase 3 will have to migrate ghgremlin's text-mode title-generation against whatever Phase 1 chose.

pipeline/DESIGN.md
```diff
@@ -0,0 +1,490 @@
...
+- resume-precondition checks (which delegate to `git.py` helpers),
+- stage selection (which stages to include, in what order),
+- "land" handoff (boss only).
+
+#### `pipeline/prompts/`
+
+Prompts that today are embedded as Python triple-quoted strings or bash
+heredocs become standalone files loaded by stage modules at import or
+call time. Embedded substitutions (`{plan_text}`, `{core_principles}`,
+etc.) use `str.format` with named keys.
```
xbrianh, line 365: The `core_principles` substitution key isn't authored alongside the other prompts in `pipeline/prompts/` — it's parsed live from `agents/pragmatic-developer.md` (a different top-level repo dir, see `localgremlin.py:171-186`). This is a cross-package read that the design doesn't mention; the bullet list of substitutions makes it look like a sibling prompt file.

Worth a sentence: either `pipeline/` will read `agents/pragmatic-developer.md` directly via a documented path resolver (and the implement stage parses out `## Core Principles` at runtime, as today), or the snippet gets snapshotted into a `pipeline/prompts/core_principles.md` and we accept the duplication. The first is closer to current behavior; either way the design should call it out so the package boundary is clear.

pipeline/DESIGN.md
```diff
@@ -0,0 +1,490 @@
...
+- Whether to ship a single `pyproject.toml` at the repo root (covering
+  both the existing scripts and `pipeline/`) or keep `pipeline/` as its
+  own self-contained project. Default: self-contained, since the existing
+  scripts are not a Python package.
+- Whether `ClaudeClient.extract_session_id` and `.extract_url` belong on
+  the client or as free functions in a `pipeline/streamjson.py` helper.
+  They're parsing-only, no I/O, so making them free functions is purer —
+  but keeping them on the client matches "everything claude-shaped goes
+  through one interface" and gives the fake one place to stub. **Default:
+  client methods, as drafted above.**
```
xbrianh: I'd flip this default. The doc itself notes (1) parsing is pure, (2) the fake "reuses real-client parsing logic" because there's nothing to fake, and (3) free functions are "purer." The only argument given for keeping them on `ClaudeClient` is the aesthetic "everything claude-shaped goes through one interface" — but adding a Protocol method whose only implementation is a thin delegation to a free function adds a layer with no mechanical payoff.

`pipeline/streamjson.py` with module-level `extract_session_id(raw: Path) -> str` and `extract_url(raw: Path, url_pattern, cmd_pattern, label) -> str` is callable from anywhere (real client, fake client, ad-hoc test, future debug tool) without instantiating a client. The `ClaudeClient` interface stays minimal — just `run()` — which makes the fake trivially small.

Not a blocker for this doc, but worth resolving here so Phase 1 doesn't have to revisit it.

pipeline/DESIGN.md
```diff
@@ -0,0 +1,490 @@
...
+
+### State dir + artifact layout
+
+- Real gremlins (under `_bg/launch.sh`) get
+  `${XDG_STATE_HOME:-$HOME/.local/state}/claude-gremlins/<GR_ID>/` with:
+  - `state.json` — the canonical bookkeeping doc (id, kind, workdir,
+    branch, status, stage, sub_stage, pid, exit_code, bail_class,
+    pipeline_args, parent_id, etc.).
+  - `artifacts/` — plan.md, review-code-*.md, stream-*.jsonl traces,
+    spec.md (boss only), ghplan-out.jsonl (gh only).
```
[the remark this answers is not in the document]
xbrianh: Reworded — the artifacts list now reads `spec.md (snapshot of argv[1] when the first positional argument is a readable file — so /design-style spec hand-offs land here for any kind, not just boss)` to match `launch.sh`'s actual precondition.

pipeline/DESIGN.md
```diff
@@ -0,0 +1,490 @@
...
+  GitHub issue (auto-titled by a one-shot `claude -p` call) and the
+  contents copied to `artifacts/plan.md`; an issue ref
+  (`42` / `#42` / `owner/repo#42` / a full `https://github.com/.../issues/N`
+  URL) is fetched via `gh issue view`. Cross-repo issue refs deliberately
+  skip the `Closes #N` link to avoid auto-closing an issue in another
+  repo. See `ghgremlin.sh:193`.
+- `bossgremlin`: `--plan <spec-path>` is required; the spec is copied to
+  `state-dir/artifacts/spec.md` by launch.sh up front
+  (`launch.sh:472`) and is the north star handed to every child via
+  `/handoff`.
```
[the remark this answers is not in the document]
xbrianh: Fixed. The bossgremlin bullet now states the spec is *not* copied to `artifacts/spec.md` today (since boss passes it via `--plan`, not positionally), and flags any future parity copy as a follow-up phase.

pipeline/DESIGN.md
```diff
@@ -0,0 +1,490 @@
...
+call time. Embedded substitutions (`{plan_text}`, `{core_principles}`,
+etc.) use `str.format` with named keys.
+
+#### `pipeline/pyproject.toml`
+
+```toml
+[project]
+name = "pipeline"
+version = "0.0.0"
+requires-python = ">=3.11"
```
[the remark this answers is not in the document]
xbrianh: Lowered to `requires-python = ">=3.10"` to match the existing tooling baseline (e.g. `gremlins.py`'s `fromisoformat('Z')` shim for <3.11). Added a note that the floor moves in lockstep if we ever drop the shim.

pipeline/DESIGN.md
```diff
@@ -0,0 +1,490 @@
...
+`launch.sh:280` for fresh launch). After Phase 1 lands `pipeline/`, the
+mapping changes:
+
+### New resolution order
+
+1. If `$HOME/.claude/pipeline/` exists and `python -m pipeline.cli` is
+   invocable, prefer:
+   ```bash
+   PIPELINE=("python" "-m" "pipeline.cli" "$KIND_SHORT")
+   ```
```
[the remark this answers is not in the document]
xbrianh: Spelled out the import path: the new `PIPELINE` array is `("env" "PYTHONPATH=$HOME/.claude" "python" "-m" "pipeline.cli" "$KIND_SHORT")`, with a note that this matches the repos "checked-in scripts under ~/.claude/" convention rather than introducing a packaging step. Kept `python` (not the version-suffixed shebang form) since the project enforces `python` for tool calls.

pipeline/DESIGN.md
xbrianh: Restated. The bullet now reads "Python tracks via `_core._reap_all`; bash today does no reaping (no `trap` in `ghgremlin.sh` — children are cleaned up only when the launcher's process tree is killed). The unified pipeline brings ghgremlin to parity with localgremlin on this point."

pipeline/DESIGN.md
xbrianh: Fixed. The doc now reads "stages from `request-copilot` onward need `pr_url`" with a citation to the unconditional resolution block at `ghgremlin.sh:480-487` that dies on missing `.pr_url`.

pipeline/DESIGN.md
xbrianh: Added both flags. The contract now lists `--spec <path>` (with the `bossgremlin.py:255-260` carve-out for handoff #1, where the rolling plan and spec are the same file) and `--rev <ref>` (so handoff sees PRs that landed remotely on gh chains).

pipeline/DESIGN.md
xbrianh: Added `lens-{holistic,detail,scope}-code.md` under `pipeline/prompts/` with a comment that they're read by both localgremlin's review-code and ghgremlin's parallel scope reviewer at `ghgremlin.sh:504` today. Migration co-locates them in `pipeline/prompts/`.

pipeline/DESIGN.md
xbrianh: Resolved in the doc: added `pipeline/gh.py` to the module layout alongside the already-listed `pipeline/git.py`, and made the testability claim explicit in the `pipeline/stages/` section — Phase 1 introduces both seams with fake variants so unit tests can drive the gh/git side too, not just `claude -p`.

pipeline/DESIGN.md
xbrianh: Fixed both: (1) `model: str | None` to match the truthiness gate the real implementation uses; (2) renamed `flags` -> `extra_flags` so an empty default no longer implies "no standard set". The protocol docstring now says explicitly that `extra_flags` is appended to the standard set the implementation always injects.

pipeline/DESIGN.md
xbrianh: Picked one and spelled it out in the real-implementation section: `output_format` substitutes into the standard `--output-format` slot, `--verbose` is kept regardless, and the `text` mode is reserved for one-shot non-streaming callers (e.g. ghgremlin's title generation). `--verbose` stays because the launcher's progress tee depends on it.

pipeline/DESIGN.md
xbrianh: Called out explicitly in the `pipeline/prompts/` section: `{core_principles}` is *not* a sibling prompt file — `pipeline/stages/implement.py` reads `agents/pragmatic-developer.md` via a documented path resolver and slices the section live (matching today's `localgremlin.py:171-186`). The doc now flags the cross-package read so the package boundary is honest.

pipeline/DESIGN.md
xbrianh: Flipped. `extract_session_id` and `extract_url` are now free functions in `pipeline/streamjson.py` (new module added to the layout), and the open question at the bottom is marked resolved. `ClaudeClient` shrinks to just `run()`, and the fake doesn't need to fake parsing.

PR #91, opened 2026-04-26
Fix launch.sh exiting 1 on successful user-invoked gremlin launch

### LINKED ISSUE #90: launch.sh: successful new-launch path returns exit code 1
## Context

Every successful, user-invoked gremlin launch returns exit code 1 even though the gremlin is running fine. Sample output:

```
Preparing worktree (detached HEAD 6ce24b1)
gremlin id:  implement-tolerant-head-advance-adapter-41264f
workdir:     /var/folders/.../aibg-ghgremlin.XXXXXX.fjDGCpvkFL
log:         /Users/.../log
state file:  /Users/.../state.json
pid:         57132

The ghgremlin gremlin is running in the background. You'll be notified in a future
Claude session for this project when it finishes.
```

Exit code: **1**. The gremlin's `state.json` shows `status: running` and the PID is alive, so the launch genuinely succeeded — but Claude Code (and any other caller doing `if launch.sh ...; then`) sees a failure. We've been ignoring the "Exit code 1" line on every launch, which trains everyone to ignore launcher exit codes — that's how a real launch failure will get missed.

## Root cause

`skills/_bg/launch.sh` ends with:

```bash
} >&$( [[ $PRINT_ID -eq 1 ]] && echo 2 || echo 1 )
[[ $PRINT_ID -eq 1 ]] && printf '%s\n' "$GR_ID"
```

That last line is the culprit. When `PRINT_ID == 0` (the normal user-invoked path — `--print-id` is only used by `bossgremlin.py` spawning child gremlins), the `[[ ]]` test returns exit status 1, the `&&` short-circuits, and `printf` never runs. There's no statement after it, so the failed test's exit status becomes the script's exit code.

Bossgremlin's child-launch path doesn't see this because it sets `--print-id`, the test passes, `printf` runs and returns 0. So this only affects user-invoked launches and is invisible inside chains.

## Approach

Trivial fix. Three reasonable shapes; pick whichever reads best:

```bash
# Option A — append unconditional success
[[ $PRINT_ID -eq 1 ]] && printf '%s\n' "$GR_ID"
exit 0

# Option B — neutralize with `|| true`
[[ $PRINT_ID -eq 1 ]] && printf '%s\n' "$GR_ID" || true

# Option C — rewrite as `if`
if [[ $PRINT_ID -eq 1 ]]; then
    printf '%s\n' "$GR_ID"
fi
```

Lean toward **A** — the most legible at the very end of a script, and it documents the intent ("we're done, succeed"). B is shortest but the `|| true` looks like a bug-hiding hack to readers unfamiliar with the gotcha.

## Tasks

- [ ] Patch `skills/_bg/launch.sh`'s final line per Option A (or chosen alternative).
- [ ] Run `scripts/sync.sh push` so `~/.claude/skills/_bg/launch.sh` picks up the fix.
- [ ] Manual smoke: launch a gremlin (any kind) without `--print-id` and confirm exit code 0. Then launch one *with* `--print-id` and confirm the GR_ID still prints to stdout and exit code is 0.

## Risks

- **set -e / set -u not in play here** — I checked, the script doesn't use either, so the bug is purely about the final exit status, not error-handling control flow. The fix is local.
- **Resume path.** The `--resume` branch (lines 76-294) ends with its own `exit 0` at line 294, so the resume path is unaffected. The bug is strictly on the new-launch path.

xbrianh:
## Summary
- Append `exit 0` to `skills/_bg/launch.sh` so the script returns 0 on the user-invoked path (`PRINT_ID=0`), where the trailing `[[ ]] && printf` test otherwise leaked exit status 1 into the script's exit code.
- The `--print-id` path used by `bossgremlin.py` was already exit-0 (the test passes, `printf` runs); this fix only affects the user-invoked path.

## Test plan
- [ ] Launch a gremlin without `--print-id`, confirm exit code 0 and that the gremlin actually runs.
- [ ] Launch a gremlin with `--print-id`, confirm `GR_ID` still prints to stdout and exit code is 0.
- [ ] Confirm `--resume` path is untouched (it has its own `exit 0` at line 294).

Closes #90

### DIFF
--- a/skills/_bg/launch.sh
+++ b/skills/_bg/launch.sh
@@ -584,3 +584,4 @@ Claude session for this project when it finishes.
 EOF
 } >&$( [[ $PRINT_ID -eq 1 ]] && echo 2 || echo 1 )
 [[ $PRINT_ID -eq 1 ]] && printf '%s\n' "$GR_ID"
+exit 0


xbrianh (commented):
Clean, minimal fix. Appending `exit 0` correctly resolves the leaked exit status from the trailing `[[ ... ]] && printf` short-circuit on the user-invoked path (`PRINT_ID=0`), where the test evaluates false and bash propagates exit code 1. The `--resume` branch already uses the same `exit 0` pattern at line 261, so this brings the two paths into symmetry. Since `set -euo pipefail` is in effect, any earlier failure would have aborted before line 587, so masking a real error here is not a concern.

No blockers; nothing to address inline.

xbrianh (commented):
Minimal, focused fix. The added `exit 0` mirrors the existing pattern in the `--resume` path (line 261), where the same `[[ $PRINT_ID -eq 1 ]] && printf` compound is followed by an explicit `exit 0` to prevent the failing test from leaking exit status 1 under `set -e`. With `set -euo pipefail` enabled, the test isn't itself fatal (it's part of a `&&` list, so `set -e` doesn't trigger), but it does become the script's last-command exit status — which is what was getting returned to the caller. The fix is correct and the scope is appropriate.

PR #92, opened 2026-04-26
ghgremlin: tolerate implement-stage commits via hand-off branch

### LINKED ISSUE #87: ghgremlin implement stage: replace HEAD-advance hard assert with tolerant adapter
## Context

`ghgremlin.sh:447-449` has a hard assertion that the implement-stage agent must leave changes uncommitted:

```bash
if [[ "$(git rev-parse HEAD)" != "$PRE_HEAD" ]]; then
    die "implementation step advanced HEAD (expected uncommitted changes only); refusing to continue"
fi
```

The prompt at line 440 politely asks the model "Do not commit or push yet." This is a request, not an enforcement, and it fails any time the implementing agent decides to checkpoint with commits — which capable models do routinely on multi-step plans.

Observed: `phase-1-scaffold-pipeline-package-and-3453cd` (a child of the active boss chain `extract-pipeline-package-6-phase-chain-61c255`) bailed at exactly this guard. The plan had 17 implementation tasks; the agent committed partway through, which is the natural and correct behavior for a coding agent working a large, multi-file change.

The guard exists because the *next* stage (`commit-pr`, line 455 onward) is responsible for branching, committing, pushing, and opening the PR. If implement already committed on `main`, commit-pr's branch creation gets confused.

But the failure mode is asymmetric: a model that *should* have committed and didn't is fine (commit-pr handles staging), but a model that committed when it shouldn't have is a hard kill. That asymmetry is wrong — both states are recoverable.

## Approach

Replace the hard assertion with a tolerant adapter. After the implement claude returns:

1. If HEAD advanced AND the worktree is on the chain's start ref (e.g. `main`), reset the implementing agent's commits onto a fresh branch and let `commit-pr` push that branch as-is. (No squash here — `commit-pr`'s normal behavior is to commit-then-push, but if commits already exist, it pushes whatever's there.)
2. If HEAD didn't advance (the current happy path), proceed unchanged.
3. The "must produce changes" check at line 450-452 stays — empty PRs are still a hard error.

The cleanest framing: implement's contract becomes "leave the working state in a form that contains the work, whether committed or staged." `commit-pr` then has two cases:
- Worktree dirty, HEAD unchanged → branch + commit + push (today's behavior).
- Worktree clean, HEAD advanced → branch + push (new branch from new HEAD; original HEAD reset to chain-start ref).

`localgremlin`'s implement stage has the same shape (`localgremlin/_core.py` runs an analogous prompt). Audit it for the same brittleness once gh is fixed; the failure mode there is less visible because there's no commit-pr stage to confuse, but the principle is the same — don't impose contract details on the agent that the pipeline can absorb.

## Tasks

- [ ] Replace `ghgremlin.sh:447-449` with the tolerant adapter described above. Preserve the "no changes produced" error at line 450-452 unchanged.
- [ ] Update `commit-pr` (`ghgremlin.sh:455-490` ish) to handle the "HEAD already advanced" branch: if commits exist above PRE_HEAD, branch them off, push, open PR. Don't double-commit.
- [ ] Audit the implement prompt at line 440. Remove the "Do not commit or push yet" sentence — it's no longer a contract, just a hint that's now misleading. Replace with neutral guidance about the stage's role.
- [ ] Add a regression test (or document a manual repro) that runs implement against a plan that includes a `git commit` step and verifies the pipeline still proceeds.

## Risks

- **commit-pr session reuse.** Line 470 does `claude -p ... --resume "$IMPL_SESSION"` so commit-pr has the implement session's context. If implement already committed, the resumed session may try to commit again. Either explicitly tell the resumed session "the work is already committed, just branch and push," or detect HEAD-advanced state and skip the resume in that branch.
- **Resume semantics.** `--resume-from commit-pr` is currently a known-broken path that rewinds to implement (line 108). If we make commit-pr tolerant of pre-committed state, that rewind might no longer be necessary. Out of scope here, but worth flagging in a follow-up.

## Open questions

- Should the implement prompt actively *recommend* committing (since commit-pr can absorb either shape)? Probably yes — committing makes the agent's checkpoints durable across model timeouts. Decide as part of the prompt rewrite.

Ref the dead Phase 1 child `phase-1-scaffold-pipeline-package-and-3453cd` (issue #86, the auto-created planning issue) as the concrete repro.

xbrianh:
## Summary

Replaces the implement stage's hard `die` on HEAD-advanced with a tolerant adapter that moves any commits the implementing agent made onto a transient hand-off branch, then resets the original ref back to PRE_HEAD. commit-pr now branches on whether the work is already committed (rename + push) or staged (create + commit + push).

## Why

The previous guard killed the pipeline any time a capable implementing agent committed mid-run — which is the natural pattern on multi-step plans (observed: dead phase-1 child `phase-1-scaffold-pipeline-package-and-3453cd`). The asymmetry was wrong: a model that *should* have committed and didn't was fine, but a model that committed when it shouldn't have was a hard kill. Both are recoverable.

## Changes

- `skills/ghgremlin/ghgremlin.sh` (implement block): capture `PRE_BRANCH` (empty on detached HEAD); replace `die` with adapter that creates `HANDOFF_BRANCH` via `git switch -c` and resets `PRE_BRANCH` back to `PRE_HEAD` when applicable. Empty-state guard preserved.
- `skills/ghgremlin/ghgremlin.sh` (commit-pr block): two prompt shapes — "rename + push" when implement already committed, original "create + commit + push" otherwise. Shared branch/closes clauses.
- `skills/ghgremlin/ghgremlin.sh` (implement prompt): removed "Do not commit or push yet"; replaced with neutral guidance recommending checkpoints.

`localgremlin/_core.py` was audited per the plan and found already tolerant (line 327 dies only when both no-commits AND clean worktree); no change needed there.

## Test plan

- [ ] Manual repro documented inline near the adapter: in a worktree, run an implement prompt that includes a `git commit` step; verify HEAD-advanced produces a hand-off branch + clean original-ref reset rather than a `die`.
- [ ] Verify commit-pr's HEAD-advanced prompt successfully renames + pushes + opens a PR with a `Closes #N` link.
- [ ] Verify commit-pr's default path (uncommitted changes) still works unchanged.

Closes #87

### DIFF
--- a/skills/ghgremlin/ghgremlin.sh
+++ b/skills/ghgremlin/ghgremlin.sh
@@ -417,7 +417,10 @@ CORE_PRINCIPLES=$(awk '/^## Core Principles/{found=1; next} found && /^## /{foun
 
 if run_stage implement; then
   # Record state before 2a so we can verify the model actually did work.
+  # PRE_BRANCH is empty on detached HEAD (the launch.sh worktree default);
+  # the post-stage adapter below treats that as "no branch to reset".
   PRE_HEAD=$(git rev-parse HEAD)
+  PRE_BRANCH=$(git symbolic-ref --short HEAD 2>/dev/null || true)
 
   set_stage implement
   echo "==> [2a/6] implementing plan"
@@ -437,18 +440,47 @@ The following is the implementation plan ${_plan_source_label}:
 
 ${ISSUE_BODY}
 
-Implement the plan above by making the code changes in this repo. Do not commit or push yet. Do NOT create any meta/scaffolding files in the repo — no \`.claude-workflow/\` directory, no \`plan.md\`, no review docs, no notes-to-self. ${_plan_location_note}" \
+Implement the plan above by making the code changes in this repo. You may commit checkpoints as you go (recommended for multi-step plans — a later stage handles branching and PR creation regardless), but do not push to any remote. Do NOT create any meta/scaffolding files in the repo — no \`.claude-workflow/\` directory, no \`plan.md\`, no review docs, no notes-to-self. ${_plan_location_note}" \
     | progress_tee)
   IMPL_SESSION=$(extract_session_id "$IMPL_OUT")
 
-  # Guard: 2a should have produced uncommitted changes and should NOT have
-  # advanced HEAD. If either invariant is broken, bail out now rather than
-  # letting 2b open an empty PR or silently drop work.
-  if [[ "$(git rev-parse HEAD)" != "$PRE_HEAD" ]]; then
-    die "implementation step advanced HEAD (expected uncommitted changes only); refusing to continue"
+  # Tolerant adapter: implement may have committed checkpoints (a normal
+  # coding-agent pattern on multi-step plans) or left work uncommitted. Both
+  # are recoverable. Empty state is still a hard error — there's nothing to PR.
+  POST_HEAD=$(git rev-parse HEAD)
+  IMPL_HEAD_ADVANCED=0
+  if [[ "$POST_HEAD" != "$PRE_HEAD" ]]; then
+    IMPL_HEAD_ADVANCED=1
   fi
-  if [[ -z "$(git status --porcelain)" ]]; then
+  if [[ "$IMPL_HEAD_ADVANCED" == "0" && -z "$(git status --porcelain)" ]]; then
     die "implementation step produced no changes; refusing to open empty PR"
+  fi
+
+  # When implement committed, move those commits onto a hand-off branch so
+  # commit-pr can push them under the canonical issue branch name without
+  # double-committing. If the gremlin started on a named branch (rather than
+  # the launch.sh detached-worktree default), reset that branch back to
+  # PRE_HEAD so the implementation commits don't leak onto the chain's
+  # start ref.
+  #
+  # Manual repro for this path: in a worktree, run `claude -p "$prompt"` with
+  # a prompt instructing the agent to make a commit, then re-enter this
+  # script at the implement-stage tail; HEAD-advanced should produce a
+  # hand-off branch and a clean original-ref reset, not a die().
+  HANDOFF_BRANCH=""
+  if [[ "$IMPL_HEAD_ADVANCED" == "1" ]]; then
+    HANDOFF_BRANCH="ghgremlin-impl-handoff-$$"
+    if git show-ref --verify --quiet "refs/heads/$HANDOFF_BRANCH"; then
+      die "hand-off branch $HANDOFF_BRANCH already exists; refusing to clobber"
+    fi
+    git switch -c "$HANDOFF_BRANCH" >/dev/null 2>&1 \
+      || die "could not create hand-off branch $HANDOFF_BRANCH"
+    if [[ -n "$PRE_BRANCH" ]]; then
+      git branch -f "$PRE_BRANCH" "$PRE_HEAD" \
+        || die "could not reset $PRE_BRANCH back to $PRE_HEAD"
+    fi
+    _commit_count=$(git rev-list --count "$PRE_HEAD..HEAD")
+    echo "    implement committed during run; moved ${_commit_count} commit(s) onto $HANDOFF_BRANCH${PRE_BRANCH:+ and reset $PRE_BRANCH}"
   fi
 fi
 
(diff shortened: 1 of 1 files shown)


xbrianh (commented):
Clean implementation of the tolerant adapter described in #87. The code preserves the empty-state guard, documents the operating assumption that the gremlin runs from a launch.sh detached-worktree (so `PRE_BRANCH` is normally empty), and threads two prompt shapes through commit-pr cleanly. A few minor observations below — nothing blocker-severity.

skills/ghgremlin/ghgremlin.sh
```diff
@@ -437,19 +440,48 @@ The following is the implementation plan ${_plan_source_label}:
...
+  # script at the implement-stage tail; HEAD-advanced should produce a
+  # hand-off branch and a clean original-ref reset, not a die().
+  HANDOFF_BRANCH=""
+  if [[ "$IMPL_HEAD_ADVANCED" == "1" ]]; then
+    HANDOFF_BRANCH="ghgremlin-impl-handoff-$$"
+    if git show-ref --verify --quiet "refs/heads/$HANDOFF_BRANCH"; then
+      die "hand-off branch $HANDOFF_BRANCH already exists; refusing to clobber"
+    fi
+    git switch -c "$HANDOFF_BRANCH" >/dev/null 2>&1 \
+      || die "could not create hand-off branch $HANDOFF_BRANCH"
```
xbrianh, line 499: Minor: redirecting both stdout and stderr to `/dev/null` swallows the actual git error message on failure, leaving operators with only the generic `could not create hand-off branch <name>` to debug. Since this only fires on a real failure, consider letting stderr through (drop the `2>&1`) so the underlying git diagnostic — e.g. "fatal: unable to write new index file", checkout conflict, etc. — surfaces in the run log.

skills/ghgremlin/ghgremlin.sh
```diff
@@ -437,19 +440,48 @@ The following is the implementation plan ${_plan_source_label}:
...
+    if [[ -n "$PRE_BRANCH" ]]; then
+      git branch -f "$PRE_BRANCH" "$PRE_HEAD" \
+        || die "could not reset $PRE_BRANCH back to $PRE_HEAD"
```
xbrianh, line 502: Worth being explicit in the comment above (or here) that this is a *destructive* rewrite of the user's local branch ref. In the launch.sh worktree path this is a no-op (`PRE_BRANCH` is empty due to `--detach`), but a direct invocation of `ghgremlin.sh` from a named branch would silently move that ref back to PRE_HEAD, dropping any reflog-only state the user might still expect to find by name. The current comment at 461–464 hints at this with "the launch.sh detached-worktree default" framing, but a one-line caveat that named-branch callers will see the branch reset would help future readers diagnose surprise.

skills/ghgremlin/ghgremlin.sh
```diff
@@ -460,10 +492,24 @@ if run_stage commit-pr; then
...
-    _pr_prompt="Create a new branch from the default branch, commit all changes with a descriptive message, push the branch, and open a PR with 'gh pr create'. Name the branch with a short descriptive slug derived from the plan title. Do NOT include any 'Closes #N' or 'Fixes #N' link in the commit message or PR body. Print ONLY the PR URL on the final line of your response."
+    _branch_clause="Name the branch with a short descriptive slug derived from the plan title."
+    _closes_clause="Do NOT include any 'Closes #N' or 'Fixes #N' link in the commit message or PR body."
   fi
+
+  # Two action shapes depending on what implement left behind. The hand-off
+  # branch only exists when implement committed during its run; otherwise
+  # we fall through to the original "create branch + commit + push" prompt.
+  if [[ "${IMPL_HEAD_ADVANCED:-0}" == "1" ]]; then
+    _commits_count=$(git rev-list --count "$PRE_HEAD..HEAD")
```
xbrianh: Nit: `_commit_count` (line 482) and `_commits_count` (here) are recomputed from the same range and differ only by a `s`. Nothing changes the worktree between the implement-stage tail and commit-pr, so you can hoist the count once and reuse — and pick one name. Cosmetic, not a bug.

skills/ghgremlin/ghgremlin.sh
```diff
@@ -460,10 +492,24 @@ if run_stage commit-pr; then
...
+    _action_clause="The implementation work is already committed on the current branch ('$HANDOFF_BRANCH'), with ${_commits_count} commit(s) above $PRE_HEAD. Rename the current branch with 'git branch -m <new-name>', push it with 'git push -u origin <new-name>', and open a PR with 'gh pr create'. Do NOT make additional commits unless required to add the 'Closes #N' line — if needed, amend the most recent commit with 'git commit --amend'."
```
xbrianh: The HEAD-advanced action clause drops the original "Create a new branch from the default branch" framing; the renamed branch's tip descends from whatever `PRE_HEAD` was. This is fine in the launch.sh-detached-worktree case (where `PRE_HEAD` ≡ the chain start ref) and `gh pr create` will still infer `--base` from the repo default. But if a caller ever invokes ghgremlin.sh directly while checked out on a feature branch ahead of `main`, those pre-existing commits will be carried into the PR diff. Worth a comment noting the assumption, or — more defensively — passing `--base <default>` explicitly in the prompt.

skills/ghgremlin/ghgremlin.sh
```diff
@@ -437,19 +440,48 @@ The following is the implementation plan ${_plan_source_label}:
...
+  # PRE_HEAD so the implementation commits don't leak onto the chain's
+  # start ref.
+  #
+  # Manual repro for this path: in a worktree, run `claude -p "$prompt"` with
+  # a prompt instructing the agent to make a commit, then re-enter this
+  # script at the implement-stage tail; HEAD-advanced should produce a
+  # hand-off branch and a clean original-ref reset, not a die().
+  HANDOFF_BRANCH=""
+  if [[ "$IMPL_HEAD_ADVANCED" == "1" ]]; then
+    HANDOFF_BRANCH="ghgremlin-impl-handoff-$$"
```
xbrianh, line 491: Naming nit: `ghgremlin-impl-handoff-$$` uses the script's PID for collision avoidance within a single invocation, but resume scenarios (where `--resume-from commit-pr` rewinds to `implement` per line 109) execute under a *different* PID. If a previous run died after `git switch -c` succeeded but before `git branch -m` in commit-pr, the stale `ghgremlin-impl-handoff-<old-pid>` branch is left behind in the worktree as cruft. Not a correctness bug — the new run picks a fresh name — but worth either documenting or having `set-stage`/`finish` sweep these. Tracking issue, not a request to fix here.

skills/ghgremlin/ghgremlin.sh
xbrianh: Done — dropped `2>&1` on the `git switch -c` failure path so the underlying git diagnostic surfaces in the run log.

skills/ghgremlin/ghgremlin.sh
xbrianh: Done — added a comment block above the `HANDOFF_BRANCH` setup that calls out the destructive-ref-rewrite behavior for direct (non-launch.sh) callers, including that any state reachable only through $PRE_BRANCH will only be findable via reflog after the reset.

skills/ghgremlin/ghgremlin.sh
xbrianh: Done — hoisted `_commit_count` into the implement block (initialized to 0, set to the real value when HEAD advanced) and reused it in commit-pr; `_commits_count` is gone.

skills/ghgremlin/ghgremlin.sh
xbrianh: Done — both directions. Added a comment in commit-pr explaining the PRE_HEAD-as-base assumption under launch.sh's detached worktree, and updated the prompt to pass `--base <default-branch>` explicitly so a direct invocation from a feature branch ahead of the default branch can't silently carry pre-existing commits into the PR diff. Applied the same `--base` clause to the non-HEAD-advanced path for symmetry.

skills/ghgremlin/ghgremlin.sh
xbrianh: OOS — you flagged this as a tracking issue, not a fix request, and the cruft case is real. Filed as #94.

skills/ghgremlin/ghgremlin.sh
```diff
@@ -437,19 +440,48 @@ The following is the implementation plan ${_plan_source_label}:
...
-  # letting 2b open an empty PR or silently drop work.
-  if [[ "$(git rev-parse HEAD)" != "$PRE_HEAD" ]]; then
-    die "implementation step advanced HEAD (expected uncommitted changes only); refusing to continue"
+  # Tolerant adapter: implement may have committed checkpoints (a normal
+  # coding-agent pattern on multi-step plans) or left work uncommitted. Both
+  # are recoverable. Empty state is still a hard error — there's nothing to PR.
+  POST_HEAD=$(git rev-parse HEAD)
+  IMPL_HEAD_ADVANCED=0
+  if [[ "$POST_HEAD" != "$PRE_HEAD" ]]; then
+    IMPL_HEAD_ADVANCED=1
```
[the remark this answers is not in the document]
xbrianh: Done — added a `git merge-base --is-ancestor "$PRE_HEAD" "$POST_HEAD"` guard. HEAD changes that aren't a fast-forward from PRE_HEAD now hard-fail with an explicit error instead of being treated as committed work to hand off.

skills/ghgremlin/ghgremlin.sh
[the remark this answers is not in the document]
xbrianh: Done — branched the HEAD-advanced commit-pr prompt on `git status --porcelain`. Clean worktree gets the original "already committed; rename + push" shape; dirty worktree gets a variant that explicitly tells the agent to include the remaining changes (amend or new commit) before pushing so they end up in the PR.

Issue #80, opened 2026-04-26
Phased boss-gremlin spec: extract pipeline/ package (ref #77)


xbrianh:
## Context

This is a boss-gremlin-ready phased spec for #77 ("Extract shared gremlin pipeline into top-level `pipeline/` Python package"). The original issue captures the full design rationale, the strawman module layout, and the alternatives considered; this spec slices that work into discrete, serially-landable phases sized for a `bossgremlin --chain-kind gh` run.

Each phase corresponds to one child gremlin and lands as one squash-merged PR. The handoff agent consumes this spec as the read-only "north star" and uses the accumulated diff to decide which phase comes next.

Ref #77. Ref #58 (the test-coverage issue this refactor unblocks).

## Goal

End state, after all phases land:

- `pipeline/` package at the repo root, mirrored to `~/.claude/pipeline/` via `scripts/sync.sh`. Stdlib-only, no third-party deps.
- All three gremlin orchestrators (`localgremlin`, `ghgremlin`, `bossgremlin`) implemented inside `pipeline/`, invoked as `python -m pipeline.cli {local,gh,boss} ...`.
- `claude -p` shell-out routed through an injectable `ClaudeClient` so unit tests can substitute a fake.
- `skills/<gremlin>/` directories contain only SKILL.md plus a thin shim launcher that delegates to `pipeline/`.
- A working pytest suite under `pipeline/tests/` exercising the new seam (the prerequisite #58 was waiting on).

## Approach (carry-over from #77)

- Top-level `pipeline/` directory at repo root, NOT under `skills/` (avoids being rsync'd into `~/.claude/skills/` as a fake skill).
- Distribute via `scripts/sync.sh` by extending `DIR_PAIRS` with `pipeline:$CLAUDE_DIR/pipeline`. No pip install, no venv.
- The `ClaudeClient` seam is the testability unlock: define an injectable interface, route every `claude -p` call through it.
- SKILL.md files stay where Claude Code expects them (`skills/<name>/SKILL.md`); only the executable scripts move.
- Foreground-skill prompts (`/handoff`, `/localreview`, `/localaddress`, `/design`) are out of scope — they keep their inline SKILL.md prompts.

## Constraints (must hold across the chain)

- **One phase per child gremlin.** Do not bundle phases. Each child produces one PR; the next phase begins only after the previous phase has landed.
- **Stdlib-only inside `pipeline/`.** No third-party imports. If a phase wants to add one, it bails and asks.
[...]

PR #95, opened 2026-04-26
handoff: filter operator tasks out of child plans

### LINKED ISSUE #88: handoff: filter operator/deploy tasks out of child plans (sync.sh push, recursive gremlin launches, etc.)
## Context

The handoff agent (`handoff.py`) propagates spec content verbatim into child plans without distinguishing **implementation tasks** (code that lands in a PR) from **operator/deploy tasks** (actions a human takes outside the worktree).

Concrete repro from the active boss chain `extract-pipeline-package-6-phase-chain-61c255`. The Phase 1 spec section reads (in part):

- Task 16: Run `scripts/sync.sh push` to mirror `pipeline/` to `~/.claude/pipeline/`
- Task 17: Smoke-run `/localgremlin` end-to-end through the rsync'd package; verify `/gremlins rescue` works against a deliberately failed local gremlin

Both made it into `handoff-002-child.md` for the Phase 1 child gremlin. Both are impossible-or-destructive when executed inside a gremlin's detached-HEAD worktree:

- **`scripts/sync.sh push`** mirrors a feature-branch worktree's contents into the user's *live* `~/.claude/`. The user would suddenly be running unmerged code.
- **`/localgremlin` from inside another gremlin** is a recursive launch from an already-detached worktree — undefined behavior at minimum, state corruption at worst.

Both also have the side effect of forcing the implementing agent to commit (sync.sh push requires committed state), which then trips the implement-stage HEAD-advance guard (separately tracked).

The boss spec at issue #80 includes these as operator-level acceptance criteria for each phase. They belong in the spec — they describe what a human verifies after a phase lands. They do **not** belong in a child gremlin's plan.

## Approach

Add a pre-implementation filter to the handoff agent. When generating a child plan, the agent identifies and excludes tasks that:

- Mutate `~/.claude/` directly (e.g. `sync.sh push`, `sync.sh pull`, hand-edits under `~/.claude/`).
- Launch another gremlin (`/localgremlin`, `/ghgremlin`, `/bossgremlin`).
- Push to a remote outside the PR flow (e.g. `git push origin main`, force-pushes, manual `gh pr merge`).
- Run `/gremlins land` / `rescue` / `stop` — these are operator commands, not workflow steps.

Implementation surface in `handoff.py`: extend the prompt to the inner `claude -p` agent with explicit instructions about the implementation-vs-operator boundary. Provide examples of each. Make clear the agent should rewrite or drop operator tasks rather than copy them — and surface them in the **rolling plan** (the parent doc) under a separate "Operator follow-ups" or similar section so they aren't lost; the boss/operator owns them between phase landings.

The boundary is: **a child gremlin operates inside a detached worktree, against a single feature branch, ending in one PR.** Anything that requires being outside that scope (the user's live config, another worktree, multiple branches, post-merge actions) is an operator task.

## Tasks

- [ ] Extend the handoff agent's inner prompt in `handoff.py` with the operator-vs-implementation boundary, examples, and the "rewrite or drop, don't copy" directive. The prompt is currently constructed inline — find the spot that builds the agent's input.
- [ ] Add an "Operator follow-ups" (or similarly named) section to the rolling plan template the agent writes. Tasks the agent classifies as operator-level land there instead of in the child plan.
- [ ] Add a self-check pass: before writing the child plan, the agent re-reads its own output and asks "is every remaining task something a code-only gremlin in a detached worktree can do?" If not, revise.
- [ ] Document the boundary in `skills/handoff/SKILL.md` so future spec authors know what gets filtered.
- [ ] Update the spec at issue #80 in this same PR (or a follow-up) to label its operator-level tasks explicitly — even with the filter, clear authorial intent helps the agent classify correctly.

## Risks

- **False positives.** A task like "update CLAUDE.md to mention pipeline/" mentions `~/.claude/`-adjacent content (since CLAUDE.md is mirrored) but is genuinely a code change in the repo. The filter must distinguish "edit the file in the repo" from "run sync.sh push." A keyword filter alone won't work — the agent has to reason about the action.
- **Lost work.** If an operator task gets filtered to "Operator follow-ups" but the operator never sees it, the chain ships with that step undone. Mitigate by surfacing the follow-ups list at chain-done time (e.g. in the boss's final summary log) and consider a `/gremlins followups <boss-id>` view.
- **Spec drift.** If the spec author writes operator tasks inline with implementation tasks, the agent has to be confident enough to ignore them. The "rewrite or drop, don't copy" directive must be strong; today's handoff prompt is permissive about copying spec content.

## Open questions

- Should the handoff agent refuse to produce a child plan whose operator-task count is high (e.g. >2), on the theory that the spec author intended the phase to be operator-driven and the boss should bail rather than spawn a doomed child? Lean: yes — `bail` is a recognized handoff exit state, and "this phase isn't a code-only phase" is a legitimate bail reason.
- Where do operator follow-ups actually surface to the user? In `boss_state.json` post-chain? In a markdown file alongside the rolling plan? Both?

Ref the dead chain `extract-pipeline-package-6-phase-chain-61c255` and child gremlin `phase-1-scaffold-pipeline-package-and-3453cd` (planning issue #86) as the concrete repro. Ref #80 (the spec that surfaced this).

xbrianh:
## Summary

- Extends the handoff agent's prompt with an explicit implementation-vs-operator boundary so tasks that can't run inside a detached-HEAD worktree (e.g. `scripts/sync.sh push`, recursive gremlin launches, `/gremlins land`, post-merge verification) no longer get copied verbatim into child plans.
- Operator tasks route to a new `## Operator follow-ups` section in the rolling plan, where the human picks them up between phase landings. Carried forward across handoffs; surfaced even at `chain-done`/`bail` so they aren't lost.
- Adds a self-check pass before the agent writes the child plan, plus a `bail`-preference heuristic when remaining work is dominated by operator tasks.
- Documents the boundary in `skills/handoff/SKILL.md` so future spec authors know what gets filtered (and can pre-label operator acceptance criteria for clarity).

Concrete repro this addresses: the dead boss chain `extract-pipeline-package-6-phase-chain-61c255`, where Phase 1's spec leaked `scripts/sync.sh push` and `/localgremlin` smoke-run tasks into the child plan.

## Test plan

- [ ] Re-run a `bossgremlin --chain-kind gh` against the issue #80 spec and confirm Phase 1's child plan no longer contains `sync.sh push` or `/localgremlin` invocations
- [ ] Confirm the rolling plan emitted by handoff #1 shows an `## Operator follow-ups` section listing the filtered items
- [ ] Confirm a chain whose remaining work is operator-only exits as `bail` rather than spawning a doomed child
- [ ] Spot-check that genuinely-implementation edits to mirrored files (`CLAUDE.md`, `scripts/sync.sh`) still flow into the child plan

Closes #88

### DIFF
--- a/skills/handoff/SKILL.md
+++ b/skills/handoff/SKILL.md
@@ -21,10 +21,19 @@ This skill runs **in the foreground** — it blocks until the inner agent finish
 ## Exit states
 
 - **`next-plan`**: implementation work remains; a child plan has been written for the next gremlin.
-- **`chain-done`**: all implementation tasks are landed; the chain is complete. Pending operator tasks do not block this state — they surface in the final rolling plan's `## Operator follow-ups` section.
-- **`bail`**: something prevents safe continuation; reason is in the signal file and updated plan. Includes the case where remaining work is dominated by operator tasks with no coherent code-only chunk to hand a child.
+- **`chain-done`**: all implementation tasks are landed; the chain is complete. Pending operator tasks do **not** block this state and are **not** an alternative bail reason — they are surfaced via the signal file's `operator_followups` array (and the final rolling plan's `## Operator follow-ups` section, if any pending). A chain whose remaining work is operator-only exits as `chain-done`.
+- **`bail`**: something prevents safe continuation; reason is in the signal file and updated plan. Reserved for genuine blockers (broken state, incoherent plan, security issue, etc.) — operator-only remaining work is not one of them.
 
 Script exit code 0 on any recognized outcome; 1 on infrastructure failure (read the signal file to distinguish outcomes).
+
+## Signal file
+
+The signal file written to `<out-stem>.state.json` always contains:
+
+- `exit_state` — `"next-plan"`, `"chain-done"`, or `"bail"`.
+- `child_plan` — absolute path to the child plan when `next-plan`; `null` otherwise.
+- `reason` — short string when `bail`; `null` otherwise.
+- `operator_followups` — array of one-line strings, one per pending operator task. Empty array (`[]`) when there are none. Populated on every exit state, including `chain-done`, so the boss orchestrator can persist the final operator-task list even when the rolling plan has been pruned to a "chain complete" note.
 
 ## Implementation vs operator tasks
 
--- a/skills/handoff/SKILL.md
+++ b/skills/handoff/SKILL.md
@@ -20,11 +20,30 @@ This skill runs **in the foreground** — it blocks until the inner agent finish
 
 ## Exit states
 
-- **`next-plan`**: work remains; a child plan has been written for the next gremlin.
-- **`chain-done`**: all tasks are implemented; the chain is complete.
-- **`bail`**: something prevents safe continuation; reason is in the signal file and updated plan.
+- **`next-plan`**: implementation work remains; a child plan has been written for the next gremlin.
+- **`chain-done`**: all implementation tasks are landed; the chain is complete. Pending operator tasks do not block this state — they surface in the final rolling plan's `## Operator follow-ups` section.
+- **`bail`**: something prevents safe continuation; reason is in the signal file and updated plan. Includes the case where remaining work is dominated by operator tasks with no coherent code-only chunk to hand a child.
 
 Script exit code 0 on any recognized outcome; 1 on infrastructure failure (read the signal file to distinguish outcomes).
+
+## Implementation vs operator tasks
+
+A child gremlin runs inside a **detached-HEAD worktree, against a single feature branch, ending in one squash-merged PR**. The handoff agent classifies every still-open task in the input plan as either:
+
+- **Implementation** — a code/doc/config change that lands inside the child's PR. These flow into the child plan.
+- **Operator** — work that requires escaping the worktree scope. These are filtered out of the child plan and land in the rolling plan's `## Operator follow-ups` section for the human to pick up between phase landings.
+
+Operator signals the agent looks for:
+
+- Mutates `~/.claude/` or other live user state (`scripts/sync.sh push`/`pull`, hand-edits under `~/.claude/`).
+- Launches another gremlin (`/localgremlin`, `/ghgremlin`, `/bossgremlin`, end-to-end smoke runs).
+- Pushes outside the PR flow (`git push origin main`, force-pushes, manual merges).
+- `/gremlins` operator commands (`land`, `rescue`, `stop`, `close`, `rm`).
+- Post-merge verification (PR CI green checks, deploy confirmation, dashboard watching).
+
+Edits to tracked repo files like `CLAUDE.md`, `scripts/sync.sh`, or `pipeline/DESIGN.md` are **implementation**, even if those files are mirrored to `~/.claude/` by `sync.sh` — the child edits the in-repo file, the human runs the sync later.
+
+Spec authors should label genuinely operator-level acceptance criteria explicitly (e.g. "Operator verification:" prefix, or a separate `## Operator acceptance` section) so the agent classifies correctly. The filter is robust to inline mixing, but explicit labels reduce ambiguity.
 
 ## Arguments
 
--- a/skills/handoff/handoff.py
+++ b/skills/handoff/handoff.py
@@ -147,26 +147,51 @@ Git diff since chain start:
 ```
 {diff_trunc}
 
+## Implementation vs operator boundary
+
+A child gremlin operates **inside a detached-HEAD worktree, against a single feature branch, ending in one squash-merged PR**. Anything that requires being outside that scope — the user's live config, another worktree, multiple branches, post-merge actions, sibling gremlin launches — is an **operator task**, owned by the human between phase landings.
+
+Classify a task as **operator** if executing it inside a child gremlin's worktree would be impossible, destructive, or undefined. Concrete signals:
+
+- **Mutates `~/.claude/` or other live user state directly**: `scripts/sync.sh push`, `scripts/sync.sh pull`, hand-edits under `~/.claude/`, copying files into `~/.claude/`. (The child has unmerged code in its worktree; mirroring that into `~/.claude/` would suddenly run unmerged code on the user's machine.)
+- **Launches another gremlin**: `/localgremlin`, `/ghgremlin`, `/bossgremlin`, or a smoke-run / end-to-end run that boils down to invoking one. Recursive gremlin launch from a detached worktree is undefined behavior.
+- **Pushes to a remote outside the PR flow**: `git push origin main`, force-pushes, manual `gh pr merge`, direct merges. The child's only remote interaction is opening (and updating) one PR.
+- **Operator commands**: `/gremlins land`, `/gremlins rescue`, `/gremlins stop`, `/gremlins close`, `/gremlins rm`. These are human controls, not workflow steps.
+- **Post-merge verification**: "verify the merged PR's CI is green", "confirm the production deploy", "watch the release dashboard". The child finishes before its PR merges.
+
+Classify as **implementation** if it is a code/doc/config change that lands in the child's PR. Examples that *look* operator-adjacent but are implementation:
+
+- "Update `CLAUDE.md` to mention `pipeline/`" — edits a tracked repo file. The fact that `CLAUDE.md` is mirrored to `~/.claude/CLAUDE.md` by `sync.sh` is irrelevant: the child edits the in-repo file, the human runs `sync.sh push` later.
+- "Extend `scripts/sync.sh` `DIR_PAIRS` with `pipeline:...`" — edits a tracked script. Implementation.
+- "Add `pipeline/DESIGN.md`" — creates a tracked file. Implementation.
+- "Run `scripts/sync.sh push --dry-run` and confirm output is clean" — operator. The dry-run reads live `~/.claude/` state and isn't a code change. (A dry-run *check encoded as a unit test* against fixture data would be implementation; a real `--dry-run` invocation against the user's live tree is not.)
+
+The distinction is **what the task changes** (tracked repo files = implementation) vs **what the task reads or mutates outside the worktree** (live user config, sibling processes, remotes outside the PR = operator). When in doubt, ask: "Could a fresh gremlin with no access to my home directory do this?" If no, operator.
+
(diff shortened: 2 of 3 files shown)


xbrianh (commented):

xbrianh (commented):
Substantively this is a good direction — the implementation/operator boundary is well-articulated and the examples (especially the `CLAUDE.md` / `sync.sh push` callouts) ground it concretely. The main thing worth resolving before merging is a design ambiguity in the exit-state rules: `chain-done` and the new operator-only `bail` overlap on the same scenario (0 implementation tasks remaining + N operator tasks remaining), and the LLM will need to pick one. A couple of smaller prompt-clarity nits inline.

skills/handoff/handoff.py
```diff
@@ -147,26 +147,51 @@ def build_prompt(
...
-   - **`bail`**: something prevents safe continuation (broken state, incoherent plan, security issue, etc.).
+3. Classify every still-open task as **implementation** or **operator** using the boundary above. Operator tasks never land in a child plan.
+4. Decide the exit state:
+   - **`chain-done`**: all *implementation* tasks in the plan are implemented and landed. (Operator tasks do not block `chain-done` — they are surfaced separately for the human operator.)
+   - **`next-plan`**: at least one *implementation* task remains; the next gremlin should tackle it.
+   - **`bail`**: something prevents safe continuation (broken state, incoherent plan, security issue, etc.). Also bail when the remaining work is essentially all operator tasks and there is no coherent code-only chunk to hand a child — a chain whose remaining phase is operator-driven should halt for the human, not spawn a doomed child. As a rough heuristic, if more than ~2 of the remaining tasks are operator-classified relative to the implementation tasks, prefer `bail` with a reason that names the operator-only nature of the remaining work.
```
xbrianh: Two issues on this bail rule:

**1. It overlaps with `chain-done`.** Consider the scenario where every implementation task in the input plan is landed, and only operator tasks remain. By line 179 that is `chain-done` ("all *implementation* tasks ... are implemented and landed. Operator tasks do not block `chain-done`"). By this bullet it is `bail` ("remaining work is essentially all operator tasks ... a chain whose remaining phase is operator-driven should halt for the human"). Both literally apply, and the choice flips chain semantics for the boss orchestrator. Worth disambiguating — e.g. scope this bail clause to the case where some implementation tasks remain but they cannot be coherently chunked into a child PR, and let the all-impl-done case fall through to `chain-done` unconditionally.

**2. The `~2 ... relative to the implementation tasks` heuristic is ambiguous.** "More than ~2 of the remaining tasks are operator-classified relative to the implementation tasks" reads three different ways: (a) `op_count > 2`, (b) `op_count - impl_count > 2`, (c) `op_count / impl_count > 2`. With 1 impl + 5 operator the right answer is probably `next-plan` with the 1 impl task — but the heuristic plausibly pushes the LLM to bail. Either drop the numeric threshold and rely on "no coherent code-only chunk", or state it concretely (e.g. "bail when zero implementation tasks remain").

skills/handoff/handoff.py
```diff
@@ -147,26 +147,51 @@ def build_prompt(
...
+- "Add `pipeline/DESIGN.md`" — creates a tracked file. Implementation.
+- "Run `scripts/sync.sh push --dry-run` and confirm output is clean" — operator. The dry-run reads live `~/.claude/` state and isn't a code change. (A dry-run *check encoded as a unit test* against fixture data would be implementation; a real `--dry-run` invocation against the user's live tree is not.)
+
+The distinction is **what the task changes** (tracked repo files = implementation) vs **what the task reads or mutates outside the worktree** (live user config, sibling processes, remotes outside the PR = operator). When in doubt, ask: "Could a fresh gremlin with no access to my home directory do this?" If no, operator.
+
+If the spec author wrote operator-flavoured language inline with implementation work, **rewrite or drop it; do not copy it verbatim into the child plan**. Operator tasks land only in the rolling plan's `## Operator follow-ups` section, where the human operator picks them up between phase landings.
+
 ## Your task
 
 1. Read the plan. Identify every task listed under `## Tasks`.
```
xbrianh: Once a previous handoff has run, the input plan can carry an `## Operator follow-ups` section in addition to `## Tasks`. The carry-forward instruction at line 190 implicitly assumes the agent reads that section, but step 1 only says "every task listed under `## Tasks`". Worth widening to: "identify every task listed under `## Tasks`, plus any pending entries under `## Operator follow-ups` from the previous rolling plan" so step 3's classification step covers them too.

skills/handoff/handoff.py
```diff
@@ -147,26 +147,51 @@ def build_prompt(
...
      - Under `## Open questions`, carry forward unresolved entries; drop entries tied to completed tasks.
      - If a task is only partly landed, keep it (rewritten if needed to reflect what remains).
-   - **`chain-done`**: minimal output. A short note that the chain is complete is enough — no leftover task list, no carried-over context. The signal file carries the structured outcome.
-   - **`bail`**: same pruning rules as `next-plan` (only remaining tasks, surrounding sections trimmed accordingly, unresolved `## Open questions` carried forward), with a bail-reason banner added prominently at the top.
+     - Add an `## Operator follow-ups` section listing every task you classified as operator-level (still pending). These are not handed to a child gremlin — they wait for the human between phase landings. Carry forward operator follow-ups from the previous rolling plan if they have not been done; drop ones that have. If there are no pending operator tasks, omit the section.
```
xbrianh: Minor: the agent is told to "drop ones that have [been done]" but operator tasks by definition leave no signal in the git diff (`sync.sh push`, recursive gremlin launches, post-merge verification, etc. all happen outside the worktree). Without a completion signal in the inputs, the safe behavior is conservative carry-forward. Worth saying so explicitly — e.g. "carry forward all pending operator follow-ups; only drop one if the input plan or git history makes its completion unambiguous" — otherwise the LLM may invent completion judgments.

skills/handoff/SKILL.md
```diff
@@ -20,12 +20,31 @@ This skill runs **in the foreground** — it blocks until the inner agent finish
 
 ## Exit states
 
-- **`next-plan`**: work remains; a child plan has been written for the next gremlin.
-- **`chain-done`**: all tasks are implemented; the chain is complete.
-- **`bail`**: something prevents safe continuation; reason is in the signal file and updated plan.
+- **`next-plan`**: implementation work remains; a child plan has been written for the next gremlin.
+- **`chain-done`**: all implementation tasks are landed; the chain is complete. Pending operator tasks do not block this state — they surface in the final rolling plan's `## Operator follow-ups` section.
+- **`bail`**: something prevents safe continuation; reason is in the signal file and updated plan. Includes the case where remaining work is dominated by operator tasks with no coherent code-only chunk to hand a child.
```
xbrianh: Same chain-done/bail overlap as flagged in handoff.py:181 — line 24 says operator tasks don't block `chain-done`, line 25 says operator-dominated remaining work bails. Whatever resolution lands in the prompt, mirror it here.

xbrianh (commented):

xbrianh:
## Design verdict on the exit-state overlap

Reviewer flagged that the implementation has both `chain-done` and an "operator-only bail" exit applying when only operator tasks remain — author needs to pick one.

**Verdict: `chain-done` + structured operator-followups list.** No "operator-only bail" exit state.

Rationale and downstream implementation in xbrianh/claude#88 (comment).

Specific changes this PR needs:

1. Drop the operator-only bail code path entirely. There's only one exit when no implementation tasks remain: `chain-done`.
2. `handoff.py` produces an `operator_followups: [string]` array in the signal file when returning `chain-done` and there were operator tasks in the rolling plan.
3. `bossgremlin.py` reads the followups, persists them into `boss_state.json`, and prints them in the chain-done log line so they land in the boss's `log` artifact.
4. (Out of scope here, follow-up issue welcome: `/gremlins` drill-in display of `final_operator_followups`.)

After updating, this PR should not need a fresh review pass on the design — only on the implementation of the resolved design.

skills/handoff/handoff.py
```diff
@@ -147,26 +147,51 @@ def build_prompt(
...
+   - **`chain-done`**: minimal output. A short note that the chain is complete is enough — no leftover task list, no carried-over context. If any operator follow-ups remain pending, list them under `## Operator follow-ups` so the human sees them in the final rolling plan; otherwise omit. The signal file carries the structured outcome.
```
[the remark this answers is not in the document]
xbrianh: Addressed. The carry-forward rule now explicitly treats the input plan's `## Operator follow-ups` section as authoritative — items are only dropped when the input plan or git history makes completion unambiguous, never inferred from diff/log alone.

skills/handoff/handoff.py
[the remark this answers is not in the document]
xbrianh: Addressed. The bail heuristic was dropped entirely per the design verdict in the issue comment above — operator-only remaining work now exits as `chain-done` and the operator tasks propagate to the boss via the new `operator_followups` field in the signal file, rather than as a bail.

xbrianh: Addressed both. The operator-only bail clause is dropped entirely (per the design verdict above) — operator-only remaining work now unambiguously exits as `chain-done`, and operator tasks reach the boss via the new `operator_followups` array in the signal file. The numeric heuristic is gone with it.

skills/handoff/handoff.py
xbrianh: Addressed. Step 1 now reads: "Identify every task listed under `## Tasks`, plus every pending entry under `## Operator follow-ups` if the input plan has that section (a previous handoff may have written it). Both sets feed step 3's classification."

skills/handoff/handoff.py
xbrianh: Addressed. The carry-forward rule is now explicitly conservative: "Only drop an entry if the input plan or git history makes its completion unambiguous (e.g. the human/operator removed it from the input plan, or a commit message explicitly states the operator step was done). Do not infer completion from git diff/log or implementation progress alone — operator tasks happen outside the worktree, so the safe behavior is conservative carry-forward."

skills/handoff/SKILL.md
xbrianh: Addressed. SKILL.md now mirrors the resolved design: `chain-done` explicitly notes that operator-only remaining work exits as `chain-done` (not bail), and `bail` is described as reserved for genuine blockers. Also added a `## Signal file` section documenting the new `operator_followups` field.

PR #96, opened 2026-04-26
ghgremlin: sweep stale impl-handoff branches

### LINKED ISSUE #94: ghgremlin: hand-off branch with PID-suffix can leak across resume scenarios
The implement-stage tolerant adapter names its hand-off branch `ghgremlin-impl-handoff-$$` (using the running script's PID for collision avoidance within a single invocation). On `--resume-from commit-pr`, the resume rewinds to the implement stage but executes under a *different* PID. If a previous run died after `git switch -c "ghgremlin-impl-handoff-<old-pid>"` succeeded but before commit-pr's `git branch -m` renamed it, the stale branch is left behind in the worktree as cruft.

This isn't a correctness bug — the new run's `$$` is fresh, so it picks an uncolliding name and proceeds. But over time, repeated rescues can leave an arbitrary number of `ghgremlin-impl-handoff-*` branches in the worktree.

Possible directions:
- Have `set-stage` / `finish` (or the rescue path) sweep `ghgremlin-impl-handoff-*` branches whose tip has been consumed.
- Use a fixed name (e.g. `ghgremlin-impl-handoff`) and refuse to start when one already exists, so the cruft is loud rather than silent.
- Document the cruft and rely on worktree teardown to clean it up.

Reviewer comment:
> Naming nit: `ghgremlin-impl-handoff-$$` uses the script's PID for collision avoidance within a single invocation, but resume scenarios (where `--resume-from commit-pr` rewinds to `implement` per line 109) execute under a *different* PID. If a previous run died after `git switch -c` succeeded but before `git branch -m` in commit-pr, the stale `ghgremlin-impl-handoff-<old-pid>` branch is left behind in the worktree as cruft. Not a correctness bug — the new run picks a fresh name — but worth either documenting or having `set-stage`/`finish` sweep these. Tracking issue, not a request to fix here.

Ref #92
https://github.com/xbrianh/claude/pull/92#discussion_r3142934652

xbrianh:
## Summary
- Sweep stale `ghgremlin-impl-handoff-*` branches in the implement-stage tail before creating the new one.
- A run that died between `git switch -c "ghgremlin-impl-handoff-$$"` and commit-pr's `git branch -m` leaves the branch behind under the dead PID's name; the next run's fresh PID avoids a collision (so this isn't a correctness bug), but the cruft accumulates.
- `git branch -D` refuses to delete a branch checked out in any worktree, so concurrent gremlins are safe.

Closes #94

## Test plan
- [ ] Manual: in a worktree, simulate a crashed run by creating a `ghgremlin-impl-handoff-9999` branch (`git branch ghgremlin-impl-handoff-9999 HEAD`) and confirm a fresh `ghgremlin.sh` run sweeps it before creating its own hand-off branch.
- [ ] Manual: confirm a `ghgremlin-impl-handoff-*` branch checked out in another worktree is left alone (no error, no deletion).

### DIFF
--- a/skills/bossgremlin/bossgremlin.py
+++ b/skills/bossgremlin/bossgremlin.py
@@ -191,6 +191,11 @@ def init_boss_state(spec_path: str, chain_kind: str, chain_base_ref: str,
         "current_child_id": None,
         "children": [],
         "handoff_records": [],
+        # Latest operator_followups list reported by handoff. Each handoff
+        # rewrites this with the conservative carry-forward set the handoff
+        # agent produced, so by chain-done it holds the final list of
+        # operator tasks the human still owes between phase landings.
+        "operator_followups": [],
     }
     save_json(os.path.join(state_dir, "boss_state.json"), boss_state)
     return boss_state
@@ -289,6 +294,15 @@ def run_handoff(gr_id: str, state_dir: str, boss_state: dict,
     if exit_state not in ("next-plan", "chain-done", "bail"):
         die(f"handoff signal file has unrecognized exit_state: {exit_state!r}")
 
+    # Coerce operator_followups to a list of strings. Old handoff signals
+    # predating the field land here as None or absent; treat that as no
+    # followups so a chain that started under an older handoff still reads.
+    raw_followups = sig.get("operator_followups")
+    if isinstance(raw_followups, list):
+        followups = [str(item) for item in raw_followups if str(item).strip()]
+    else:
+        followups = []
+
     now = datetime.datetime.utcnow().strftime("%Y-%m-%dT%H:%M:%SZ")
     boss_state["handoff_records"].append({
         "timestamp": now,
@@ -299,12 +313,18 @@ def run_handoff(gr_id: str, state_dir: str, boss_state: dict,
         "exit_state": exit_state,
         "child_plan": sig.get("child_plan"),
         "bail_reason": sig.get("reason"),
+        "operator_followups": followups,
     })
     boss_state["handoff_count"] = n
+    boss_state["operator_followups"] = followups
     if os.path.isfile(out_path):
         boss_state["current_plan"] = out_path
 
     log(f"handoff {n} result: {exit_state}")
+    if followups:
+        log(f"  operator follow-ups carried by handoff {n}: {len(followups)}")
+        for item in followups:
+            log(f"    - {item}")
     return exit_state, sig
 
 
@@ -493,6 +513,16 @@ def main(argv):
 
             if exit_state == "chain-done":
                 log("chain complete")
+                followups = boss_state.get("operator_followups") or []
+                if followups:
+                    log(
+                        f"operator follow-ups ({len(followups)}) — owed by the "
+                        f"human between phase landings:"
+                    )
+                    for item in followups:
+                        log(f"  - {item}")
+                else:
+                    log("operator follow-ups: (none)")
                 set_stage(gr_id, "done")
                 save_boss_state(state_dir, boss_state)
                 sys.exit(0)
--- a/skills/ghgremlin/ghgremlin.sh
+++ b/skills/ghgremlin/ghgremlin.sh
@@ -488,6 +488,19 @@ Implement the plan above by making the code changes in this repo. You may commit
   HANDOFF_BRANCH=""
   _commit_count=0
   if [[ "$IMPL_HEAD_ADVANCED" == "1" ]]; then
+    # Sweep stale hand-off branches from prior crashed runs. A run that died
+    # between `git switch -c` below and commit-pr's `git branch -m` leaves
+    # its branch under the dead PID's name; the new run picks a fresh PID
+    # so the live work isn't blocked, but the stale branch lingers as cruft.
+    # `git branch -D` refuses to delete a branch checked out in any worktree,
+    # so concurrent gremlins are safe.
+    while IFS= read -r _stale; do
+      [[ -n "$_stale" ]] || continue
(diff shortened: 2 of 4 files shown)


xbrianh (commented):
## Scope review

### Hitchhikers (significant)

The plan describes one piece of work: sweep stale `ghgremlin-impl-handoff-*` branches. That work is delivered in commit `a786c7b` and is exactly 13 lines in `skills/ghgremlin/ghgremlin.sh` — clean and on-target.

The PR also contains two unrelated commits:

- `a983e6a` — *handoff: filter operator tasks out of child plans*
- `54922c3` — *handoff: drop operator-only bail; surface operator_followups via signal file*

These touch `skills/handoff/SKILL.md`, `skills/handoff/handoff.py`, and `skills/bossgremlin/bossgremlin.py`, adding the entire "implementation vs operator task" classification feature, the `operator_followups` signal-file field, and the boss-orchestrator carry-forward / final-summary logic. None of that is mentioned in the plan, the reviewer comment, or the issue (#94) the sweep work closes — it is a separate feature concern.

Recommended fix: rebase this PR down to just `a786c7b` (the sweep) and land the two operator_followups commits as their own PR. They are logically independent — one is worktree-cruft hygiene in `ghgremlin.sh`, the other is a behavioral change to the handoff signal protocol — and bundling them makes both harder to bisect and review.

### Mixed-concern commits

Each individual commit is concern-pure, but the PR conflates the two concerns the plan kept separate. Splitting the PR also resolves this.

### Skipped plan items / exceeded scope / behavior change without tests

None — the sweep change itself matches the plan exactly and is a self-contained hygiene fix where unit tests would be disproportionate.

skills/bossgremlin/bossgremlin.py
```diff
@@ -191,6 +191,11 @@ def init_boss_state(spec_path: str, chain_kind: str, chain_base_ref: str,
         "current_child_id": None,
         "children": [],
         "handoff_records": [],
+        # Latest operator_followups list reported by handoff. Each handoff
+        # rewrites this with the conservative carry-forward set the handoff
+        # agent produced, so by chain-done it holds the final list of
+        # operator tasks the human still owes between phase landings.
+        "operator_followups": [],
```
xbrianh, line 198: These lines (and the rest of the `operator_followups` plumbing in this file, plus the changes to `skills/handoff/`) are not mentioned anywhere in the PR title, description, or the linked issue #94. They come from commits `a983e6a` and `54922c3`, which are conceptually separate from the sweep-stale-handoff change in `a786c7b`.

Consider splitting: keep this PR focused on #94 (just the `ghgremlin.sh` sweep, ~13 lines), and open a separate PR for the operator-followups feature with its own description explaining the implementation/operator boundary, the conservative carry-forward rule, and the signal-file format change. That would make both reviews easier to reason about and bisect.

skills/ghgremlin/ghgremlin.sh
```diff
@@ -488,6 +488,19 @@ Implement the plan above by making the code changes in this repo. You may commit
...
   if [[ "$IMPL_HEAD_ADVANCED" == "1" ]]; then
+    # Sweep stale hand-off branches from prior crashed runs. A run that died
+    # between `git switch -c` below and commit-pr's `git branch -m` leaves
+    # its branch under the dead PID's name; the new run picks a fresh PID
+    # so the live work isn't blocked, but the stale branch lingers as cruft.
+    # `git branch -D` refuses to delete a branch checked out in any worktree,
+    # so concurrent gremlins are safe.
+    while IFS= read -r _stale; do
+      [[ -n "$_stale" ]] || continue
+      if git branch -D "$_stale" 2>/dev/null; then
```
xbrianh, line 499: Minor: `2>/dev/null` swallows every failure mode here, not just "branch is checked out in another worktree." That's deliberate per the comment above (and correct for the common case — concurrent gremlins, branches still in use), but it also hides genuinely unexpected errors (corrupt ref database, fs permissions, etc.). Since the sweep is purely janitorial and any such error would have surfaced in the next operation anyway, leaving stderr suppressed is probably fine — just calling out that you're trading diagnosability for noise reduction in this loop.

xbrianh (commented):
## Review summary

The sweep change itself (`skills/ghgremlin/ghgremlin.sh`) is small, well-commented, and correct — `git for-each-ref` + `git branch -D` is exactly the right tool, and the comment about `git branch -D` refusing to delete worktree-checked-out branches is accurate, so concurrent gremlins are safe. The PID-collision corner case (where `$$` matches a stale branch) is also strictly improved: previously the immediately-following `git show-ref` check would `die`, now the sweep clears the stale ref and the new run proceeds cleanly.

**Main concern: scope mismatch between PR metadata and diff.** The title (`ghgremlin: sweep stale impl-handoff branches`), description, and `Closes #94` all describe only commit `a786c7b`. But the PR also includes two further commits (`a983e6a` and `54922c3`) that introduce a substantial new concept — `operator_followups` — across `skills/handoff/{SKILL.md,handoff.py}` and `skills/bossgremlin/bossgremlin.py` (~110 lines net). Those changes look sound on their own merits, but they are not what the PR claims to be doing, and a reviewer reading just the description would miss them. Either split the operator-followups commits into their own PR (issue #94 would be a clean ~13-line change), or rewrite the title/description to cover all three commits.

The operator_followups implementation itself is fine: backward-compatible read (`isinstance(raw_followups, list)`), `init_boss_state` always seeds `[]`, the chain-done log path uses `... or []` to handle resume-from-pre-PR-state cleanly, and the carry-forward rule in the handoff prompt is conservative in the right direction (operator tasks won't be silently lost when the agent can't see them in the diff).

PR #97, opened 2026-04-26
rescue: widen Phase A context, add structural verdict

### LINKED ISSUE #89: headless rescue: widen Phase A context, add 'structural' verdict, narrow 'unsalvageable'
## Context

Headless rescue (`gremlins.py rescue --headless <id>`) has produced two confidently-wrong "unsalvageable" verdicts in a single boss chain:

**Failure 1 — boss `extract-pipeline-package-6-phase-chain-61c255` after Phase 0 land:**

> Phase A: agent declared the failure unsalvageable (gh pr merge failed because boss worktree is on detached HEAD and the merge call didn't pass PR #83 explicitly — code-level bug in the landing path, no state.json field controls this.)

The agent's diagnosis was wrong on two counts: (a) `gh pr merge` was being passed `pr_url` explicitly (the bug was missing `cwd=`, not missing PR ref), and (b) the fix had **already landed** in `main` and synced to `~/.claude/skills/gremlins/gremlins.py`. The agent didn't see the fix because it ran with the boss's worktree on the chain's pre-fix base ref.

**Failure 2 — child `phase-1-scaffold-pipeline-package-and-3453cd` after Phase 1 implement bail:**

> Phase A: agent declared the failure unsalvageable (implement stage advanced HEAD instead of leaving uncommitted changes; no state.json edit can fix this — pipeline_args contains only a valid --plan path with nothing to remove)

Correct symptom, wrong scope. The agent searched only `pipeline_args` for a fix — the actual fixes (relax the implement-stage guard in `ghgremlin.sh`, or edit `handoff-002-child.md` to drop operator tasks) are outside its consideration set entirely.

Pattern: **`unsalvageable` is fired whenever the agent's narrow context can't surface a state.json or pipeline-arg change, even when the actual remedy lies in pipeline source code or in a sibling artifact (the child plan).**

## Approach

Two changes, in order.

### 1. Widen Phase A's context

Today the agent runs in the gremlin's *worktree*, which contains:

- the product code at the chain's base ref (often pre-fix)
- no view of `~/.claude/skills/` (the up-to-date pipeline code that's actually executing)
- no easy way to inspect the parent boss's `boss_state.json`, sibling child plans, or the spec

Give the agent a curated view that includes:

- The pipeline source actually executing (`~/.claude/skills/<kind>/*.sh` and `~/.claude/skills/_bg/*.sh`), so it can recognize bugs in the pipeline rather than only in pipeline args.
- For child gremlins of a boss: read-only access to the parent's state dir (`~/.local/state/claude-gremlins/<parent-id>/`) so it can inspect the rolling plan, child plan, and prior children's outcomes.
- The current `git log` / `git diff` from `main` (not the worktree's base ref), so it can see fixes that landed after the chain started.

The agent's prompt should explicitly call out these locations and instruct it to inspect them when forming a verdict.

### 2. Tighten the `unsalvageable` bar

Today `unsalvageable` is the easy escape hatch: any failure the agent can't reduce to "tweak state.json" defaults to unsalvageable. The `transient` and `fixed` markers are underused.

Redefine the verdict space:

- **`fixed`** — the agent edited code or config to address the root cause; pipeline can resume.
- **`transient`** — the agent did nothing, but the failure has an external cause that's already resolved (e.g. flaky network, race, or a fix landed elsewhere). Pipeline relaunches and is expected to succeed.
- **`structural`** (new) — the failure points at a real bug in the pipeline or child plan that the agent recognizes but cannot itself fix in this rescue context (e.g. brittle guard, contract mismatch). Records the diagnosis, files an issue (or annotates state.json with `needs_human: true`), and exits non-zero. Distinct from `unsalvageable` because the chain *can* be salvaged by a human edit; the agent is just not the right actor.
- **`unsalvageable`** (narrowed) — reserved for genuinely unrecoverable states: corrupted state dir, missing worktree, conflicting git state with no clean rewind. Should be rare.

The boss should treat `structural` differently from `unsalvageable`: `structural` halts the chain and surfaces the diagnosis; `unsalvageable` halts and gives up. Both still need operator attention, but the framing differs and `structural` becomes a clear signal to look at the pipeline / plan, not the gremlin's internals.

## Tasks

- [ ] In `gremlins.py`'s headless-rescue Phase A prompt, expand the context the agent is told about: pipeline source paths under `~/.claude/skills/`, parent state-dir for boss children, current `main` ref. Reword the "what you can do" section to include code edits in those locations.
- [ ] Add a `structural` verdict to the marker JSON schema (`{"status": "fixed" | "transient" | "structural" | "unsalvageable", "summary": "..."}`).
- [ ] Update the verdict-handling switch in `gremlins.py`: `structural` writes `bail_reason=structural`, `bail_detail=<summary>`, exits non-zero, and (optionally) creates a GitHub issue describing the structural problem.
- [ ] Update `bossgremlin.py`'s child-failure handling to display `structural` distinctly from `unsalvageable` in the chain log, and document the difference in `bossgremlin/SKILL.md`.
- [ ] Update `gremlins/SKILL.md`'s `rescue --headless` section to document the new verdict.

## Risks

- **Wider context = larger prompts = slower / more expensive Phase A.** Today's 1800s timeout is generous; if Phase A consistently runs the full timeout, the chain stalls. Mitigate by structuring the context as files-on-demand (let the agent read what it needs) rather than dumping everything into the prompt.
- **`structural` becomes the new escape hatch.** If the threshold for `structural` is too low, the agent will use it for ordinary failures it should have fixed itself. Keep the framing strict: `structural` requires the agent to articulate *what* in the pipeline is wrong and *why* a code edit elsewhere is needed.
- **`fixed` verdicts that mutate `~/.claude/` directly.** If the agent fixes a bug in `~/.claude/skills/<kind>/*.sh`, that change is invisible to the user's repo and will be overwritten on the next `sync.sh pull`. Either (a) require fixes land in the repo first, or (b) document that headless `fixed` verdicts may need to be replicated in the source repo. Lean: (a) — `fixed` should mean state-dir / worktree edits only; pipeline-source fixes go through `structural`.

## Open questions

- Should `structural` auto-file an issue, or just annotate state? Auto-filing risks spam from repeated rescues hitting the same structural bug; annotation alone risks the bug being forgotten. Compromise: annotate state, and if the same `bail_detail` substring appears across N gremlins, prompt the operator (via session-summary hook) to file one issue covering all.
- Should this also apply to *interactive* rescue, not just headless? Interactive rescue's Phase A inherits the same narrow worktree context. The verdict-space change is independent of interactivity, but the wider-context change benefits both. Lean: do both, in this issue.

Ref the dead chain `extract-pipeline-package-6-phase-chain-61c255` (two failed Phase A verdicts in one chain) as the concrete repro. Ref the related implement-guard and handoff-filter issues (filed separately).

xbrianh:
## Summary

- Widens headless-rescue Phase A's context: the agent now sees the pipeline source it's running under (`~/.claude/skills/<kind>/*.sh`, `~/.claude/skills/_bg/*.sh`), the parent boss's state dir for boss children (rolling plan, sibling child plans, prior outcomes), and the product repo's `main` ref — structured as files-on-demand so prompt size stays bounded.
- Adds a `structural` verdict to the marker JSON schema for failures the agent identifies as real bugs in pipeline source or sibling child plans but cannot safely fix from the rescue context. `bossgremlin.py` records `outcome=structural:<detail>` and surfaces the agent's diagnosis in the chain-halt message — distinct from `unsalvageable` (genuinely unrecoverable state).
- Narrows `fixed` to state-only edits: direct edits under `~/.claude/skills/` would be invisible to the user's repo and overwritten on the next `sync.sh pull`, so pipeline-source bugs are routed through `structural` instead.

## Test plan

- [ ] Reproduce a child failure in a boss chain where the fix has already landed on `main`; confirm Phase A returns `transient` (or recognizes the upstream fix) instead of `unsalvageable`.
- [ ] Reproduce a child failure where the bug is in `~/.claude/skills/<kind>/*.sh` or in a sibling `handoff-*-child.md`; confirm Phase A returns `structural` and the boss halts with the diagnosis in `boss_state.json` and the log.
- [ ] Confirm a state.json-only flake still resolves as `fixed` and Phase B relaunches normally.
- [ ] Confirm a corrupted-state-dir scenario still resolves as `unsalvageable`.

Closes #89


xbrianh (commented):

PR #99, opened 2026-04-26
ghgremlin: anchor launch worktree at origin/<default-branch>

### LINKED ISSUE #98: ghgremlin: standalone launches should base worktree at origin/<default>, not HEAD
## Context

Standalone `/ghgremlin` launches inherit commits from the user's current branch instead of basing the worktree at `origin/<default-branch>`. Concrete failure mode:

1. User is on a feature branch `my-wip` with N commits not yet merged to `main`.
2. User runs `/ghgremlin <task>` (or `/ghgremlin --plan <ref>`).
3. `_bg/launch.sh:493` runs `git worktree add --detach "$WORKDIR" HEAD`, anchoring the gremlin's worktree at the tip of `my-wip` — including the N un-merged commits.
4. The gremlin implements its task on top of those commits.
5. `commit-pr` opens a PR against `main`. The PR diff includes the gremlin's intended changes **plus** the N un-merged WIP commits. Reviewers see scope creep that has nothing to do with the gremlin's task; CI runs against an unintended base; the user's WIP gets accidentally proposed for merge.

Boss-gh chains already avoid this — `a88cd23` ("Anchor gh bossgremlin chains to the repo default branch") sets the chain base to the repo's default branch explicitly. Standalone `/ghgremlin` launches went unfixed.

`localgremlin` is intentionally different: it's a feature-branch tool that squash-lands onto whatever branch the user is on, so `HEAD` is the right base for it. Scope here is `ghgremlin` only.

## Approach

In `_bg/launch.sh`, special-case `ghgremlin` worktree creation: anchor at `origin/<default-branch>` after a `git fetch`, instead of `HEAD`.

Rough shape:

```bash
if [[ "$KIND" == "ghgremlin" ]]; then
    git -C "$PROJECT_ROOT" fetch origin --quiet \
        || die "git fetch origin failed"
    _default=$(git -C "$PROJECT_ROOT" symbolic-ref refs/remotes/origin/HEAD 2>/dev/null \
        | sed 's@^refs/remotes/origin/@@')
    [[ -n "$_default" ]] || die "could not determine origin's default branch"
    SETUP_KIND="worktree"
    git -C "$PROJECT_ROOT" worktree add --detach "$WORKDIR" "origin/$_default" >/dev/null \
        || die "git worktree add (origin/$_default) failed"
else
    # localgremlin (or future kinds that want HEAD): leave as-is
    ...
fi
```

The default-branch resolution mirrors what `bossgremlin.py` already does for `--chain-kind gh` (look at the `chain_base_ref` resolution in `bossgremlin.py:init_boss_state`-area). Hoist a small helper if both call sites end up identical — but don't refactor more than the bug fix needs.

The existing `git fetch origin` is a network call at launch time, which slightly slows launch on a flaky network. That's the right tradeoff: the alternative (silently basing on a stale `origin/main`) is worse than a slow launch on a bad connection.

## Tasks

- [ ] In `_bg/launch.sh` around line 493, branch on `KIND`: `ghgremlin` resolves `origin/<default>` after a fetch and bases the worktree there. `localgremlin` keeps the current `HEAD` behavior.
- [ ] Consider whether `bossgremlin` itself needs any change here — it already handles its chain-base separately via `bossgremlin.py`, but the *boss's own worktree* is still created via the same `launch.sh` path. Verify whether the boss worktree base matters (the boss doesn't open a PR; it spawns child ghgremlins which create their own worktrees). If it doesn't matter, leave `bossgremlin` on `HEAD`.
- [ ] Update the comment block at `_bg/launch.sh:477-481` to describe the new ghgremlin behavior.
- [ ] Manual smoke: from a feature branch with un-merged commits, run `/ghgremlin <small task>`, confirm the PR diff contains only the gremlin's changes (no WIP from the feature branch).

## Risks

- **`origin/HEAD` not set.** Some clones don't have `refs/remotes/origin/HEAD` configured. Fall back to `git remote show origin | sed -n 's/  HEAD branch: //p'` (slower, but reliable), or fall back to `main` then `master`. Don't silently use `HEAD` — that reintroduces the bug.
- **Network failure at launch.** A `git fetch` failure should `die` rather than fall back to a stale ref. Re-running on a working network is cheap; opening a PR against the wrong base is not.
- **`/ghgremlin --resume` path.** The resume branch (line 76+) reuses the existing worktree, so this fix doesn't affect resume — the worktree was already created at the original launch's base. No change needed there.
- **Feature-branch commits the user *wanted* in the gremlin's diff.** Edge case: someone has stacked WIP they intentionally want as the gremlin's base. Today's behavior accidentally supports that; the fix removes it. Mitigation: that workflow can be served by an explicit flag in a follow-up (e.g. `--base <ref>` on the launcher) — out of scope here. The default should be safe.

## Open questions

- Should the launcher print the resolved base ref so the user can see what the gremlin will diff against? Today the launcher prints `gremlin id`, `workdir`, `log`, `state file`, `pid` — adding `base: origin/main` as one more line is cheap and informative. Lean: yes.
- The PR's `--base` flag in `commit-pr` (`ghgremlin.sh:462-465`-ish) already opens against the default branch. Verify that `commit-pr` doesn't somehow infer base from the worktree's start ref (it shouldn't, but worth a glance during implementation).

Filed against the standalone `/ghgremlin` flow. Boss-gh chains are unaffected — they already anchor correctly via `a88cd23`.

xbrianh:
## Summary

Standalone `/ghgremlin` launches were creating their worktree at `HEAD`, so any un-merged commits on the caller's feature branch leaked into the gremlin's eventual PR diff. This patch special-cases `ghgremlin` in `_bg/launch.sh`: fetch `origin`, resolve the default branch via `git symbolic-ref refs/remotes/origin/HEAD` (with `git remote show origin` as a fallback for clones missing that ref), and base the detached worktree at `origin/<default>`.

Both `git fetch` failure and an unresolved default branch `die` rather than fall back to `HEAD` — opening a PR against the wrong base is more expensive than re-running on a working network.

`localgremlin` keeps its `HEAD` base (it's a feature-branch tool that squash-lands onto whatever branch the user is on). `bossgremlin`'s own worktree also stays at `HEAD`; `gh` chains already pin `chain_base_ref` via `bossgremlin.py`'s `get_default_branch`, and child `ghgremlin`s re-enter `launch.sh` and resolve their own bases.

The launcher now prints a `base:` line when the worktree base was resolved (i.e. for `ghgremlin`) so the caller can see what the gremlin will diff against. `--resume` is unaffected: it reuses the existing worktree, whose base was set at the original launch.

## Test plan

- [ ] From a feature branch with un-merged commits ahead of `main`, run `/ghgremlin <small task>`; confirm the PR diff contains only the gremlin's changes.
- [ ] Confirm `localgremlin` still bases its worktree at `HEAD` (squash-land target unchanged).
- [ ] Confirm `bossgremlin --chain-kind gh` still functions; child `ghgremlin`s anchor at `origin/<default>` independently.
- [ ] Verify the launcher's stdout shows `base: origin/<default>` for `ghgremlin` and omits the line for other kinds.
- [ ] Force a `git fetch` failure (e.g. unreachable remote) and confirm `launch.sh` `die`s rather than silently falling through.

Closes #98

### DIFF
--- a/skills/_bg/launch.sh
+++ b/skills/_bg/launch.sh
@@ -474,12 +474,23 @@ if [[ -n "$_spec_copy_pending" ]]; then
     cp "$_spec_copy_pending" "$STATE_DIR/artifacts/spec.md" || die "could not copy spec to artifacts: $_spec_copy_pending"
 fi
 
-# Isolated workdir setup. For localgremlin in a git repo we create a named
-# branch (bg/localgremlin/<GR_ID>) so the commits the gremlin makes stay
-# reachable after finish.sh runs. For ghgremlin we use --detach because the
-# gremlin's stage 2b creates and pushes its own issue-N-<slug> branch; a
-# named bg/* ref would be a no-op.
+# Isolated workdir setup. Three branches:
+#   localgremlin: named branch bg/localgremlin/<GR_ID> based at HEAD so the
+#     commits the gremlin makes stay reachable after finish.sh runs.
+#     localgremlin is a feature-branch tool and squash-lands onto whatever
+#     branch the user is on, so HEAD is the right base.
+#   ghgremlin: detached worktree based at origin/<default-branch> (after a
+#     fetch). ghgremlin opens a PR against the default branch; basing the
+#     worktree at HEAD would let un-merged commits on the user's current
+#     branch leak into the PR diff. --detach is correct because stage 2b
+#     creates and pushes its own issue-N-<slug> branch — a named bg/* ref
+#     would be a no-op.
+#   bossgremlin: detached worktree at HEAD. The boss itself doesn't open
+#     PRs; it spawns child gremlins that re-enter launch.sh and resolve
+#     their own bases (gh chains additionally pin chain_base_ref via
+#     bossgremlin.py's get_default_branch).
 BRANCH=""
+WORKTREE_BASE_DISPLAY=""
 if [[ $IS_GIT -eq 1 ]]; then
     WORKDIR=$(mktemp -d -t "aibg-$KIND.XXXXXX") || die "mktemp failed"
     rmdir "$WORKDIR" || die "rmdir $WORKDIR failed"
@@ -488,6 +499,28 @@ if [[ $IS_GIT -eq 1 ]]; then
         BRANCH="bg/localgremlin/$GR_ID"
         git -C "$PROJECT_ROOT" worktree add -b "$BRANCH" "$WORKDIR" HEAD >/dev/null \
             || die "git worktree add -b failed"
+    elif [[ "$KIND" == "ghgremlin" ]]; then
+        # Network call at launch time — slower on a flaky connection, but the
+        # alternative (silently basing on a stale origin/<default>) is worse
+        # than a slow launch on a bad connection. die rather than fall back
+        # to HEAD: opening a PR against the wrong base is expensive.
+        git -C "$PROJECT_ROOT" fetch origin --quiet \
+            || die "git fetch origin failed"
+        # symbolic-ref refs/remotes/origin/HEAD is the cheap path. Some
+        # clones lack it (e.g. cloned with --no-checkout, or origin/HEAD
+        # was never set); fall back to `git remote show origin` which
+        # asks the server for the HEAD branch directly.
+        _default=$(git -C "$PROJECT_ROOT" symbolic-ref refs/remotes/origin/HEAD 2>/dev/null \
+            | sed 's@^refs/remotes/origin/@@' || true)
+        if [[ -z "$_default" ]]; then
+            _default=$(git -C "$PROJECT_ROOT" remote show origin 2>/dev/null \
+                | sed -n 's/^  HEAD branch: //p' || true)
+        fi
+        [[ -n "$_default" ]] || die "could not resolve origin's default branch"
+        SETUP_KIND="worktree"
+        WORKTREE_BASE_DISPLAY="origin/$_default"
+        git -C "$PROJECT_ROOT" worktree add --detach "$WORKDIR" "$WORKTREE_BASE_DISPLAY" >/dev/null \
+            || die "git worktree add ($WORKTREE_BASE_DISPLAY) failed"
     else
         SETUP_KIND="worktree"
         git -C "$PROJECT_ROOT" worktree add --detach "$WORKDIR" HEAD >/dev/null \
@@ -574,7 +607,8 @@ fi
 {
     cat <<EOF
 gremlin id:  $GR_ID
-workdir:     $WORKDIR
+workdir:     $WORKDIR${WORKTREE_BASE_DISPLAY:+
+base:        $WORKTREE_BASE_DISPLAY}
 log:         $STATE_DIR/log
 state file:  $STATE_FILE
 pid:         ${PID:-unknown}
--- a/skills/_bg/launch.sh
+++ b/skills/_bg/launch.sh
@@ -88,13 +88,14 @@ if [[ -n "$RESUME_GR_ID" ]]; then
     STATE_FILE="$STATE_DIR/state.json"
     [[ -d "$STATE_DIR" && -f "$STATE_FILE" ]] || die "no state at $STATE_DIR"
 
-    RESUME_KIND=$(jq   -r '.kind         // ""' "$STATE_FILE")
-    WORKDIR=$(jq       -r '.workdir      // ""' "$STATE_FILE")
-    BRANCH=$(jq        -r '.branch       // ""' "$STATE_FILE")
-    STAGE=$(jq         -r '.stage        // ""' "$STATE_FILE")
-    STATUS=$(jq        -r '.status       // ""' "$STATE_FILE")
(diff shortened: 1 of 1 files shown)


xbrianh (commented):

skills/_bg/launch.sh
```diff
@@ -488,6 +499,28 @@ if [[ $IS_GIT -eq 1 ]]; then
...
+        # symbolic-ref refs/remotes/origin/HEAD is the cheap path. Some
+        # clones lack it (e.g. cloned with --no-checkout, or origin/HEAD
+        # was never set); fall back to `git remote show origin` which
+        # asks the server for the HEAD branch directly.
+        _default=$(git -C "$PROJECT_ROOT" symbolic-ref refs/remotes/origin/HEAD 2>/dev/null \
+            | sed 's@^refs/remotes/origin/@@' || true)
+        if [[ -z "$_default" ]]; then
+            _default=$(git -C "$PROJECT_ROOT" remote show origin 2>/dev/null \
+                | sed -n 's/^  HEAD branch: //p' || true)
+        fi
```
xbrianh: Consider using `gh repo view --json defaultBranchRef -q .defaultBranchRef.name` here instead of the git plumbing + remote-show fallback. Rationale:

- ghgremlin already requires `gh` (line 276 of this file), so no new dependency.
- `bossgremlin.py::get_default_branch` already uses exactly this approach — having two different resolution mechanisms in the same repo invites drift.
- It removes the need for the `git remote show origin` fallback (which is a second network round-trip after `fetch`, so it's slow when it fires).
- Strictly speaking, the prior `git fetch origin --quiet` is then unnecessary just for resolving the name — though you'd still want it (or change to `git fetch origin <default> --quiet`) so `worktree add origin/$default` sees up-to-date refs.

Non-blocking; current implementation is correct.

skills/_bg/launch.sh
```diff
@@ -488,6 +499,28 @@ if [[ $IS_GIT -eq 1 ]]; then
...
+        [[ -n "$_default" ]] || die "could not resolve origin's default branch"
+        SETUP_KIND="worktree"
+        WORKTREE_BASE_DISPLAY="origin/$_default"
```
xbrianh, line 521: `WORKTREE_BASE_DISPLAY` is printed at launch time but not persisted to `state.json`. Consider adding it to the jq state-init block around line 562 (e.g. `worktree_base: $worktree_base`). Two reasons:

- `/gremlins` introspection can't currently surface what a ghgremlin's worktree was based on once the launch stdout scrolls away.
- The `--resume` path (line 247-258) also doesn't show the base since it's not in state.json — minor consistency gap.

Minor / nice-to-have.

xbrianh (commented):
Correctness of the fix is solid: ghgremlin's worktree now anchors at `origin/<default-branch>` after a fresh fetch, with a sensible die-rather-than-fall-back posture. The change is well-scoped and the in-line rationale is excellent.

Two non-blocking observations:

1. **Resolution mechanism is duplicated.** `bossgremlin.py::get_default_branch` already resolves the default branch with `gh repo view --json defaultBranchRef -q .defaultBranchRef.name`. ghgremlin already requires `gh` (launch.sh:276). Using `gh` here too would (a) match the bossgremlin code path, (b) avoid the slow `git remote show origin` fallback (separate network round-trip), and (c) make the prior `git fetch` unnecessary for the resolution step itself.

2. **`base:` is launch-time-only.** `WORKTREE_BASE_DISPLAY` is shown on stdout but isn't persisted to `state.json`, so `/gremlins` can't surface it later (and `--resume` can't print it). Persisting `worktree_base` in the jq state-init would close that gap cheaply.

No security or correctness blockers. No tests added — consistent with the rest of this codebase.

skills/_bg/launch.sh
xbrianh: Adopted in c111ae2 — switched to `gh repo view --json defaultBranchRef -q .defaultBranchRef.name`, matching bossgremlin.py's resolution path. Dropped the `git remote show origin` fallback (and its second round-trip).

skills/_bg/launch.sh
xbrianh: Done in c111ae2. Persisted as `worktree_base` in the jq state-init, and the `--resume` stdout block now reads it back and surfaces a matching `base:` line. Surfacing it in `/gremlins` itself is left as a follow-up.

skills/_bg/launch.sh
```diff
@@ -488,6 +499,28 @@ if [[ $IS_GIT -eq 1 ]]; then
...
+        git -C "$PROJECT_ROOT" worktree add --detach "$WORKDIR" "$WORKTREE_BASE_DISPLAY" >/dev/null \
+            || die "git worktree add ($WORKTREE_BASE_DISPLAY) failed"
```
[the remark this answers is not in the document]
xbrianh: Good catch — fixed in c111ae2. Now fetching the default branch with an explicit refspec (`refs/heads/$_default:refs/remotes/origin/$_default`), matching bossgremlin.py's `fetch_origin_branch`. The `die` message also names the branch so a single-branch-clone failure is unambiguous.

PR #100, opened 2026-04-26
ghgremlin: sweep stale impl-handoff branches in implement stage

### LINKED ISSUE #94: ghgremlin: hand-off branch with PID-suffix can leak across resume scenarios

xbrianh:
## Summary

A previous ghgremlin run that died after `git switch -c ghgremlin-impl-handoff-$$` succeeded but before commit-pr's `git branch -m` rename left an orphan branch behind. PID-based naming meant the new run picked an uncolliding name and proceeded — not a correctness bug, but stale branches accumulated in `refs/heads/` across rescues.

After creating the new hand-off branch (which moves HEAD off any stale one), best-effort `git branch -D` every other `ghgremlin-impl-handoff-*` ref. `git branch -D` refuses to delete branches checked out by another worktree, so concurrent gremlins running in sibling worktrees are protected.

## Test plan

- [ ] Manual repro: in a worktree, create a fake `ghgremlin-impl-handoff-12345` ref pointing at HEAD, then drive ghgremlin through the implement stage with a plan that produces commits — confirm the stale branch is gone after the hand-off step.
- [ ] Confirm a concurrent ghgremlin in another worktree is unaffected: its `ghgremlin-impl-handoff-<pid>` is checked out there, so the sweep's `git branch -D` skips it.

Closes #94

### DIFF
--- a/skills/ghgremlin/ghgremlin.sh
+++ b/skills/ghgremlin/ghgremlin.sh
@@ -501,6 +501,18 @@ Implement the plan above by making the code changes in this repo. You may commit
       git branch -f "$PRE_BRANCH" "$PRE_HEAD" \
         || die "could not reset $PRE_BRANCH back to $PRE_HEAD"
     fi
+    # Sweep stale ghgremlin-impl-handoff-* branches from prior failed runs:
+    # a previous run may have died after `git switch -c` but before commit-pr's
+    # `git branch -m` rename, leaving an orphan branch with a different PID
+    # suffix in this repo's refs/heads/. We're now off any such branch (the
+    # switch above moved HEAD to $HANDOFF_BRANCH), so deleting them is safe.
+    # `git branch -D` refuses to delete branches checked out by another
+    # worktree, so a concurrent gremlin's hand-off branch is protected.
+    while IFS= read -r _stale; do
+      [[ -z "$_stale" || "$_stale" == "$HANDOFF_BRANCH" ]] && continue
+      git branch -D "$_stale" >/dev/null 2>&1 || true
+    done < <(git for-each-ref --format='%(refname:short)' \
+                 'refs/heads/ghgremlin-impl-handoff-*')
     _commit_count=$(git rev-list --count "$PRE_HEAD..HEAD")
     echo "    implement committed during run; moved ${_commit_count} commit(s) onto $HANDOFF_BRANCH${PRE_BRANCH:+ and reset $PRE_BRANCH}"
   fi
--- a/skills/ghgremlin/ghgremlin.sh
+++ b/skills/ghgremlin/ghgremlin.sh
@@ -505,12 +505,25 @@ Implement the plan above by making the code changes in this repo. You may commit
     # a previous run may have died after `git switch -c` but before commit-pr's
     # `git branch -m` rename, leaving an orphan branch with a different PID
     # suffix in this repo's refs/heads/. We're now off any such branch (the
-    # switch above moved HEAD to $HANDOFF_BRANCH), so deleting them is safe.
-    # `git branch -D` refuses to delete branches checked out by another
-    # worktree, so a concurrent gremlin's hand-off branch is protected.
+    # switch above moved HEAD to $HANDOFF_BRANCH), so deleting already-merged
+    # ones is safe. `git branch -d` refuses to delete divergent branches, so
+    # unique implementation commits from a prior failed run are preserved
+    # (logged as a warning so an operator can recover them); it also refuses
+    # to delete branches checked out by another worktree, so a concurrent
+    # gremlin's hand-off branch is protected. The explicit "$HANDOFF_BRANCH"
+    # skip below is redundant (git would refuse to delete the currently
+    # checked-out branch anyway) but kept for intent clarity.
     while IFS= read -r _stale; do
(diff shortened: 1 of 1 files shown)


xbrianh (commented):

skills/ghgremlin/ghgremlin.sh
```diff
@@ -501,6 +501,18 @@ Implement the plan above by making the code changes in this repo. You may commit
...
     fi
+    # Sweep stale ghgremlin-impl-handoff-* branches from prior failed runs:
+    # a previous run may have died after `git switch -c` but before commit-pr's
+    # `git branch -m` rename, leaving an orphan branch with a different PID
+    # suffix in this repo's refs/heads/. We're now off any such branch (the
+    # switch above moved HEAD to $HANDOFF_BRANCH), so deleting them is safe.
+    # `git branch -D` refuses to delete branches checked out by another
+    # worktree, so a concurrent gremlin's hand-off branch is protected.
+    while IFS= read -r _stale; do
+      [[ -z "$_stale" || "$_stale" == "$HANDOFF_BRANCH" ]] && continue
```
xbrianh, line 517: Minor: the `"$_stale" == "$HANDOFF_BRANCH"` skip is defensive — `git branch -D` would already refuse to delete the branch currently checked out in this worktree, so this short-circuit is just a faster/quieter path rather than a correctness requirement. Worth keeping (it makes intent obvious), but the comment block above could note that the check is redundant-but-explicit so a future reader doesn't think it's load-bearing.

skills/ghgremlin/ghgremlin.sh
```diff
@@ -501,6 +501,18 @@ Implement the plan above by making the code changes in this repo. You may commit
...
+      git branch -D "$_stale" >/dev/null 2>&1 || true
```
xbrianh: Nit: `>/dev/null 2>&1 || true` swallows the reason a delete fails. The expected refusal case (checked out in another worktree) is intentional, but a genuine FS/permission error would also be silently skipped, leaving the stale branch to accumulate again. Not worth blocking on — the alternative (failing a gremlin run because of cleanup noise) is worse — but if you want a slightly louder failure mode, redirecting stderr to the run log instead of `/dev/null` would surface unexpected git errors without changing control flow.

xbrianh (commented):
Targeted, well-scoped cleanup. The sweep correctly runs *after* the new hand-off branch is checked out (so HEAD is off any stale ref), the for-each-ref pattern is properly quoted to avoid shell glob expansion, and best-effort `branch -D` plus the worktree-checked-out refusal correctly protects concurrent sibling gremlins. No blockers. Two small observations inline.

skills/ghgremlin/ghgremlin.sh
xbrianh: Comment block updated to call out that the `$HANDOFF_BRANCH` skip is redundant-but-explicit (git would refuse to delete the currently checked-out branch anyway), kept for intent clarity.

skills/ghgremlin/ghgremlin.sh
xbrianh: Dropped the `2>&1` so unexpected git errors (FS/permission, etc.) reach stderr instead of being silently swallowed. The worktree-checked-out refusal prints its own diagnostic, and `|| true` still prevents that expected refusal from failing the run.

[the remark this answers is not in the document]
xbrianh: Good catch — applied the ancestor-check approach. Switched to `git branch -d` gated by `git merge-base --is-ancestor "$_stale" HEAD`, so already-merged stale branches are deleted but divergent ones (carrying unique commits from a prior failed run) are left in place with a warning to stderr so an operator can recover them.

PR #101, opened 2026-04-26
rescue: widen Phase A context, add structural verdict

### LINKED ISSUE #89: headless rescue: widen Phase A context, add 'structural' verdict, narrow 'unsalvageable'

xbrianh:
## Summary

- Widen the headless and interactive Phase A rescue prompts so the diagnosis agent inspects the pipeline source actually executing under `~/.claude/skills/`, the product repo at current `main`, and (for boss children) the parent boss's state dir — not just the gremlin's narrow worktree.
- Split the verdict space into four: `fixed`, `transient`, `structural` (new), `unsalvageable` (narrowed). `structural` covers failures that point at a real bug in the pipeline source or a sibling artifact (e.g. a malformed child plan) that the agent recognizes but cannot fix here.
- Scope `fixed` to `state.json` + worktree edits only — pipeline source is mirrored from a source repo and would be clobbered by the next sync, so bugs there must surface as `structural`.
- bossgremlin's chain log now distinguishes `STRUCTURAL` (fix and resume) from `UNSALVAGEABLE` (write off) so an operator skimming a halted chain knows which way to act.

Motivation: in chain `extract-pipeline-package-6-phase-chain-61c255`, headless rescue fired `unsalvageable` twice for failures whose actual remedies lived in pipeline source / a sibling child plan — outside the agent's consideration set entirely.

Closes #89

## Test plan

- [ ] Manually trigger a child failure where the bug lives in `~/.claude/skills/<kind>/*.sh`. Confirm headless Phase A reads the pipeline source, returns `structural`, and that `gremlins.py` writes `bail_reason=structural` with the agent's diagnosis in `bail_detail`.
- [ ] Trigger a child failure under a boss chain. Confirm boss log surfaces `STRUCTURAL` distinctly from `UNSALVAGEABLE` and includes the diagnosis line.
- [ ] Confirm a state-only fix still flows as `fixed` and proceeds to Phase B relaunch.
- [ ] Confirm the four-verdict marker schema (`fixed | transient | structural | unsalvageable`) is documented in `skills/gremlins/SKILL.md` and that `skills/bossgremlin/SKILL.md` documents the operator distinction.


xbrianh (commented):

skills/bossgremlin/bossgremlin.py
```diff
@@ -643,7 +654,32 @@ def main(argv):
...
+                    )
+                    if bail_detail:
+                        log(f"  diagnosis: {bail_detail}")
+                elif bail_reason == "unsalvageable":
+                    log(
+                        f"child {current_child_id} bailed: UNSALVAGEABLE — run "
+                        f"cannot be recovered (giving up)"
+                    )
+                    if bail_detail:
+                        log(f"  detail: {bail_detail}")
```
xbrianh, line 701: The new `STRUCTURAL` / `UNSALVAGEABLE` branches log `bail_detail`, but the `else` branch (which handles every other headless bail reason — `phase_a_timeout`, `phase_a_no_marker`, `phase_a_bad_marker`, `phase_a_claude_error`, `attempts_exhausted`, `excluded_class:*`) still emits only `rescue refused for X (reason)` and drops the agent/wrapper diagnosis. `_write_bail` populates `bail_detail` on every one of those paths, so within the spirit of this PR it'd be cheap to print it here too:

```python
else:
    log(f"rescue refused for {current_child_id} ({bail_reason or 'no bail_reason'})")
    if bail_detail:
        log(f"  detail: {bail_detail}")
```

skills/bossgremlin/bossgremlin.py
```diff
@@ -643,7 +654,32 @@ def main(argv):
...
+                # from with a human edit, but the agent isn't the right actor.
+                # `unsalvageable` means the run is dead and giving up. Surface
+                # the difference in the boss log so an operator skimming a
+                # finished chain can tell whether to look at the pipeline or
+                # write off the run.
+                if bail_reason == "structural":
+                    log(
+                        f"child {current_child_id} bailed: STRUCTURAL — points at a "
+                        f"bug in the pipeline source or a sibling artifact that "
+                        f"needs a human edit elsewhere"
```
xbrianh: The preamble (`points at a bug in the pipeline source or a sibling artifact that needs a human edit elsewhere`) restates what `bail_detail` already says (or should say) — the agent's diagnosis is the actionable part. Consider tightening to one line, e.g.:

```python
log(f"child {current_child_id} bailed: STRUCTURAL — pipeline/sibling-artifact bug, human edit required")
if bail_detail:
    log(f"  diagnosis: {bail_detail}")
```

Otherwise an operator scrolling the boss log gets a generic essay on every structural bail before the actual diagnosis.

skills/bossgremlin/SKILL.md
```diff
@@ -54,6 +54,15 @@ The boss appears in `/gremlins` like any other gremlin (KIND=boss). Its stage co
 - `/gremlins stop <boss-id>` — sends SIGTERM to the boss; it stops the current child (or handoff) and exits. The chain is halted; children that already landed stay landed.
 - `/gremlins rescue <boss-id>` — resumes a dead or stalled boss. The boss re-reads `boss_state.json` and continues from where it stopped.
 
+## When a child bails
+
+If a child fails and headless rescue declines to proceed, the boss halts the chain and records the child's `bail_reason` in `boss_state.json["children"][...]`. Two reasons in particular signal different operator actions:
+
+- **`structural`** — the rescue agent recognized a real bug in the pipeline source (under `~/.claude/skills/`) or in a sibling artifact (e.g. a malformed child plan under the boss's own state dir) that *can* be salvaged, but only by a human editing the pipeline / plan. The boss log surfaces this with a `STRUCTURAL` line plus the agent's diagnosis. Look at the named file, fix it (in the source repo for pipeline changes, then `scripts/sync.sh push`), then `/gremlins rescue <boss-id>` to resume.
```
xbrianh: Two doc nits worth addressing:

1. "...then `/gremlins rescue <boss-id>` to resume" — strictly, the failed child is *not* re-attempted: bossgremlin appends `bailed:structural` to `boss_state["children"]` and sets `current_child_id=None`, so a resumed boss runs a fresh `handoff` rather than retrying the failed phase. In practice that handoff usually produces the same next-plan because the diff hasn't moved, but the wording here implies something more deterministic than what the code does. Suggest: "…then `/gremlins rescue <boss-id>` to resume — the boss restarts at the next handoff (the failed phase will normally re-plan into a fresh child)."

2. `scripts/sync.sh push` is the right ritual *for this repo* but reads as inscrutable to anyone who hasn't internalized the mirroring setup. A parenthetical pointer ("…in the source repo for pipeline changes — for this config repo that's `scripts/sync.sh push` to mirror back into `~/.claude/`") would help readers who land here cold.

skills/gremlins/gremlins.py
```diff
@@ -604,9 +697,11 @@ def _run_headless_phase_a(workdir: str, prompt: str, marker_path: str):
...
     summary = marker.get("summary") or ""
-    if status not in ("fixed", "transient", "unsalvageable"):
+    if status not in ("fixed", "transient", "structural", "unsalvageable"):
         return "bad_marker", f"marker has invalid status: {status!r}"
 
+    if status == "structural":
+        return status, summary or "agent declared failure structural"
     if status == "unsalvageable":
         return status, summary or "agent declared failure unsalvageable"
     return status, ""
```
xbrianh: The agent's `summary` is preserved for `structural`/`unsalvageable` but silently dropped for `fixed`/`transient`. That's the field where the agent typically says what it actually changed (e.g. "removed bad --model flag from pipeline_args") and it never reaches the wrapper's stdout — line 863 just logs `Phase A complete (status: fixed); handing off to Phase B...`. Consider returning the summary in all four cases and including it in the success-path print so the operator (and whoever reads the log post-mortem) can see *what* the diagnosis was, not just that there was one.

skills/gremlins/gremlins.py
```diff
@@ -443,49 +496,87 @@ def build_rescue_prompt(state, log_tail, state_file_path, log_file_path,
...
+   gremlin stays terminal.
 
 Constraints for headless mode:
 - Do NOT prompt for input; there is no TTY.
 - Do NOT call `exit` or otherwise abort — finish normally so the wrapper can read the marker.
 - If you cannot write the marker file, the wrapper will treat that as an unsalvageable failure.
-- Your working directory is a scratch space. The only file outside it you may touch is `state.json` at `{state_file_path}`.
+- Permitted edits: `state.json` at `{state_file_path}`, files inside the
+  worktree at `{workdir or '(unknown)'}`, and your scratch working directory.
+  Pipeline source under `~/.claude/skills/` is read-only for this rescue —
```
xbrianh, line 572: Edge case: when `state.get("workdir")` is falsy this becomes `Permitted edits: state.json at <path>, files inside the worktree at (unknown), and your scratch working directory.` — telling the agent it may write to `(unknown)` is at best confusing, at worst gives them license to write anywhere. The path *should* always be set in practice, but `do_rescue` already guards against missing workdir at line 740-742. Worth either dropping the worktree clause when `workdir` is empty, or asserting it's non-empty here so the prompt is never built with `(unknown)`.

skills/gremlins/gremlins.py
```diff
@@ -443,49 +496,87 @@ def build_rescue_prompt(state, log_tail, state_file_path, log_file_path,
...
+  surface bugs there via `structural`.
```
xbrianh, line 573: Consistency: "Pipeline source under `~/.claude/skills/` is read-only for this rescue — surface bugs there via `structural`." Worth also explicitly listing the marker file at `{marker_path}` as a permitted write; an over-literal agent could read the bullet above and conclude it must not write outside the worktree, then fail to write the marker (which becomes `phase_a_no_marker`). Step 3 already tells them to write it, but the constraint section reads like a hard list.

skills/gremlins/gremlins.py
```diff
@@ -443,49 +496,87 @@ def build_rescue_prompt(state, log_tail, state_file_path, log_file_path,
...
+     (`~/.claude/skills/<kind>/*.sh`, `~/.claude/skills/_bg/*.sh`, or the
+     associated python) or in a sibling artifact (e.g. a malformed child plan
+     under the parent boss's state dir) that you recognize but cannot fix here.
+     Use this when the remedy is "edit the pipeline" or "edit the child plan",
+     not "tweak `state.json`". Articulate *what* in the pipeline / plan is wrong
+     and *why* a fix elsewhere is required — the operator will read your summary.
+   - **unsalvageable**: reserved for genuinely unrecoverable states — corrupted
+     state dir, missing worktree, conflicting git state with no clean rewind.
+     This should be rare. If you are tempted to pick `unsalvageable` because the
+     fix isn't in `state.json`, you almost certainly want `structural` instead.
```
xbrianh: Nit: "If you are tempted to pick `unsalvageable` because the fix isn't in `state.json`, you almost certainly want `structural` instead." — but the section above also tells the agent that worktree edits qualify as `fixed`. The hint here only mentions `state.json`. Worth mentioning worktree too, or making the hint match the broader `fixed` definition: "If the fix isn't a `state.json`/worktree edit you can make, you almost certainly want `structural` instead."

xbrianh (commented):
Solid widening of Phase A context and the four-verdict split is well documented. A few items worth a look before merge:

1. **Boss-log regression for non-structural/non-unsalvageable bails** — the PR adds detail surfacing for two reasons but drops the previous one-liner's `bail_detail` for everything else (`phase_a_*`, `attempts_exhausted`, `excluded_class:*`). Easy to extend.
2. **`bossgremlin/SKILL.md` overstates resume semantics** — the failed child is recorded as bailed and `current_child_id=None`, so resuming runs a fresh handoff rather than retrying the failed phase. Should say so explicitly.
3. **`fixed`/`transient` agent summary is silently discarded** — the agent often explains *what* it changed; that text never reaches the log.
4. Smaller wording / edge-case items inline.

Nothing blocking; mostly polish + a doc nuance.

skills/gremlins/gremlins.py
```diff
@@ -422,11 +422,60 @@ def build_rescue_prompt(state, log_tail, state_file_path, log_file_path,
...
+        f"  worktree. Look here when the failure smells like a code-level issue in",
+        f"  the orchestrator rather than a problem with the gremlin's own state.",
+        "",
+        f"- **Worktree** (the failed gremlin's checkout, may lag behind `main`):",
+        f"    - `{workdir}` — the gremlin runs here at the chain's base ref.",
+        "",
+        f"- **Product repo at current `main`** (catches fixes that landed after the",
+        f"  chain started — the worktree's base ref will not see them):",
+        f"    - `git -C {project_root or '<project_root>'} log -20 main`",
+        f"    - `git -C {project_root or '<project_root>'} diff <base-ref>..main`",
```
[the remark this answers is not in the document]
xbrianh: Fixed: switched to `origin/HEAD` (so it works for repos whose default branch isn't `main`), used `git merge-base HEAD origin/HEAD` to derive the diff range so `<base-ref>` is no longer dangling, and quoted the `-C` path.

skills/gremlins/gremlins.py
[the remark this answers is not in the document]
xbrianh: Fixed: `summary` must now be a string (objects/arrays are rejected as `bad_marker`), and the wrapper collapses internal whitespace and caps length to 500 chars before persisting/printing it. Boss log gets a separate one-line cap on top of that for chain-log readability.

skills/bossgremlin/bossgremlin.py
```diff
@@ -643,7 +654,32 @@ def main(argv):
...
+                    )
+                    if bail_detail:
+                        log(f"  diagnosis: {bail_detail}")
```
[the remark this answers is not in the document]
xbrianh: Fixed: bail_detail now goes through `_summarize_for_log` (collapses whitespace, caps to ~240 chars) before logging. The full untruncated value still lives in state.json.

skills/bossgremlin/bossgremlin.py
xbrianh: Fixed: the else branch now logs `detail: <bail_detail>` when present, matching the structural/unsalvageable arms.

skills/bossgremlin/bossgremlin.py
xbrianh: Adopted the suggestion: STRUCTURAL preamble is now one line (`pipeline/sibling-artifact bug, human edit required`), with the diagnosis on the next line.

skills/bossgremlin/SKILL.md
xbrianh: Both addressed: the `structural` bullet now spells out that resume runs a fresh handoff (failed child marked bailed, `current_child_id` cleared) rather than retrying the failed phase, and `scripts/sync.sh push` now has a parenthetical explaining it's the mirror-back ritual for this config repo.

skills/gremlins/gremlins.py
xbrianh: Fixed: `summary` is now returned for all four verdicts. The `fixed`/`transient` success-path log includes the diagnosis (`Phase A complete (status: fixed, diagnosis: <summary>); ...`) when the agent provided one.

skills/gremlins/gremlins.py
xbrianh: Fixed: when `workdir` is empty the worktree clause is dropped from the constraints line entirely (and from the `Inspect before deciding` section it shows `(no workdir recorded in state — skip this section)` instead of the path), so the agent never sees `(unknown)` as a permitted write target.

skills/gremlins/gremlins.py
xbrianh: Fixed: the constraints bullet now explicitly lists the marker file as a permitted write ("the marker file at <path> (you MUST write this — that's how the wrapper reads your verdict)"), so the read-only-skills rule can't be misread as forbidding the marker write.

skills/gremlins/gremlins.py
xbrianh: Adopted the suggestion: hint now reads "If the fix isn't a `state.json` or worktree edit you can make here, you almost certainly want `structural` instead."

PR #102, opened 2026-04-26
gremlins: gate interactive rescue Phase B on marker file

### LINKED ISSUE #93: interactive rescue: align Phase A → Phase B handoff with headless marker protocol (fix prose/exit-code mismatch)
## Context

Interactive rescue (`gremlins.py rescue <id>`) Phase A can exit 0 — triggering Phase B's automatic resume — while its prose explicitly tells the operator to abort. Phase B then relaunches the gremlin into the same broken state and it re-dies at the same stage.

Concrete repro just now, on `fix-launch-sh-new-launch-exit-code-4111d0`. Phase A's full output:

> ## Diagnosis
>
> The `ghreview` stage failed at line 504 of `/Users/xath/.claude/skills/ghgremlin/ghgremlin.sh` ...
>
> ## Cannot fix via state.json
>
> This is a hardcoded path in `ghgremlin.sh`, not anything driven by `pipeline_args`. ...
>
> **Operator action required:** Ctrl-C during Phase A to prevent the automatic resume, then patch ... Re-resume after that.
>
> I have left `state.json` and all other files untouched.

The agent then exited 0. The wrapper saw exit 0, fired Phase B, the resumed gremlin re-died at `ghreview` immediately.

There are two things wrong, but they share a root: **the agent's prose channel and exit-code channel are independent, and the wrapper trusts only the exit code.** When the agent identifies a structural problem it cannot fix (and recommends operator intervention), it has no protocol for telling the wrapper "do not proceed to Phase B."

The headless variant has this protocol — `{"status": "fixed" | "transient" | "unsalvageable"}` in a marker file. Interactive rescue does not. Its sole signal is the exit code, and the agent's default disposition is to exit 0 unless something blew up.

## Approach

Adopt the same marker-file protocol for interactive rescue. The agent writes `<state-dir>/artifacts/rescue-<ts>.done` with `{"status": ..., "summary": ...}` and the wrapper consults it before deciding whether to invoke Phase B.

Status mapping for interactive (mirrors headless plus the `structural` case proposed in #89):

- **`fixed`** — agent edited code/state to address the root cause; Phase B proceeds.
- **`transient`** — no edit needed, expected to succeed on retry; Phase B proceeds.
- **`structural`** / **`unsalvageable`** — agent cannot fix it from inside this rescue; Phase B does **not** run; wrapper writes `bail_reason` to state.json and exits non-zero. Operator follows the prose guidance.
- **No marker file present** — the agent crashed mid-diagnosis or the prompt didn't reach the marker-write step. Today this is treated as exit 0 = success; under the new protocol, treat as a hard error: do not run Phase B.

The interactive prompt should also be updated to instruct the agent to write the marker before exiting, with the same status vocabulary as headless.

This makes the Phase A → Phase B handoff structured, matches the headless contract, and removes the prose/exit-code mismatch where the agent says "stop" and the wrapper hears "go."

## Tasks

- [ ] Update the interactive rescue agent prompt in `gremlins.py` (the Phase A branch) to require writing the marker JSON before exit, with the same `{"status", "summary"}` schema as headless.
- [ ] Update interactive Phase A's wrapper logic to read the marker after the agent exits, and gate Phase B on `status in {"fixed", "transient"}`. Treat missing/unparseable marker as a hard error (no Phase B), not as success.
- [ ] On `structural` / `unsalvageable` (or missing marker), write `bail_reason` and `bail_detail` to state.json and exit non-zero — same surface area as headless.
- [ ] Document the protocol in `gremlins/SKILL.md`'s `rescue` section. Today the SKILL describes the marker only under `rescue --headless`; lift it to the shared section since both variants now use it.
- [ ] Manual smoke: trigger a rescue against a known-structural failure, confirm the agent's marker is `structural`/`unsalvageable`, confirm Phase B is not invoked, confirm exit code is non-zero and `bail_reason` is set in state.json.

## Risks

- **Agent ignores the marker requirement.** If the agent forgets to write the marker, the wrapper now treats that as a hard error rather than an implicit success. This is the right default — the alternative (silent Phase B on a forgotten marker) is exactly today's bug. The prompt should be explicit and structured enough that any reasonable model follows it.
- **Backward compatibility with mid-rescue interruption (Ctrl-C).** Today, Ctrl-C during Phase A leaves no marker and the wrapper sees a non-zero exit from the interrupted agent — Phase B doesn't run. That path stays the same: interrupted agent → no marker → wrapper exits non-zero, no Phase B.
- **Overlap with #89.** That issue proposes adding `structural` to the verdict space and widening Phase A context. The work here is narrower: aligning interactive's protocol with headless. They can land independently or together; recommend ordering this *after* #89 so the verdict vocabulary is finalized first.

## Open questions

- Should interactive rescue display the marker `summary` to the operator after Phase A, even on `fixed`/`transient`, before kicking off Phase B? Lean: yes — the operator is sitting at the terminal already; a one-line summary of what the agent did is information they want.
- Does interactive rescue actually need a separate marker prompt, or can it reuse the headless prompt verbatim? The original split exists because interactive can be Ctrl-C'd mid-diagnosis. With the marker protocol the prompts converge; reusing one prompt is simpler. Lean: reuse.

Ref the dead `fix-launch-sh-new-launch-exit-code-4111d0` rescue attempt as the concrete repro. Ref #89 (Phase A context + verdict space).

xbrianh:
## Summary

- Interactive rescue Phase A could exit 0 while its prose told the operator to abort, causing Phase B to relaunch the gremlin into the same broken state.
- Adopt the headless marker-file protocol for interactive: agent writes `<state-dir>/artifacts/rescue-<ts>.done` with `{status, summary}`; wrapper gates Phase B on `status in {fixed, transient}`.
- `unsalvageable` / missing marker / unparseable marker / non-zero claude exit now write `bail_reason` + `bail_detail` to `state.json` and skip Phase B (matches headless surface area). KeyboardInterrupt still preserves state with no bail.
- Drop the `headless` branch from `build_rescue_prompt` — both modes share one prompt. Lift the marker contract in `SKILL.md` from the `--headless` bullet up to the shared `rescue` bullet.

## Test plan

- [x] `python -m py_compile skills/gremlins/gremlins.py`
- [x] Functional checks of `_read_rescue_marker` (no_marker / bad_marker / fixed / unsalvageable / invalid status) and `build_rescue_prompt` (marker contract + status vocabulary present)
- [ ] Manual: trigger a rescue against a known-structural failure; confirm marker is `unsalvageable`, Phase B is not invoked, exit non-zero, `bail_reason` set in state.json

Closes #93


xbrianh (commented):

Plan tasks 1–4 land cleanly in the diff: the interactive prompt is unified with the headless one (per the plan's open-question lean toward reuse), `do_rescue` now reads the marker after interactive Phase A and gates Phase B on `status ∈ {fixed, transient}`, all failure paths (`unsalvageable`, `no_marker`, `bad_marker`, `phase_a_claude_error`) write `bail_reason`/`bail_detail` and exit non-zero, and `SKILL.md` lifts the marker contract to the shared `rescue` section. The `structural` verdict is intentionally deferred per the plan's #89 overlap note. Task 5 is manual smoke, not in-diff. The shared `_read_rescue_marker` extraction is in service of de-duping the marker logic across both modes — not a hitchhiker.

skills/gremlins/gremlins.py
```diff
@@ -760,23 +766,50 @@ def do_rescue(target: str, headless: bool = False) -> bool:
             # actually relaunches.
             print(f"Phase A complete (status: {status}); handing off to Phase B...")
         else:
-            print("Phase A: running diagnosis agent inline — Ctrl-C to abort.")
+            print(f"Phase A: running diagnosis agent inline — Ctrl-C to abort.")
```
xbrianh: Stray `f` prefix — this string has no interpolation. Drop the `f` to match the original line (and avoid a lint warning under `ruff`/`pyflakes` `F541`):

```python
print("Phase A: running diagnosis agent inline — Ctrl-C to abort.")
```

skills/gremlins/gremlins.py
```diff
@@ -543,6 +534,37 @@ def _write_bail(sf: str, wdir: str, bail_reason: str, bail_detail: str = "") ->
...
+    if not isinstance(marker, dict):
+        return "bad_marker", "marker file is not a JSON object"
+
+    status = marker.get("status")
+    summary = marker.get("summary") or ""
+    if status not in ("fixed", "transient", "unsalvageable"):
+        return "bad_marker", f"marker has invalid status: {status!r}"
+
+    if status == "unsalvageable":
+        return status, summary or "agent declared failure unsalvageable"
```
xbrianh: When the agent omits `summary` on `unsalvageable`, this default (`"agent declared failure unsalvageable"`) ends up double-printed at the call site:

```
Phase A: agent declared the failure unsalvageable (agent declared failure unsalvageable)
```

(see line 799 / the headless `print` path) Either return an empty `msg` here and let the caller decide on a fallback, or have the call sites suppress parenthesized detail when it's only the placeholder.

skills/gremlins/gremlins.py
```diff
@@ -590,26 +612,12 @@ def _run_headless_phase_a(workdir: str, prompt: str, marker_path: str):
...
-        return "bad_marker", f"marker has invalid status: {status!r}"
-
-    if status == "unsalvageable":
-        return status, summary or "agent declared failure unsalvageable"
-    return status, ""
+    status, msg = _read_rescue_marker(marker_path)
+    # Headless callers don't surface the summary on success, so blank it out
+    # to preserve the original two-tuple contract (msg empty for fixed/transient).
+    if status in ("fixed", "transient"):
+        return status, ""
```
xbrianh: This blank-the-summary dance only exists because `_run_headless_phase_a`'s historical contract was "empty msg on success." Now that `_read_rescue_marker` is the shared helper, it's leaky to have one caller (interactive) trust the summary and the other (headless) discard it. Two options to consider:

1. Drop the special case entirely — let headless return `(status, summary)` too, and update the headless caller in `do_rescue` (lines 744–767) to use the summary if you want it logged on success.
2. Or pull the blanking into `_read_rescue_marker` itself — but then interactive loses the summary, which seems worse.

Not blocking; just an observation that the abstraction is slightly leaky as it stands.

skills/gremlins/gremlins.py
```diff
@@ -694,18 +702,17 @@ def do_rescue(target: str, headless: bool = False) -> bool:
...
-        except OSError:
-            pass
-        ts = datetime.datetime.utcnow().strftime("%Y%m%dT%H%M%SZ")
-        marker_path = os.path.join(artifacts_dir, f"rescue-{ts}.done")
+    try:
+        os.makedirs(artifacts_dir, exist_ok=True)
+    except OSError:
+        pass
+    ts = datetime.datetime.utcnow().strftime("%Y%m%dT%H%M%SZ")
+    marker_path = os.path.join(artifacts_dir, f"rescue-{ts}.done")
```
xbrianh: Second-resolution timestamp + no pre-cleanup creates a small footgun: if a rescue is retried within the same UTC second (e.g. operator hits Ctrl-C, immediately reruns `/gremlins rescue`), the new run computes the same `marker_path` and a stale marker from the prior attempt would be picked up by `_read_rescue_marker` as if the new agent wrote it.

Cheap fixes: `os.unlink(marker_path)` (best-effort) right before launching `claude -p`, or bump resolution to `%Y%m%dT%H%M%S_%f` / append PID. Edge-case but real, and the new bail-on-stale-marker behavior makes the consequence noisier than it used to be.

skills/gremlins/gremlins.py
```diff
@@ -760,23 +766,50 @@ def do_rescue(target: str, headless: bool = False) -> bool:
...
                 print(f"Inspect the log at {log_path} and worktree at {workdir} for details.")
                 return False
+
+            # Zero exit: the marker is the source of truth. Without it the
+            # agent is presumed to have abdicated the protocol — bail rather
+            # than silently launching Phase B into the same broken state.
+            status, msg = _read_rescue_marker(marker_path)
+            if status == "unsalvageable":
+                _write_bail(sf, wdir, "unsalvageable", msg)
+                print(f"Phase A: agent declared the failure unsalvageable ({msg})")
```
xbrianh: Behavior change worth flagging in the PR description (or a follow-up): the interactive path now writes `bail_reason=unsalvageable` and touches `finished` on this branch, where previously the operator could just rerun `rescue` after thinking about it.

`do_rescue`'s preflight only refuses `running`, `dead:finished`, or `stalled:*`, so `dead:bailed:unsalvageable` is still re-rescuable — but the operator will see a scarier liveness in `/gremlins` between attempts and may not realize they can retry. Consider a one-line print here pointing at `rescue` as a valid next step (mirroring the `state preserved — rerun /gremlins rescue` line on the KeyboardInterrupt path).

xbrianh (commented):
Solid unification of the marker protocol — interactive now has the same agent→wrapper contract as headless, and the bail vocabulary is consistent across both. The core change (gating Phase B on a successful marker read) is the right fix for the original bug. A few minor cleanup notes below; none are blockers.

skills/gremlins/gremlins.py
xbrianh: Fixed in 30c9ed5 — dropped the stray `f` prefix.

skills/gremlins/gremlins.py
xbrianh: Fixed in 30c9ed5 — `_read_rescue_marker` now returns an empty msg on unsalvageable when the agent omits `summary`, and both call sites suppress the parenthesized detail in that case (printing the bare line instead).

skills/gremlins/gremlins.py
xbrianh: Fixed in 30c9ed5 — `marker_path` now includes microseconds and PID (`%Y%m%dT%H%M%S_%fZ-<pid>`), plus a best-effort `os.unlink` before launching `claude -p`. Two retries in the same UTC second can no longer read each other's stale markers.

skills/gremlins/gremlins.py
xbrianh: Fixed in 30c9ed5 — interactive unsalvageable path now prints "Gremlin marked dead:bailed:unsalvageable — rerun /gremlins rescue if you want to try again." so the operator sees the retry path even though liveness looks terminal.

skills/gremlins/gremlins.py
```diff
@@ -471,23 +472,13 @@ def build_rescue_prompt(state, log_tail, state_file_path, log_file_path,
 
 5. Stop. Do NOT re-run the failed stage or any remaining stages yourself — the wrapper reads the marker and (on `fixed`/`transient`) hands off to a background resume that relaunches the pipeline starting at {stage}. On `unsalvageable`, the wrapper writes a bail reason and the gremlin stays terminal.
 
-Constraints for headless mode:
+Constraints:
 - Do NOT prompt for input; there is no TTY.
 - Do NOT call `exit` or otherwise abort — finish normally so the wrapper can read the marker.
-- If you cannot write the marker file, the wrapper will treat that as an unsalvageable failure.
+- If you do not write the marker file, the wrapper will treat that as a hard error and bail with `phase_a_no_marker`.
 - Your working directory is a scratch space. The only file outside it you may touch is `state.json` at `{state_file_path}`.
```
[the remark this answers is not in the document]
xbrianh: Fixed in 30c9ed5 — Constraints now explicitly list both `state.json` and the marker path as allowed targets outside the scratch dir, and the "no TTY" sentence is reframed ("Headless runs have no TTY at all; interactive runs share the operator's terminal but the agent must still complete autonomously") so it matches reality in interactive mode.

skills/gremlins/gremlins.py
```diff
@@ -543,6 +534,37 @@ def _write_bail(sf: str, wdir: str, bail_reason: str, bail_detail: str = "") ->
...
+        return "bad_marker", "marker file is not a JSON object"
+
+    status = marker.get("status")
+    summary = marker.get("summary") or ""
+    if status not in ("fixed", "transient", "unsalvageable"):
+        return "bad_marker", f"marker has invalid status: {status!r}"
+
+    if status == "unsalvageable":
+        return status, summary or "agent declared failure unsalvageable"
+    return status, summary
```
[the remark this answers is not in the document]
xbrianh: Fixed in 30c9ed5 — `_read_rescue_marker` now type-checks `summary`: `None` becomes "", `str` is passed through, and any other type returns `bad_marker` with a wrapper-generated `bail_detail` so the contract stays string-typed end-to-end.

skills/gremlins/gremlins.py
```diff
@@ -694,18 +702,17 @@ def do_rescue(target: str, headless: bool = False) -> bool:
...
-        try:
-            os.makedirs(artifacts_dir, exist_ok=True)
-        except OSError:
-            pass
-        ts = datetime.datetime.utcnow().strftime("%Y%m%dT%H%M%SZ")
-        marker_path = os.path.join(artifacts_dir, f"rescue-{ts}.done")
+    try:
+        os.makedirs(artifacts_dir, exist_ok=True)
+    except OSError:
+        pass
```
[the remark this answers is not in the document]
xbrianh: Fixed in 30c9ed5 along the lines of your suggestion — captures the OSError, then validates `isdir` + `os.access(W_OK | X_OK)` and bails with `wrapper_artifacts_dir_unavailable` (writing `bail_reason`/`bail_detail` only in headless, since interactive doesn't bail-write on the preflight failures it already gates). Phase A is no longer reached when the artifacts dir can't be prepared.

skills/gremlins/SKILL.md
```diff
@@ -37,14 +37,16 @@ The script produces a small table. Each row is one gremlin:
...
-- `rescue --headless <id>`: same two-phase flow, run end-to-end with no TTY and no user input — for unattended callers (humans walking away, future bossgremlin). Differences from interactive:
+- `rescue <id>` (interactive, default): two-phase diagnose-then-resume for a dead or stalled gremlin. **Phase A (foreground):** reads the failure log and existing artifacts, then spawns a `claude -p` agent **inline** in the gremlin's worktree to diagnose and fix the underlying issue. The agent's prompt requires it to write a marker file at `<state-dir>/artifacts/rescue-<ts>.done` containing `{"status": "fixed" | "transient" | "unsalvageable", "summary": "..."}` before exiting — this is the agent → wrapper handoff signal, used by both interactive and headless rescue. Ctrl-C to abort — state is preserved, no marker is required, no handoff happens. **Phase B (background):** the wrapper reads the marker after Phase A and acts on its `status`:
```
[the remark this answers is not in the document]
xbrianh: Fixed in 30c9ed5 — applied your suggested wording almost verbatim ("in a separate scratch working directory" + the "do not assume" caveat). Thanks.

skills/gremlins/gremlins.py
xbrianh: Acknowledged — leaving as-is for this PR. The headless caller has been intentionally narrowed to the original `(status, '')` two-tuple so its existing match-on-empty-msg path stays stable; option 1 (drop the special case, let headless surface the summary on success) is a real improvement but it's a refactor of the headless logging path that's out of scope here. Won't file a follow-up unless this gets noisier in practice.

PR #106, opened 2026-04-26
pipeline: Phase 0 DESIGN.md inventorying gremlin sources

xbrianh:
## Summary
- Adds `pipeline/DESIGN.md` as the Phase 0 deliverable for the gremlin-pipeline migration. No code moves; this is the binding spec for Phases 1–5.
- Inventoried `skills/localgremlin/{localgremlin.py,_core.py}`, `skills/ghgremlin/ghgremlin.sh`, `skills/bossgremlin/bossgremlin.py`, and the `skills/_bg/` shell helpers to document the shared surface.

## What the doc covers
- **Module layout** — refined from the strawman with explicit per-stage prompt files, `clients/` (real + fake), `stages/`, `orchestrators/`, `prompts/lenses/`, and a `tests/` tree.
- **`ClaudeClient` interface** — exact `run(...)` signature derived from the six `claude -p` call sites, supporting stream-json, text, `--resume <session>`, parallel invocation, and capture-events for URL extraction.
- **`launch.sh --resume` contract** — before/after spawn shape, `PYTHONPATH=$HOME/.claude` env var, kind→subcommand mapping, byte-stable resume stage names, per-phase rollout.
- **Rescue marker-protocol surface** — vocabulary tables for `bail_class` (4 upstream values), agent verdict `status` (4 values: `fixed`/`transient`/`structural`/`unsalvageable`), and `bail_reason` (8+ values written by `gremlins.py`). Every name flagged byte-stable across the migration.
- **`impl-handoff` branch lifecycle** — pre/post HEAD capture, four-way outcome classification, `git switch -c`, PRE_BRANCH reset, stale-branch sweep, commit-pr prompt selection, mapped to Phase 3 modules.

## Open question resolved
- `pipeline/` lives at the repo root (parallel to `skills/`); Phase 5 extends `scripts/sync.sh` `DIR_PAIRS` so it mirrors to `~/.claude/pipeline/`.

Closes #105

## Test plan
- [ ] Skim `pipeline/DESIGN.md` end-to-end and confirm the inventoried call-sites match (line numbers cited inline).
- [ ] Confirm the `bail_class` vs `bail_reason` field distinction is correctly described against `liveness.sh:50-57` and `bossgremlin.py:410-418`.
- [ ] Confirm the `launch.sh` before/after dispatch shape matches the existing `_bg/launch.sh:597-598` and `:230-240` blocks.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #108, opened 2026-04-26
gremlins: fix gh-land halting boss chain on detached-HEAD cwd

### LINKED ISSUE #107: gremlins: fix gh PR land when boss worktree is on detached HEAD
## Context

Boss-driven `gremlins land <gh-child>` halts the chain when the boss's worktree is on detached HEAD, even though the child PR successfully merged on GitHub.

Concrete failure observed in this repo on 2026-04-26 (boss `extract-pipeline-package-via-phased-d04390`, child `phase-0-inventory-gremlin-sources-and-3b9c15`, PR #106):

- Boss worktree: `/private/var/folders/.../aibg-bossgremlin.XXXXXX.0qeIOmeJAT` (detached HEAD at b06a4ad).
- `_land_gh` (`skills/gremlins/gremlins.py:1569`) runs `gh pr merge <pr_url> --squash --delete-branch` with `cwd = state.project_root`. For a boss-launched child, the child's `project_root` is the boss's worktree (detached HEAD), so `gh` fails on the post-merge `--delete-branch` step which needs a current branch to switch off the deleted branch.
- gh exits non-zero with `error: could not determine current branch: failed to run git: not on any branch`.
- The PR DID merge on GitHub (mergedAt 2026-04-26T20:10:32Z, mergeCommit 49d2d72) — but `_land_gh` returned False, and `bossgremlin.py:653` halted the chain with "child pipeline succeeded but land failed".

## Approach

Two related fixes in `skills/gremlins/gremlins.py`:

1. **Resolve a non-detached cwd before invoking gh.** Add a small helper `_resolve_landing_cwd(state) -> str` near `_fast_forward_main`. When the gremlin has a `parent_id`, walk the chain up to the topmost ancestor and use that ancestor's `project_root` (the user's actual repo, on a real branch). Fall back to `state.project_root` if the chain can't be resolved. Use this resolved cwd for both `gh pr merge` and the subsequent `_fast_forward_main` call.

2. **Re-verify PR state on `gh pr merge` non-zero exit.** Even after fix #1, gh may fail on transient post-merge cleanup. After a non-zero exit, run `gh pr view <url> --json state` and treat `MERGED` as success (log a warning). Defense-in-depth so a gh cleanup hiccup never rolls back a successfully-merged PR into a chain halt.

Why these two together: fix #1 is the structural root cause; fix #2 hardens against any other post-merge failure mode (different gh versions, transient errors, branch-protection edge cases). Together they make the gh-land path robust across the boss/non-boss split without changing behavior for non-boss callers (where `parent_id` is empty and `state.project_root` is already the right cwd).

## Tasks

- [ ] Add `_resolve_landing_cwd(state: dict) -> str` near `_fast_forward_main` (around `skills/gremlins/gremlins.py:1079`). Walk `state.parent_id` via `os.path.join(STATE_ROOT, parent_id, 'state.json')` and `load_state`, with cycle protection (`seen` set) and a fallback to `state.project_root`.
- [ ] In `_land_gh` (`skills/gremlins/gremlins.py:1569`), replace the existing `project_root = state.get("project_root") or ""` line with a call to the new helper. Keep the `cwd = project_root if project_root and os.path.isdir(project_root) else None` guard.
- [ ] After the `gh pr merge` call (around line 1653), on non-zero exit, run `gh pr view <pr_url> --json state` and short-circuit to the success path (log warning, call `_fast_forward_main`, `_cleanup_gremlin`, `return True`) when `state == "MERGED"`. Preserve the existing `"already merged"` substring check as a faster pre-check.
- [ ] Add a one-line code comment at the cwd-resolution site explaining the detached-HEAD failure mode and citing gh's `could not determine current branch` error so future readers can grep for it.
- [ ] Verify behavior is unchanged for non-boss-launched gh gremlins (empty `parent_id`): the helper returns `state.project_root` verbatim.

## Open questions

- Should the helper be reused by `_land_local` and `_land_boss` for symmetry, or is the detached-HEAD failure unique enough to gh that we keep the helper local to `_land_gh`? (Recommendation: keep it local for now — local/boss landing currently runs from a clean cwd; revisit if a similar failure surfaces.)
- Is there a case where the child's `project_root` legitimately differs from the boss's (e.g. a future cross-repo workflow)? Today the answer is no, but the helper's fallback-to-self behavior leaves room for that.

## Notes for the implementer

- Repo syncs `skills/` to `~/.claude/skills/` via `scripts/sync.sh`. The operator will run `scripts/sync.sh push` after merge; the gremlin does NOT need to.
- Do NOT touch the bossgremlin orchestrator or `launch.sh --resume` path.
- The existing `gh pr view` call shape at `skills/gremlins/gremlins.py:1581` is the right template for the post-merge re-verification.

xbrianh:
## Summary
- Resolve a non-detached cwd before invoking `gh pr merge` in `_land_gh`: new `_resolve_landing_cwd` helper walks `parent_id` up to the topmost ancestor's `project_root` (the user's real repo on a real branch), with cycle protection and a fallback to the gremlin's own `project_root`. Non-boss callers (empty `parent_id`) get `state.project_root` verbatim, preserving existing behavior.
- Re-verify PR state on non-zero `gh pr merge` exit via `gh pr view --json state`, and treat `MERGED` as success (warning) so post-merge cleanup hiccups never roll back a merged PR into a chain halt. Existing `"already merged"` short-circuit preserved.

## Why
Boss-driven `gremlins land <gh-child>` was halting the chain when the boss's worktree (used as the child's `project_root`) was on detached HEAD. `gh pr merge --squash --delete-branch` succeeded at merging on GitHub but failed on the post-merge branch-switch step (`could not determine current branch: failed to run git: not on any branch`), so `_land_gh` returned False and the boss reported "child pipeline succeeded but land failed" — even though PR #106 had actually merged.

## Test plan
- [ ] Non-boss `gh` gremlin lands as before (empty `parent_id` → helper returns `state.project_root` verbatim).
- [ ] Boss-launched child gh gremlin lands cleanly when the boss's worktree is on detached HEAD.
- [ ] If `gh pr merge` exits non-zero but `gh pr view` reports `MERGED`, landing proceeds with a warning and cleans up.
- [ ] If `gh pr merge` exits non-zero and the PR is not merged, landing still fails as before.

Closes #107

### DIFF
--- a/skills/gremlins/gremlins.py
+++ b/skills/gremlins/gremlins.py
@@ -1076,6 +1076,36 @@ def expected_branch(state: dict, gr_id: str):
     return None
 
 
+def _resolve_landing_cwd(state: dict) -> str:
+    """Return a project_root suitable as cwd for `gh pr merge --delete-branch`.
+
+    For boss-launched children, state.project_root is the boss's worktree, which
+    is on a detached HEAD. After --delete-branch, gh tries to switch off the
+    deleted branch and fails with "could not determine current branch: failed
+    to run git: not on any branch". Walk parent_id up to the topmost ancestor
+    (the user's actual repo, on a real branch) to avoid that.
+    """
+    own_root = state.get("project_root") or ""
+    parent_id = state.get("parent_id") or ""
+    if not parent_id:
+        return own_root
+
+    seen = set()
+    current = state
+    while True:
+        pid = current.get("parent_id") or ""
+        if not pid or pid in seen:
+            break
+        seen.add(pid)
+        parent_sf = os.path.join(STATE_ROOT, pid, "state.json")
+        parent_state = load_state(parent_sf)
+        if not parent_state:
+            break
+        current = parent_state
+
+    return current.get("project_root") or own_root
+
+
 def _fast_forward_main(cwd):
     """Attempt to fast-forward local main to origin/main after a gh PR merge."""
     r = subprocess.run(["git", "fetch", "origin"], capture_output=True, text=True, cwd=cwd)
@@ -1574,7 +1604,7 @@ def _land_gh(gr_id: str, sf: str, wdir: str, state: dict, force: bool = False) -
         print("This gremlin may have been launched before pr_url tracking was added to ghgremlin.sh.")
         return False
 
-    project_root = state.get("project_root") or ""
+    project_root = _resolve_landing_cwd(state)
     cwd = project_root if project_root and os.path.isdir(project_root) else None
 
     print(f"Checking PR: {pr_url}")
@@ -1654,8 +1684,25 @@ def _land_gh(gr_id: str, sf: str, wdir: str, state: dict, force: bool = False) -
         if "already merged" in r.stdout.lower() or "already merged" in r.stderr.lower():
             print("PR was already merged.")
         else:
-            print(f"error: gh pr merge failed: {r.stderr.strip() or r.stdout.strip()}")
-            return False
+            # gh may exit non-zero on post-merge cleanup (e.g. --delete-branch
+            # tries to switch off the deleted branch and fails on a detached
+            # HEAD cwd) even though the PR did merge. Re-verify before bailing.
+            err = r.stderr.strip() or r.stdout.strip()
+            v = subprocess.run(
+                ["gh", "pr", "view", pr_url, "--json", "state"],
+                capture_output=True, text=True, cwd=cwd,
+            )
+            verified_merged = False
+            if v.returncode == 0:
+                try:
+                    verified_merged = json.loads(v.stdout).get("state") == "MERGED"
+                except json.JSONDecodeError:
+                    pass
+            if verified_merged:
+                print(f"warning: gh pr merge exited non-zero ({err}) but PR is MERGED on GitHub — proceeding with cleanup.")
+            else:
+                print(f"error: gh pr merge failed: {err}")
+                return False
     else:
         print("PR merged.")
 
--- a/skills/gremlins/gremlins.py
+++ b/skills/gremlins/gremlins.py
@@ -1090,20 +1090,30 @@ def _resolve_landing_cwd(state: dict) -> str:
     if not parent_id:
         return own_root
 
-    seen = set()
+    # Pre-seed cycle protection with the starting state's id so a pathological
+    # cycle that loops back through the starting gremlin trips on first revisit.
+    seen = {state.get("id") or ""}
     current = state
     while True:
         pid = current.get("parent_id") or ""
-        if not pid or pid in seen:
-            break
+        if not pid:
+            # Clean termination: reached the topmost ancestor. Note: if its
+            # project_root is empty/missing (e.g. corrupted boss state.json),
+            # the own_root fallback may still be detached — strictly no worse
+            # than the original failure mode.
+            return current.get("project_root") or own_root
+        if pid in seen:
+            # Cycle in parent chain — fall back to own_root rather than
+            # returning a possibly-detached intermediate ancestor.
+            return own_root
         seen.add(pid)
         parent_sf = os.path.join(STATE_ROOT, pid, "state.json")
         parent_state = load_state(parent_sf)
         if not parent_state:
-            break
+            # Unreadable parent state — fall back to own_root rather than
+            # returning a possibly-detached intermediate ancestor.
+            return own_root
         current = parent_state
-
-    return current.get("project_root") or own_root
 
 
 def _fast_forward_main(cwd):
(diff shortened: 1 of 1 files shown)


xbrianh (commented):

skills/gremlins/gremlins.py
```diff
@@ -1076,6 +1076,36 @@ def expected_branch(state: dict, gr_id: str):
...
+    if not parent_id:
+        return own_root
+
+    seen = set()
+    current = state
+    while True:
+        pid = current.get("parent_id") or ""
+        if not pid or pid in seen:
+            break
+        seen.add(pid)
```
xbrianh, line 1109: Minor: cycle protection doesn't pre-seed `seen` with the original state's id, so a pathological cycle that loops back through the starting gremlin (state.id=A, parent_id=B, B.parent_id=A) takes one extra iteration before tripping. It still terminates safely, so this is purely a clarity/defensiveness nit — `seen.add(state.get("id") or "")` before the loop would make the invariant crisper. Feel free to ignore.

skills/gremlins/gremlins.py
```diff
@@ -1654,8 +1684,25 @@ def _land_gh(gr_id: str, sf: str, wdir: str, state: dict, force: bool = False) -
...
+            verified_merged = False
+            if v.returncode == 0:
+                try:
+                    verified_merged = json.loads(v.stdout).get("state") == "MERGED"
+                except json.JSONDecodeError:
+                    pass
+            if verified_merged:
+                print(f"warning: gh pr merge exited non-zero ({err}) but PR is MERGED on GitHub — proceeding with cleanup.")
+            else:
+                print(f"error: gh pr merge failed: {err}")
```
xbrianh: When `gh pr view` itself fails on the re-verify (e.g. transient network) after a `gh pr merge` non-zero exit, this prints `error: gh pr merge failed: <err>` even though we don't actually know whether the merge succeeded. Not worth changing the control flow over (conservatively bailing is right), but the message could note that verification was inconclusive so the operator knows to check the PR state manually before reaching for `rescue` or re-running `land`. Optional polish.

skills/gremlins/gremlins.py
```diff
@@ -1076,6 +1076,36 @@ def expected_branch(state: dict, gr_id: str):
...
+        if not pid or pid in seen:
+            break
+        seen.add(pid)
+        parent_sf = os.path.join(STATE_ROOT, pid, "state.json")
+        parent_state = load_state(parent_sf)
+        if not parent_state:
+            break
+        current = parent_state
+
+    return current.get("project_root") or own_root
```
xbrianh: Worth noting for future readers: if the topmost ancestor's `project_root` is itself empty/missing (e.g. a corrupted boss state.json), the `or own_root` fallback returns the child's detached-HEAD project_root — i.e. we're back to the original failure mode. That's strictly no worse than before and is the right fallback choice, but a one-line comment here ("fallback may still be detached if parent state is unrecoverable") would save the next reader the head-scratch.

xbrianh (commented):
Targeted, low-risk fix. The two-layer approach is sound: walking up `parent_id` to a non-detached cwd addresses the structural root cause, and the post-merge re-verification via `gh pr view --json state` is good defense-in-depth that won't mask real merge failures (the only way `verified_merged` becomes true is if GitHub itself reports `MERGED`, which can't happen for a conflicted/blocked PR). Behavior preservation for non-boss callers (empty `parent_id` short-circuits to `state.project_root`) is explicit and correct.

A couple of minor observations inline; nothing blocking.

skills/gremlins/gremlins.py
```diff
@@ -1076,6 +1076,36 @@ def expected_branch(state: dict, gr_id: str):
...
+
+
```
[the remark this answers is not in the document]
xbrianh: Fixed in 9e41142. The function now returns own_root explicitly on cycle detection and on unreadable parent state, and only returns the topmost ancestor's project_root on clean termination (empty parent_id). This matches the docstring's stated fallback behavior.

skills/gremlins/gremlins.py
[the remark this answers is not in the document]
xbrianh: Fixed in 9e41142. The verify path now captures stderr (or the JSONDecodeError) into verify_err and surfaces it in the failure message when verification is inconclusive.

skills/gremlins/gremlins.py
xbrianh: Done in 9e41142 — seen now starts with state.get('id') so a same-gremlin cycle trips on first revisit.

skills/gremlins/gremlins.py
xbrianh: Addressed in 9e41142 alongside Copilot's related suggestion: the failure message now reads 'gh pr merge failed: <err>; verification inconclusive: <verify_err>' when gh pr view fails or returns unparseable output, with an inline comment noting the operator should check PR state manually before reaching for rescue or re-running land.

skills/gremlins/gremlins.py
xbrianh: Added in 9e41142 — the clean-termination return now carries an inline note that a corrupted topmost-ancestor project_root falls through to own_root and may still be detached, but is strictly no worse than the original failure mode.

PR #114, opened 2026-04-26
pipeline: Phase 2 — test suite against the new seam

### LINKED ISSUE #111: pipeline: add Phase 2 test suite against the new seam
# pipeline Phase 2: tests against the new seam

## Context

The `pipeline/` package skeleton was created in Phase 1 and the localgremlin orchestrator fully migrated. `pipeline/clients/fake.py` already contains a working `FakeClaudeClient`. `pipeline/tests/` has only `__init__.py` and `conftest.py` (which adds the repo root to `sys.path` so `import pipeline` resolves without installing). No test files exist yet. This phase adds the test suite — no production-code changes.

Key existing interfaces to test against:

- `pipeline/runner.py`: `run_stages(stages, *, resume_from=None)` and `install_signal_handlers(client)`.
- `pipeline/clients/fake.py`: `FakeClaudeClient(fixtures={label: events_list_or_path}, text_results={label: str})`, `.calls: list[RecordedCall]`.
- `pipeline/stages/plan.py`: `run_plan_stage(*, client, plan_model, plan_file, instructions, raw_path)` — raises `RuntimeError` if `plan_file` doesn't exist or is empty after `client.run`.
- `pipeline/stages/implement.py`: `run_implement_stage(...)` — raises on empty diff/no changes.
- `pipeline/stages/review_code.py`: triple-lens fan-out using `FakeClaudeClient`.
- `pipeline/stages/address_code.py`: `run_address_code_stage(...)`.
- `pipeline/orchestrators/local.py`: `local_main(argv)`, `review_main(argv)`, `address_main(argv)` — all accept a `client=` kwarg for injection.

## Approach

Write three test files and one documentation file. All tests use `FakeClaudeClient`; none spawn real `claude -p` subprocesses. Use `tmp_path` (pytest) for temporary directories. Inline event fixtures as Python lists (no fixture files on disk needed unless a stage explicitly reads a `.jsonl` from disk, in which case write the fixture bytes via `FakeClaudeClient`'s `raw_path` write-back).

Read the actual stage and orchestrator source before writing each test file so the assertions match the real interfaces. The conftest already wires `sys.path`.

## Tasks

- [ ] Task 1: Write `pipeline/tests/test_runner.py`. Cover: `run_stages` executes all stages in order; `run_stages` with `resume_from` skips stages before the named stage; `run_stages` raises `ValueError` on unknown `resume_from` name; `run_stages` stops at first exception raised by a stage callable; `install_signal_handlers` wires SIGINT/SIGTERM to call `client.reap_all()` (test by sending `signal.SIGINT` to the current process with a FakeClaudeClient that records `reap_all` calls, catching `SystemExit`).

- [ ] Task 2: Write `pipeline/tests/test_stages_local.py`. Cover at minimum: `run_plan_stage` raises `RuntimeError` when the plan file is absent after `client.run` (FakeClaudeClient returns without creating the file); `run_plan_stage` succeeds when the test pre-creates the plan file (simulating what the real claude would write); `run_implement_stage` raises when the worktree produces no diff (test with a real temp git repo initialized with a first commit, no further changes); the address-code stage reads review files from the session dir and calls `client.run` with a prompt that includes their content. Read the actual stage source first to confirm exact raise conditions and call shapes before writing assertions.

- [ ] Task 3: Write `pipeline/tests/test_orchestrator_local.py`. Cover: smoke test of `local_main` in `--plan <path>` mode (skips plan stage, runs implement → review-code → address-code) using a `FakeClaudeClient` injected via the `client=` kwarg; verify the FakeClaudeClient's `.calls` list contains entries for the expected labels in the expected order; verify `review_main` and `address_main` each call `client.run` with the expected label. Use `tmp_path`, a minimal temp git repo (initialized with `git init` + first commit so the worktree is not dirty), and inline event fixtures.

- [ ] Task 4: Write `pipeline/README.md` documenting how to run the tests: `cd pipeline && PYTHONPATH=. python -m pytest` (or equivalently, `PYTHONPATH=$(pwd)/.. python -m pytest` from inside `pipeline/`). Keep it brief — one or two sentences is enough.

- [ ] Task 5: Verify `cd pipeline && PYTHONPATH=. python -m pytest` exits 0 before committing. Fix any failures found before opening the PR.

## Open questions

- The implement stage's empty-diff check depends on `git status --porcelain` in the worktree. Tests that exercise this path need a real temp git repo. If setting up a real git repo in tests turns out to require more scaffolding than expected, it is acceptable to focus the implement-stage test on the `run` call recording (label, model forwarded) and defer the empty-diff raise to a follow-up, as long as the three test files exist and `pytest` passes.

xbrianh:
## Summary

- Adds `pipeline/tests/test_runner.py` (5 tests): `run_stages` ordering and resume semantics, signal handler wiring
- Adds `pipeline/tests/test_stages_local.py` (4 tests): plan/implement/address-code stages via `FakeClaudeClient`
- Adds `pipeline/tests/test_orchestrator_local.py` (3 tests): `local_main`/`review_main`/`address_main` smoke tests with injected fake client
- Adds `pipeline/README.md` with one-line test-run instructions
- All 12 tests pass; no production code changes

## Test plan

- [x] `cd pipeline && PYTHONPATH=. python -m pytest` exits 0 (12 passed in 0.29s)

Closes #111

### DIFF
--- a/pipeline/tests/test_orchestrator_local.py
+++ b/pipeline/tests/test_orchestrator_local.py
@@ -3,13 +3,9 @@ import re
 
 import pytest
 
+from conftest import MINIMAL_EVENTS
 from pipeline.clients.fake import FakeClaudeClient
 from pipeline.orchestrators.local import address_main, local_main, review_main
-
-MINIMAL_EVENTS = [
-    {"type": "system", "subtype": "init", "session_id": "test-session-1"},
-    {"type": "result", "subtype": "success"},
-]
 
 # Labels the triple review emits (default sonnet models).
 _REVIEW_LABELS = {
@@ -27,14 +23,14 @@ class _ReviewCreatingClient(FakeClaudeClient):
     def run(self, prompt, *, label, **kwargs):
         if label.startswith("review-code:"):
             m = re.search(r"`([^`]+\.md)`\s+is the canonical", prompt)
-            if m:
-                out = pathlib.Path(m.group(1))
-                out.parent.mkdir(parents=True, exist_ok=True)
-                out.write_text("# Review\n\n## Findings\nNone.\n")
+            assert m, f"regex did not match review-code prompt for label {label!r}"
+            out = pathlib.Path(m.group(1))
+            out.parent.mkdir(parents=True, exist_ok=True)
+            out.write_text("# Review\n\n## Findings\nNone.\n")
         return super().run(prompt, label=label, **kwargs)
 
 
-def _common_patches(monkeypatch, session_dir):
+def _common_patches(monkeypatch):
     """Apply monkeypatches shared across orchestrator smoke tests."""
     import shutil
     monkeypatch.setattr(shutil, "which", lambda n: "/fake/claude" if n == "claude" else None)
@@ -57,7 +53,7 @@ def test_local_main_plan_mode(tmp_path, monkeypatch):
     plan_file.write_text("# Plan\nDo stuff.\n")
 
     monkeypatch.chdir(tmp_path)
-    _common_patches(monkeypatch, session_dir)
+    _common_patches(monkeypatch)
     monkeypatch.setattr(
         "pipeline.orchestrators.local.resolve_session_dir", lambda: session_dir
     )
@@ -93,7 +89,7 @@ def test_local_main_plan_mode(tmp_path, monkeypatch):
 # ---------------------------------------------------------------------------
 
 def test_review_main_calls_client(tmp_path, monkeypatch):
-    _common_patches(monkeypatch, tmp_path)
+    _common_patches(monkeypatch)
     monkeypatch.setattr("pipeline.orchestrators.local.in_git_repo", lambda: False)
 
     client = _ReviewCreatingClient(
@@ -115,7 +111,7 @@ def test_address_main_calls_client(tmp_path, monkeypatch):
             f"# {lens.title()} Review\n\n## Findings\nNone.\n"
         )
 
-    _common_patches(monkeypatch, tmp_path)
+    _common_patches(monkeypatch)
     monkeypatch.setattr("pipeline.orchestrators.local.in_git_repo", lambda: False)
 
     client = FakeClaudeClient(fixtures={"address-code": MINIMAL_EVENTS})
--- a/pipeline/tests/test_orchestrator_local.py
+++ b/pipeline/tests/test_orchestrator_local.py
@@ -1,0 +1,127 @@
+import pathlib
+import re
+
+import pytest
+
+from pipeline.clients.fake import FakeClaudeClient
+from pipeline.orchestrators.local import address_main, local_main, review_main
+
+MINIMAL_EVENTS = [
+    {"type": "system", "subtype": "init", "session_id": "test-session-1"},
+    {"type": "result", "subtype": "success"},
+]
+
+# Labels the triple review emits (default sonnet models).
+_REVIEW_LABELS = {
+    "review-code:holistic:sonnet",
+    "review-code:detail:sonnet",
+    "review-code:scope:sonnet",
+}
+
+
+class _ReviewCreatingClient(FakeClaudeClient):
+    """FakeClaudeClient that writes the review output file when a review-code
+    label is called. Extracts the output path from the prompt so it lands at
+    exactly the path run_triple_review expects to exist after the workers finish."""
+
+    def run(self, prompt, *, label, **kwargs):
+        if label.startswith("review-code:"):
+            m = re.search(r"`([^`]+\.md)`\s+is the canonical", prompt)
+            if m:
+                out = pathlib.Path(m.group(1))
+                out.parent.mkdir(parents=True, exist_ok=True)
+                out.write_text("# Review\n\n## Findings\nNone.\n")
+        return super().run(prompt, label=label, **kwargs)
+
+
+def _common_patches(monkeypatch, session_dir):
+    """Apply monkeypatches shared across orchestrator smoke tests."""
+    import shutil
+    monkeypatch.setattr(shutil, "which", lambda n: "/fake/claude" if n == "claude" else None)
+    monkeypatch.setattr(
+        "pipeline.orchestrators.local.install_signal_handlers", lambda c: None
+    )
+    monkeypatch.setattr(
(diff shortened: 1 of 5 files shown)


xbrianh (commented):

pipeline/tests/test_orchestrator_local.py
```diff
@@ -0,0 +1,127 @@
...
+        if label.startswith("review-code:"):
+            m = re.search(r"`([^`]+\.md)`\s+is the canonical", prompt)
+            if m:
+                out = pathlib.Path(m.group(1))
+                out.parent.mkdir(parents=True, exist_ok=True)
+                out.write_text("# Review\n\n## Findings\nNone.\n")
+        return super().run(prompt, label=label, **kwargs)
+
+
+def _common_patches(monkeypatch, session_dir):
```
xbrianh: `session_dir` is accepted but never used inside the function — all three call-sites pass it but it has no effect. Either drop the parameter or use it.

```python
def _common_patches(monkeypatch):  # remove the dead arg
```

pipeline/tests/test_orchestrator_local.py
```diff
@@ -0,0 +1,127 @@
...
+
+class _ReviewCreatingClient(FakeClaudeClient):
+    """FakeClaudeClient that writes the review output file when a review-code
+    label is called. Extracts the output path from the prompt so it lands at
+    exactly the path run_triple_review expects to exist after the workers finish."""
+
+    def run(self, prompt, *, label, **kwargs):
+        if label.startswith("review-code:"):
+            m = re.search(r"`([^`]+\.md)`\s+is the canonical", prompt)
+            if m:
```
xbrianh: Silent failure risk: when `m` is `None` (i.e., the prompt template changes and the pattern stops matching), the `if m:` block is skipped, no file is written, and the test fails later with a confusing "review file not found" error rather than pointing here. Consider asserting the match is non-None:

```python
if label.startswith("review-code:"):
    m = re.search(r"`([^`]+\.md)`\s+is the canonical", prompt)
    assert m, f"regex did not match review-code prompt for label {label!r}"
    out = pathlib.Path(m.group(1))
    out.parent.mkdir(parents=True, exist_ok=True)
    out.write_text("# Review\n\n## Findings\nNone.\n")
```

pipeline/tests/test_runner.py
```diff
@@ -0,0 +1,75 @@
...
+        ("a", lambda: log.append("a")),
+        ("b", failing),
+        ("c", lambda: log.append("c")),
+    ]
+    with pytest.raises(RuntimeError, match="boom"):
+        run_stages(stages)
+    assert log == ["a"]
+
+
+def test_install_signal_handlers_calls_reap_on_sigint():
```
xbrianh, line 62: `install_signal_handlers` wires up both SIGINT and SIGTERM, but only SIGINT is exercised here. A matching test for SIGTERM (reap_calls increments and exit code 130) would complete the coverage. As-is, a regression in the SIGTERM path would go undetected.

pipeline/tests/test_stages_local.py
```diff
@@ -0,0 +1,135 @@
...
+import subprocess
+
+import pytest
+
+from pipeline.clients.fake import FakeClaudeClient
+from pipeline.stages.address_code import run_address_code_stage
+from pipeline.stages.implement import run_implement_stage
+from pipeline.stages.plan import run_plan_stage
+
+MINIMAL_EVENTS = [
```
xbrianh: `MINIMAL_EVENTS` is defined identically in both this file and `test_orchestrator_local.py`. Move it to a `conftest.py` at the `tests/` level so both modules import it from one place.

xbrianh (commented):
Solid test suite — good coverage of ordering, resume semantics, signal handling, and the stage contracts. Four small issues worth fixing before merge.

pipeline/tests/test_orchestrator_local.py
xbrianh: Fixed — dropped the  parameter from  and updated all three call sites.

pipeline/tests/test_orchestrator_local.py
xbrianh: Fixed — replaced `if m:` with `assert m, f"regex did not match ..."` so a prompt-template drift fails loudly at the right place.

pipeline/tests/test_runner.py
xbrianh: Added `test_install_signal_handlers_calls_reap_on_sigterm` alongside the existing SIGINT test. Both now invoke the registered handler directly rather than sending a real OS signal.

pipeline/tests/test_stages_local.py
xbrianh: Done — moved MINIMAL_EVENTS to conftest.py; both test modules now import it from there.

pipeline/tests/test_runner.py
```diff
@@ -0,0 +1,75 @@
...
+    old_sigterm = signal.getsignal(signal.SIGTERM)
+    try:
+        install_signal_handlers(client)
+        with pytest.raises(SystemExit) as exc_info:
+            os.kill(os.getpid(), signal.SIGINT)
+        assert exc_info.value.code == 130
+        assert client.reap_calls == 1
+    finally:
+        signal.signal(signal.SIGINT, old_sigint)
+        signal.signal(signal.SIGTERM, old_sigterm)
```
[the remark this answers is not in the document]
xbrianh: Fixed — both SIGINT and SIGTERM tests now fetch the installed handler via `signal.getsignal` and invoke it directly, avoiding the real OS signal entirely.

pipeline/tests/test_orchestrator_local.py
```diff
@@ -0,0 +1,127 @@
...
+                out = pathlib.Path(m.group(1))
+                out.parent.mkdir(parents=True, exist_ok=True)
+                out.write_text("# Review\n\n## Findings\nNone.\n")
+        return super().run(prompt, label=label, **kwargs)
```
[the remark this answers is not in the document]
xbrianh: The regex is the only path-derivation approach available inside `run()` without threading `session_dir` through the client interface — the label encodes lens+model but not the session directory. The silent-failure risk is now addressed by asserting the match (see the related comment fix), so a prompt-template change will fail loudly at this line rather than downstream.

PR #120, opened 2026-04-27
gremlins: persist a Markdown report on every rescue attempt

### LINKED ISSUE #112: gremlins: write rescue report to state dir on each rescue attempt
## Context

When \`/gremlins rescue\` is invoked (interactive or headless), the outcome of the diagnosis step and the relaunch step is logged to stdout but not persisted anywhere. After the session ends, there is no record of what the diagnosis agent concluded, whether a relaunch was attempted, or what the outcome was. This makes post-mortems on frequently-rescued gremlins difficult and leaves the \`rescue_count\` counter as the only evidence that rescues occurred.

## Approach

After each rescue attempt concludes — success or failure, interactive or headless — write a structured Markdown report into the gremlin's state directory. The report captures everything an operator would want for a post-mortem without requiring them to reconstruct the sequence from memory or logs.

Reports use a timestamp-qualified filename (\`rescue-<UTC-ts>.md\`) so multiple attempts accumulate without overwriting. The drill-in view (\`/gremlins <id>\`) already lists all files under \`<state-dir>/artifacts/\` and the state dir itself, so reports land somewhere the operator is already pointed at.

Putting the file in the state dir root (not \`artifacts/\`) keeps it alongside \`log\`, \`state.json\`, and the \`finished\`/\`closed\` markers — the operational record layer — rather than mixed with plan/review outputs in \`artifacts/\`.

## Tasks

- [ ] Add a \`write_rescue_report(wdir, report: dict) -> None\` helper in \`gremlins.py\` that:
  - Derives the filename as \`rescue-<YYYYmmddTHHMMSSZ>.md\` (UTC, same format as the existing marker timestamp)
  - Writes a Markdown report with these sections: **Rescue Attempt** (timestamp, attempt number = rescue\_count + 1 at time of call), **Gremlin Context** (id, kind, description, failed stage, model if present in state), **Diagnosis** (verdict/status, summary from marker or bail reason, headless vs interactive), **Relaunch** (attempted yes/no, outcome: success / failed / skipped)
  - Best-effort: catches and swallows all exceptions (same "never break a session" principle as the rest of the file)

- [ ] Call \`write_rescue_report\` at the end of \`do_rescue\` covering all terminal paths:
  - Headless bail paths (excluded class, attempts exhausted, diagnosis failure modes, structural/unsalvageable verdicts) — report relaunch as "skipped"
  - Interactive bail paths (non-zero claude exit, structural, unsalvageable, no/bad marker) — report relaunch as "skipped"
  - Relaunch launcher missing / relaunch failed — report relaunch as "failed" with reason
  - Relaunch successful — report relaunch as "success"
  - Interactive Ctrl-C abort: skip the report (state is preserved with no change, rescue\_count not incremented, consistent with the current "no bail" behaviour)

- [ ] Update \`do_drill_in\` to list rescue reports (files matching \`rescue-*.md\` in \`wdir\`) under a dedicated **rescue reports** section, separate from the artifacts listing, so they're immediately visible when inspecting a gremlin

- [ ] Update \`SKILL.md\` to mention that rescue reports are written to the state directory on each attempt and are visible in the drill-in view

## Open questions

- Should rescue reports also be written inside \`artifacts/\` (where they'd be preserved after \`land\` but pruned with \`rm\`) rather than the state dir root? The state dir is removed by \`rm\`, so reports disappear on cleanup — which seems fine for operational records. If long-term auditability is needed this could be revisited.
- The \`rescue_count\` in state.json is incremented by \`launch.sh --resume\` (the relaunch step), not by \`do_rescue\` itself. So at the moment \`write_rescue_report\` is called, \`rescue_count\` reflects the count *before* this attempt. The report should note attempt N = rescue\_count + 1 and make clear the count is pre-increment.

xbrianh:
## Summary

- Add `write_rescue_report(wdir, report)` (best-effort, never raises) that writes `<state-dir>/rescue-<UTC-ts>.md` with sections **Rescue Attempt** / **Gremlin Context** / **Diagnosis** / **Relaunch**. Attempt number = `rescue_count + 1`, annotated as pre-increment since `launch.sh --resume` is what actually bumps the counter.
- Wrap `do_rescue`'s body in a `try`/`finally` keyed on a mutable `report` dict and an `aborted` flag. Every terminal path (headless excluded-class / attempts-exhausted / Phase A timeout / claude_exit / no_marker / bad_marker / structural / unsalvageable, the interactive equivalents, the wrapper artifacts-dir failure, Phase B launcher missing / exec failure / non-zero exit, success) populates `verdict`, `summary`, `relaunch_outcome`, `relaunch_reason`. Ctrl-C aborts skip the report so state stays exactly as before the attempt.
- Update `do_drill_in` to scan `<wdir>` for `rescue-*.md` and print them under their own `rescue reports:` block, separate from `artifacts:`.
- SKILL.md: add a paragraph after the headless rescue's "Bail visibility" note explaining the new report file, accumulation, and drill-in surfacing.

Reports live in the state dir root (alongside `log` / `state.json` / `finished`) — operational record layer, not the plan/review artifacts layer. They are removed by `rm` along with the rest of the state dir; long-term auditability could revisit this later.

Closes #112

## Test plan

- [x] `python -m py_compile skills/gremlins/gremlins.py`
- [x] Smoke: `write_rescue_report` against a tempdir produces the expected Markdown with all sections + parent boss id when present + outcome/reason fields.
- [x] Smoke: `write_rescue_report('/nonexistent/path', {})` is a silent no-op (best-effort guarantee).
- [x] Smoke: `do_drill_in` against a state dir containing both `artifacts/plan.md` and `rescue-*.md` files lists them under separate `artifacts:` and `rescue reports:` headings.
- [ ] Manual: trigger a real rescue (interactive + headless) end-to-end and confirm the resulting report file matches the printed log.


xbrianh (commented):

skills/gremlins/gremlins.py
```diff
@@ -632,6 +632,84 @@ def _write_bail(sf: str, wdir: str, bail_reason: str, bail_detail: str = "") ->
...
+      - attempt_number: int (rescue_count + 1; pre-increment, since
+        launch.sh --resume bumps the counter on relaunch — not do_rescue)
+      - headless: bool
+      - verdict: str (agent verdict, Phase A failure mode, or wrapper refusal)
+      - summary: str (agent summary or wrapper-generated diagnostic)
+      - relaunch_outcome: str ('success' / 'failed' / 'skipped')
+      - relaunch_reason: str (optional explanation for failed/skipped paths)
+    """
+    try:
+        ts = datetime.datetime.utcnow().strftime("%Y%m%dT%H%M%SZ")
```
xbrianh: Filename uses 1-second resolution (`%Y%m%dT%H%M%SZ`), but two rescue attempts in the same UTC second silently overwrite each other — breaking the PR description's stated guarantee that "the timestamp keeps each filename unique" and that "Multiple attempts accumulate."

This is a realistic case for the wrapper-refusal paths (`excluded_class`, `attempts_exhausted`, `wrapper_artifacts_dir_unavailable`), which return in well under a second — a scripted/back-to-back invocation, or an operator-driven retry loop, would lose history.

The `marker_path` code 300 lines down uses `%Y%m%dT%H%M%S_%fZ` + `os.getpid()` for the same uniqueness concern (line 958). Suggest matching that pattern here:

```python
ts = datetime.datetime.utcnow().strftime("%Y%m%dT%H%M%S_%fZ")
path = os.path.join(wdir, f"rescue-{ts}-{os.getpid()}.md")
```

(The `do_drill_in` glob `startswith("rescue-") and endswith(".md")` will continue to pick them up, and lex sort still equals chronological sort.)

skills/gremlins/gremlins.py
```diff
@@ -632,6 +632,84 @@ def _write_bail(sf: str, wdir: str, bail_reason: str, bail_detail: str = "") ->
...
+        path = os.path.join(wdir, f"rescue-{ts}.md")
+        state = report.get("state") or {}
+        gr_id = state.get("id") or os.path.basename(wdir)
+        kind = state.get("kind") or ""
+        description = state.get("description") or ""
+        stage = state.get("stage") or ""
+        model = state.get("model")
+        parent_id = state.get("parent_id") or ""
+        attempt = int(report.get("attempt_number") or 0)
+        prior = max(attempt - 1, 0)
```
xbrianh: Nit: `max(attempt - 1, 0)` clamp is dead defense — `rescue_count` is normalized to a non-negative int at line 850-854, so `attempt_number = rescue_count + 1 >= 1` and `attempt - 1 >= 0` is guaranteed. Plain `prior = attempt - 1` would be clearer.

xbrianh (commented):
Clean refactor: the `try`/`finally` keyed on a mutable `report` dict + `aborted` flag is a tidy way to thread report-emission through every terminal path in `do_rescue`, and the Ctrl-C carve-out matches the existing "preserve state on operator abort" semantics. Smoke tests cover the happy paths and the silent no-op contract.

One real correctness concern: the report filename uses 1-second resolution and can silently overwrite a prior report when two rescue attempts land in the same UTC second — directly contradicting the PR description's "Multiple attempts accumulate (the timestamp keeps each filename unique)" guarantee. The neighboring `marker_path` code already uses microseconds + PID for exactly this reason. Plus a small style nit on the `prior` clamp.

skills/gremlins/gremlins.py
```diff
@@ -632,6 +632,84 @@ def _write_bail(sf: str, wdir: str, bail_reason: str, bail_detail: str = "") ->
...
+        headless = bool(report.get("headless"))
+        verdict = report.get("verdict") or ""
+        summary = report.get("summary") or ""
+        relaunch_outcome = report.get("relaunch_outcome") or ""
+        relaunch_reason = report.get("relaunch_reason") or ""
+        ts_human = datetime.datetime.utcnow().strftime("%Y-%m-%dT%H:%M:%SZ")
+
+        attempted = "yes" if relaunch_outcome in ("success", "failed") else "no"
```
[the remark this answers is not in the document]
xbrianh: Addressed in 136bd98 — added an explicit `relaunch_attempted` boolean to the report dict, defaulted to False and flipped to True only right before `subprocess.run` is invoked. The two preflight-refusal paths (launcher not executable, FileNotFoundError-on-exec... actually wait — FileNotFoundError is raised *during* the run call, so I treat that as 'attempted'; only the `os.access` preflight short-circuit leaves it False) now correctly produce `Attempted: no` while still surfacing `Outcome: failed` and the diagnostic reason. `write_rescue_report` also falls back to outcome-inference when the key is absent, for older callers.

skills/gremlins/gremlins.py
```diff
@@ -632,6 +632,84 @@ def _write_bail(sf: str, wdir: str, bail_reason: str, bail_detail: str = "") ->
...
+      - headless: bool
+      - verdict: str (agent verdict, Phase A failure mode, or wrapper refusal)
+      - summary: str (agent summary or wrapper-generated diagnostic)
+      - relaunch_outcome: str ('success' / 'failed' / 'skipped')
+      - relaunch_reason: str (optional explanation for failed/skipped paths)
+    """
+    try:
+        ts = datetime.datetime.utcnow().strftime("%Y%m%dT%H%M%SZ")
+        path = os.path.join(wdir, f"rescue-{ts}.md")
+        state = report.get("state") or {}
```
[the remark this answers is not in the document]
xbrianh: Addressed in 136bd98 — switched the filename to `rescue-%Y%m%dT%H%M%S_%fZ-<pid>.md`, matching the marker_path uniqueness pattern ~290 lines below. SKILL.md updated. Verified with a smoke test that three back-to-back writes within the same UTC second produce three distinct files.

skills/gremlins/gremlins.py
xbrianh: Addressed in 136bd98 — adopted the suggested `%Y%m%dT%H%M%S_%fZ` + `os.getpid()` pattern verbatim. Updated SKILL.md to reflect the new filename pattern. The `do_drill_in` glob (`startswith("rescue-") and endswith(".md")`) continues to pick them up, and lex-sort still equals chronological sort.

skills/gremlins/gremlins.py
xbrianh: Addressed in 136bd98 — replaced with `prior = attempt - 1` and added a comment naming the upstream invariant (rescue_count normalized to non-negative int).

PR #124, opened 2026-04-27
bossgremlin: accept GitHub issue refs for --plan

### LINKED ISSUE #82: bossgremlin: accept issue reference in --plan (snapshot at launch)
## Context

`ghgremlin --plan` accepts either a local file path or a GitHub issue reference (`42`, `#42`, `owner/repo#42`, or full URL). `bossgremlin --plan` and `localgremlin --plan` only accept a local file path — `_bg/launch.sh:364-371` fails fast on anything else.

This is annoying for the common boss-gremlin flow: write the spec as a GitHub issue (so it's reviewable, linkable, persistent), then launch a chain against it. Today that requires a manual `gh issue view <n> --json body -q .body > /tmp/spec.md` step before every launch.

The original argument against issue refs was that the boss spec must be immutable for the lifetime of the chain. That's solvable by snapshotting the issue body once at launch (the same way `ghgremlin.sh` already does into `$ARTIFACTS_DIR/plan.md`), then treating the snapshot as the canonical spec for every subsequent handoff.

Scope: `bossgremlin` only. `localgremlin` is out of scope here — it has no natural "post the spec as an issue first" workflow, and adding issue refs there is a separable decision.

## Approach

1. **Hoist the issue-ref parsing out of `ghgremlin.sh`.** The regex block at `ghgremlin.sh:228-265` (handles `42` / `#42` / `owner/repo#42` / `https://github.com/.../issues/42`) is the only copy in the repo. Move it into a small shared helper (e.g. `skills/_bg/resolve-plan.sh` sourced by both gremlins, or a Python equivalent if we'd rather not touch bash). `ghgremlin.sh` keeps its current behavior by sourcing the helper.
2. **Update `_bg/launch.sh:364-371`** so `bossgremlin` is no longer in the "file path required" branch. Issue-ref validation moves to the resume-from-state-dir path: launch records the raw `--plan` arg in `state.json` and defers fetching to `bossgremlin.py`.
3. **In `bossgremlin.py`,** before the first handoff, resolve `--plan` to a local snapshot under `~/.local/state/claude-gremlins/<boss-id>/spec.md` (mirroring `ghgremlin.sh`'s `$ARTIFACTS_DIR/plan.md`). Use `gh issue view` for issue refs; `cp` for file paths. Persist `issue_url` / `issue_num` into `boss_state.json` so they appear in status output and survive resume.
4. **Update the handoff invocation** to pass the snapshot path as `--spec`, not the original arg. The handoff agent already takes a `--spec` (added in #68) — this just changes what file it points at.
5. **Resume semantics.** On `/gremlins rescue`, the snapshot is authoritative; do not re-fetch the issue (matches `ghgremlin.sh:191-199`). Document this in `bossgremlin/SKILL.md`.
6. **Update `bossgremlin/SKILL.md`** `--plan` description: mirror `ghgremlin/SKILL.md`'s wording for the dual file-or-issue-ref form, including the cross-repo `owner/repo#42` shape.

## Tasks

- [ ] Extract the plan-source resolver from `ghgremlin.sh:184-265` into a shared helper. Both `ghgremlin.sh` and `bossgremlin.py` (or a thin shell wrapper around it) consume the same logic.
- [ ] Loosen `_bg/launch.sh:364-371` so `bossgremlin` is no longer required to point at an existing file. Keep the `localgremlin` constraint untouched.
- [ ] In `bossgremlin.py`, resolve `--plan` once at chain start: snapshot to `<state-dir>/spec.md`, persist `issue_url` / `issue_num` into `boss_state.json`. The handoff agent receives the snapshot path.
- [ ] On rescue, treat the snapshot as authoritative — never re-fetch.
- [ ] Update `skills/bossgremlin/SKILL.md`'s `--plan` blurb to document the file-or-issue-ref form.
- [ ] Update `argument-hint` frontmatter in `bossgremlin/SKILL.md` from `<spec-path>` to `<spec-path|issue-ref>`.
- [ ] Add a smoke test path: launch `bossgremlin --plan <issue-ref>` and verify `spec.md` lands in the state dir with the issue body.

## Risks

- **Network at launch.** Today the boss launcher exits in <1s with no network calls. Adding `gh issue view` to that path means a transient GitHub outage can fail launch. Mitigate: do the fetch inside `bossgremlin.py`'s startup (after the launcher has detached), not in `launch.sh`. A failed fetch surfaces as a `bail` on the very first handoff with a clear error in `state.json`.
- **Cross-repo issue sources.** `ghgremlin` already supports `owner/repo#42`. For boss, the chain itself runs against the current repo's `main`, so a cross-repo spec source is fine — it's just spec content. But `bossgremlin.py` should not assume the issue lives in the chain's target repo. Carry over `ghgremlin.sh:228-265`'s `_target_repo` resolution.
- **Helper-extraction scope creep.** The cleanest hoist of `ghgremlin.sh:184-265` rewrites `ghgremlin.sh`'s plan-source block. Resist the urge to refactor adjacent ghgremlin code in the same PR — unrelated diff noise.

## Open questions

- Helper language: shell function sourced by both, or Python module called via subprocess from `ghgremlin.sh`? Bash sourcing keeps `ghgremlin.sh` self-contained but means a new file under `_bg/`. Python is more testable but adds a process boundary inside ghgremlin's hot path.
- Should the snapshot live at `<state-dir>/spec.md` (parallel to `ghgremlin`'s `artifacts/plan.md`) or `<state-dir>/artifacts/spec.md` (under the existing artifacts subdir)? Default: `<state-dir>/spec.md`, since the boss has no `artifacts/` subdir today and `spec.md` is conceptually different from per-stage artifacts.

Ref #77 (the chain currently running against issue #80 had to materialize the spec to `/tmp` manually — this issue is the cleanup).

xbrianh:
## Summary

- Mirror ghgremlin's `--plan` dual file-or-issue-ref form for bossgremlin: `42`, `#42`, `owner/repo#42`, and full GitHub issue URLs are now accepted alongside local file paths. Removes the manual `gh issue view <n> > /tmp/spec.md` dance.
- At chain start the boss snapshots the spec into `<state-dir>/spec.md`; every handoff reads the snapshot, never the original input. On rescue the snapshot is authoritative — no re-fetch. `issue_url` / `issue_num` are persisted into `boss_state.json`.
- The issue-ref parser and `gh issue view --json …` wrapper move into `pipeline/gh_utils.py` so both orchestrators share one implementation.
- `launch.sh`'s file-existence pre-check is narrowed to `localgremlin` (which has no issue-ref form). For bossgremlin, the network fetch runs inside the orchestrator after launch detaches, so a transient GitHub outage surfaces as a bail in the boss log instead of failing the launch.
- `skills/bossgremlin/SKILL.md` updated: `argument-hint` → `<spec-path|issue-ref>`, `--plan` blurb documents all four shapes and the snapshot semantics, and `spec.md` is added to the artifacts list.

## Test plan

- [x] Six new unit tests for `_resolve_plan_source` (file path, empty file, single-repo issue ref, cross-repo issue ref, unknown shape, idempotent rescue).
- [x] New `boss_main` smoke test launches with `--plan 80` and verifies `spec.md` lands in the state dir with the issue body and `issue_url` / `issue_num` are persisted into `boss_state.json`.
- [x] Full pipeline test suite: 62 passed.
- [ ] Manual: launch a real boss against a live GitHub issue (`bossgremlin --plan <issue-ref> --chain-kind gh`) and verify `spec.md` is written before the first handoff.
- [ ] Manual: rescue a chain mid-flight and confirm no second `gh issue view` call.

Closes #82

### DIFF
--- a/pipeline/gh_utils.py
+++ b/pipeline/gh_utils.py
@@ -54,19 +54,32 @@ def parse_issue_ref(
     return None, None
 
 
+VIEW_ISSUE_TIMEOUT = 30  # seconds; bounds `gh issue view` shell-out
+
+
 def view_issue(issue_ref: str, repo: str) -> dict:
     """Fetch ``number``, ``url``, ``body`` for an issue via ``gh issue view``.
 
-    Returns the parsed JSON dict. Raises ``RuntimeError`` when ``gh`` fails or
-    returns unparseable output.
+    Returns the parsed JSON dict. Raises ``RuntimeError`` when ``gh`` fails,
+    times out, or returns unparseable output. The timeout is bounded so a
+    hung ``gh`` (network stall, credential prompt) cannot block chain start
+    indefinitely.
     """
-    r = subprocess.run(
-        [
-            "gh", "issue", "view", issue_ref, "--repo", repo,
-            "--json", "number,url,body",
-        ],
-        capture_output=True, text=True, check=False,
-    )
+    try:
+        r = subprocess.run(
+            [
+                "gh", "issue", "view", issue_ref, "--repo", repo,
+                "--json", "number,url,body",
+            ],
+            capture_output=True, text=True, check=False,
+            timeout=VIEW_ISSUE_TIMEOUT,
+        )
+    except subprocess.TimeoutExpired as exc:
+        raise RuntimeError(
+            f"timed out after {VIEW_ISSUE_TIMEOUT}s while resolving issue "
+            f"{issue_ref!r} in {repo!r} via `gh issue view`; check GitHub "
+            f"CLI authentication, prompts, and network connectivity"
+        ) from exc
     if r.returncode != 0:
         raise RuntimeError(
             f"could not resolve issue {issue_ref!r} in {repo!r}: {r.stderr.strip()}"
--- a/pipeline/gh_utils.py
+++ b/pipeline/gh_utils.py
@@ -6,10 +6,11 @@ live here so the stage modules stay focused on orchestration.
 
 from __future__ import annotations
 
+import json
 import re
 import subprocess
 import sys
-from typing import Callable, List, Optional
+from typing import Callable, List, Optional, Tuple
 
 
 def get_repo() -> str:
@@ -23,6 +24,60 @@ def get_repo() -> str:
             f"not in a gh-recognized repo: {r.stderr.strip() or r.stdout.strip()}"
         )
     return r.stdout.strip()
+
+
+def parse_issue_ref(
+    plan_source: str, repo: str
+) -> Tuple[Optional[str], Optional[str]]:
+    """Parse an issue reference string into ``(target_repo, issue_num)``.
+
+    Recognized shapes (matching ghgremlin's --plan contract):
+      * ``42`` / ``#42``                              → (repo, "42")
+      * ``owner/name#42``                             → ("owner/name", "42")
+      * ``https://github.com/owner/name/issues/42``   → ("owner/name", "42")
+
+    Returns ``(None, None)`` when ``plan_source`` doesn't match any shape so
+    the caller can distinguish issue refs from local file paths.
+    """
+    m = re.match(r"^#?([0-9]+)$", plan_source)
+    if m:
+        return repo, m.group(1)
+    m = re.match(r"^([A-Za-z0-9_.-]+/[A-Za-z0-9_.-]+)#([0-9]+)$", plan_source)
+    if m:
+        return m.group(1), m.group(2)
+    m = re.match(
+        r"^https://github\.com/([A-Za-z0-9_.-]+/[A-Za-z0-9_.-]+)/issues/([0-9]+)(#.*)?$",
+        plan_source,
+    )
+    if m:
+        return m.group(1), m.group(2)
+    return None, None
+
+
+def view_issue(issue_ref: str, repo: str) -> dict:
+    """Fetch ``number``, ``url``, ``body`` for an issue via ``gh issue view``.
+
+    Returns the parsed JSON dict. Raises ``RuntimeError`` when ``gh`` fails or
+    returns unparseable output.
+    """
+    r = subprocess.run(
+        [
+            "gh", "issue", "view", issue_ref, "--repo", repo,
+            "--json", "number,url,body",
+        ],
+        capture_output=True, text=True, check=False,
+    )
+    if r.returncode != 0:
+        raise RuntimeError(
+            f"could not resolve issue {issue_ref!r} in {repo!r}: {r.stderr.strip()}"
+        )
+    try:
+        return json.loads(r.stdout)
+    except json.JSONDecodeError as exc:
+        raise RuntimeError(
+            f"gh issue view returned invalid JSON for {issue_ref!r} in {repo!r}. "
+            f"stdout: {r.stdout.strip()} stderr: {r.stderr.strip()}"
+        ) from exc
 
 
 def extract_gh_url(
--- a/pipeline/orchestrators/boss.py
+++ b/pipeline/orchestrators/boss.py
@@ -491,27 +491,58 @@ def _resolve_plan_source(plan: str, state_dir: str) -> Tuple[str, str, str]:
     spec_dest = os.path.join(state_dir, "spec.md")
 
     if os.path.isfile(spec_dest) and os.path.getsize(spec_dest) > 0:
+        # Rescue path: a previous run already wrote the snapshot. Recover any
+        # issue_url / issue_num it persisted to state.json so the rescue
+        # doesn't silently strip the issue link from boss_state.json. (We
+        # avoid re-fetching from GitHub — the snapshot is authoritative.)
         log(f"reusing existing spec snapshot: {spec_dest}")
-        return spec_dest, "", ""
+        recovered_url = ""
+        recovered_num = ""
+        try:
+            state_data = load_json(os.path.join(state_dir, "state.json"))
+            recovered_url = state_data.get("issue_url") or ""
+            recovered_num = state_data.get("issue_num") or ""
+        except Exception:
+            pass
+        return spec_dest, recovered_url, recovered_num
+
+    # Classify the input shape *before* shelling out to `gh repo view`. For a
+    # typo like `--plan not-a-ref`, this lets us fail fast with a clear error
+    # instead of forcing the caller to be inside a gh-recognized repo first.
+    target_repo, issue_ref = parse_issue_ref(plan, "")
 
     if os.path.isfile(plan):
-        if os.path.getsize(plan) == 0:
-            die(f"--plan: file is empty: {plan}")
-        shutil.copyfile(plan, spec_dest)
+        try:
+            if os.path.getsize(plan) == 0:
+                die(f"--plan: file is empty: {plan}")
+            shutil.copyfile(plan, spec_dest)
+        except OSError as exc:
+            die(f"--plan: failed to read/copy local plan file {plan!r}: {exc}")
         log(f"plan source (file): {plan} → {spec_dest}")
         return spec_dest, "", ""
+
+    if target_repo is None and issue_ref is None:
+        # parse_issue_ref returned (None, None) for the bare-number form
+        # only because we passed an empty repo. Re-check by looking for the
+        # bare-number shape directly so the error message is accurate.
+        if not re.match(r"^#?[0-9]+$", plan):
+            die(f"--plan: not a readable file or recognized issue reference: {plan}")
 
     if shutil.which("gh") is None:
         die(f"--plan: gh CLI not found; required to resolve issue reference {plan!r}")
 
-    try:
-        repo = get_repo()
-    except RuntimeError as exc:
-        die(f"--plan: {exc}")
-
-    target_repo, issue_ref = parse_issue_ref(plan, repo)
-    if target_repo is None:
-        die(f"--plan: not a readable file or recognized issue reference: {plan}")
+    # Resolve the bare-number form against the current repo; cross-repo
+    # forms (`owner/name#42` or full URLs) already carry their own repo.
+    if target_repo is None or target_repo == "":
+        try:
+            target_repo = get_repo()
+        except RuntimeError as exc:
+            die(f"--plan: {exc}")
+        # Re-parse against the resolved repo to populate issue_ref for the
+        # bare-number form.
+        target_repo, issue_ref = parse_issue_ref(plan, target_repo)
+        if target_repo is None:
+            die(f"--plan: not a readable file or recognized issue reference: {plan}")
 
     try:
         issue_data = view_issue(issue_ref, target_repo)
@@ -527,6 +558,9 @@ def _resolve_plan_source(plan: str, state_dir: str) -> Tuple[str, str, str]:
 
     with open(spec_dest, "w", encoding="utf-8") as f:
         f.write(body + "\n")
+    # Persist the issue identifiers to state.json immediately so a crash
+    # between this point and init_boss_state() doesn't strip them on rescue.
+    patch_state(issue_url=issue_url, issue_num=issue_num)
     log(f"plan source (issue {target_repo}#{issue_ref}): {issue_url} → {spec_dest}")
     return spec_dest, issue_url, issue_num
 
@@ -549,10 +583,11 @@ def _maybe_set_description_from_spec(state_dir: str) -> None:
     if data.get("description_explicit"):
         return
     try:
+        # Read up to 50 lines; .splitlines() stops at EOF, so short specs
+        # still yield their leading lines (the prior list-comprehension
+        # form raised StopIteration on short files and dropped the H1).
         with open(spec_file, encoding="utf-8") as f:
-            head_lines = [next(f) for _ in range(50)]
-    except StopIteration:
-        head_lines = []
+            head_lines = f.read().splitlines()[:50]
     except Exception:
         return
     h1 = ""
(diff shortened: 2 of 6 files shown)


xbrianh (commented):
## Scope review

The diff matches the plan closely. All seven plan tasks are addressed: helper extraction (`parse_issue_ref` / `view_issue` in `pipeline/gh_utils.py`), `_bg/launch.sh` loosened for `bossgremlin` only, `_resolve_plan_source` snapshots into `<state-dir>/spec.md`, idempotent rescue, `boss_state.json` now carries `issue_url` / `issue_num`, SKILL.md `--plan` blurb and `argument-hint` updated, and a `boss_main` smoke test verifies the issue-ref → snapshot path end-to-end.

One scope flag:

### Behavior change without dedicated tests — `_maybe_set_description_from_spec`

`pipeline/orchestrators/boss.py` adds `_maybe_set_description_from_spec` (~30 lines) and wires it into chain start. This function is necessary scope (the launch.sh diff deliberately leaves DESCRIPTION empty for `bossgremlin` issue-refs, so something must fill it), and that's a fine inclusion.

However, it has several distinct branches — missing state.json, missing spec.md, malformed JSON, `description_explicit=True`, no H1 in the first 50 lines, H1 truncation to 60 chars — and none of them are exercised by a unit test. The new `test_boss_main_plan_issue_ref_snapshots_spec` smoke test asserts on `spec.md`, `spec_path`, `issue_url`, `issue_num`, but does not assert that the description was filled from the H1. `_make_gremlin_state` writes a state.json with no `description` field at all, so the smoke test wouldn't even catch a regression where description-fill silently no-ops.

This is a minor flag — the plan only called out a smoke test for the snapshot path, not for description-fill. But since description-fill is observable behavior added by this PR (and is what makes `/gremlins` listings work for issue-ref launches), at least one assertion in the smoke test on `state.json["description"]` (e.g. `"Plan from issue"`) would close the loop cheaply.

### Not flagged

- The `_parse_issue_ref = parse_issue_ref` shim in `pipeline/orchestrators/gh.py` is a deliberate test-compat alias and the comment says so — within scope of the helper extraction.
- The `_resolve_plan_source` rewrite in `gh.py` (replacing the inline `subprocess.run` block with `view_issue`) is the dedup the plan called for. Within scope.
- No hitchhiker refactors of adjacent ghgremlin code, matching the explicit risk callout.
- No mixed concerns visible in this single-PR diff.

pipeline/orchestrators/boss.py
```diff
@@ -462,6 +472,99 @@ def _parse_boss_args(argv: List[str]) -> argparse.Namespace:
...
+        return
+    try:
+        data = load_json(state_file)
+    except Exception:
+        return
+    if data.get("description_explicit"):
+        return
+    try:
+        with open(spec_file, encoding="utf-8") as f:
+            head_lines = [next(f) for _ in range(50)]
```
xbrianh: **Bug: this drops the H1 for every spec shorter than 50 lines.**

`[next(f) for _ in range(50)]` raises `StopIteration` on the first short-iterator call. The list comp is *not* a generator, so PEP 479 doesn't kick in — but when the exception fires, the partial result accumulated by the comprehension is discarded. Control jumps to `except StopIteration:` and `head_lines = []`, so the H1 loop runs over an empty list and `description` is never patched.

Verified on a 3-line file:
```
>>> f = io.StringIO('line1\nline2\nline3\n')
>>> try: head = [next(f) for _ in range(50)]
... except StopIteration: head = []
>>> head
[]
```
Most specs are well under 50 lines, so this path effectively never fills in the description for issue-ref boss launches (which is the whole reason this helper exists — `launch.sh` now leaves DESCRIPTION empty for `bossgremlin`+`--plan <issue-ref>`).

Mirror what `gh.py`'s `_update_description_from_plan` does:
```python
lines = open(spec_file, encoding="utf-8").read().splitlines()[:50]
```
or use `itertools.islice(f, 50)`. Either preserves the lines actually read.

This is also why the new `test_boss_main_plan_issue_ref_snapshots_spec` smoke test passes despite the bug — the test verifies `spec.md` and `boss_state.json`, but doesn't assert anything about `state.json["description"]`. Worth adding that assertion once the read is fixed.

pipeline/orchestrators/boss.py
```diff
@@ -462,6 +472,99 @@ def _parse_boss_args(argv: List[str]) -> argparse.Namespace:
...
+    before persisting boss_state.json.
+
+    Returns ``(spec_path, issue_url, issue_num)``. ``issue_url`` /
+    ``issue_num`` are empty strings for local-file inputs.
+    """
+    spec_dest = os.path.join(state_dir, "spec.md")
+
+    if os.path.isfile(spec_dest) and os.path.getsize(spec_dest) > 0:
+        log(f"reusing existing spec snapshot: {spec_dest}")
+        return spec_dest, "", ""
```
xbrianh: Idempotent path returns `""` for `issue_url` / `issue_num` even when the snapshot originated from an issue ref. If the chain crashed in the narrow window between writing `spec.md` and writing `boss_state.json`, the rescue will re-enter this branch, take the snapshot, and `init_boss_state` will persist empty strings for both fields — losing the issue link that `/gremlins` status now uses (per the field's docstring on L196–198) and silently breaking the `Closes #N` provenance for any downstream tooling that reads it.

The window is small but the failure mode is silent. Two reasonable mitigations:
- Persist `issue_url` / `issue_num` to `state.json` (via `patch_state`) immediately after the snapshot write, then read them back on the idempotent path.
- Or, more cheaply: re-parse the original `plan` arg here (it's still in `pipeline_args`) and re-derive `issue_url` / `issue_num` from the snapshot — no network call needed since the body already contains the issue identifiers in its frontmatter for typical specs (or just call `view_issue` again, since the rescue is rare).

At minimum, leave a comment acknowledging the trade-off so a future reader doesn't assume idempotency is loss-free.

pipeline/orchestrators/boss.py
```diff
@@ -462,6 +472,99 @@ def _parse_boss_args(argv: List[str]) -> argparse.Namespace:
...
+        die(f"--plan: gh CLI not found; required to resolve issue reference {plan!r}")
+
+    try:
+        repo = get_repo()
+    except RuntimeError as exc:
+        die(f"--plan: {exc}")
+
+    target_repo, issue_ref = parse_issue_ref(plan, repo)
+    if target_repo is None:
+        die(f"--plan: not a readable file or recognized issue reference: {plan}")
```
xbrianh: Minor: `get_repo()` is called before we know whether `plan` is a recognized issue ref shape. For a typo like `--plan not-a-ref`, this means we shell out to `gh repo view` and require the caller to be inside a gh-recognized repo just to fail with the recognizable error. Cheap fix: call `parse_issue_ref(plan, "")` first to detect the shape, and only resolve `repo` for the bare-number form (`42` / `#42`) that actually needs it. Not a blocker — just wasteful in the error path.

pipeline/orchestrators/gh.py
```diff
@@ -131,21 +131,11 @@ def _read_state_field(sf: Optional[pathlib.Path], field: str) -> str:
...
-        plan_source,
-    )
-    if m:
-        return m.group(1), m.group(2)
-    return None, None
+# Test compatibility shim — tests import _parse_issue_ref from this module
+# directly (and patch _fetch_issue_body below). The canonical implementation
+# lives in pipeline.gh_utils.parse_issue_ref so the boss orchestrator can
+# share it.
+_parse_issue_ref = parse_issue_ref
```
xbrianh, line 138: Test compatibility shim is fine, but `_fetch_issue_body` (L141, in this file) still uses raw `subprocess.run` rather than the new shared `view_issue` helper. Two near-identical `gh issue view` callsites in the same module are an attractive nuisance — consider routing `_fetch_issue_body` through `view_issue` too (it already returns `body`), so the resume-past-plan path picks up any future fixes for free. Not blocking.

xbrianh (commented):
Solid migration overall — sharing `parse_issue_ref` and `view_issue` in `gh_utils` is a clean factor, the chain-start snapshot semantics make rescue genuinely safer, and the test coverage for `_resolve_plan_source` is thorough. Two issues worth flagging before merge: (1) a real correctness bug in `_maybe_set_description_from_spec` that drops the H1 for any spec under 50 lines (i.e., almost all of them); (2) a smaller data-integrity gap in the idempotent rescue path where `issue_url` / `issue_num` get persisted as empty strings even when the snapshot came from an issue. Inline comments below.

pipeline/orchestrators/boss.py
```diff
@@ -462,6 +472,99 @@ def _parse_boss_args(argv: List[str]) -> argparse.Namespace:
...
+        data = load_json(state_file)
+    except Exception:
+        return
+    if data.get("description_explicit"):
+        return
+    try:
+        with open(spec_file, encoding="utf-8") as f:
+            head_lines = [next(f) for _ in range(50)]
+    except StopIteration:
+        head_lines = []
```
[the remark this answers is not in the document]
xbrianh: Fixed. Replaced `[next(f) for _ in range(50)]` with `f.read().splitlines()[:50]` to mirror `gh.py`'s `_update_description_from_plan`; the smoke test now asserts `state.json["description"]` is filled from the H1 to lock in regression coverage.

pipeline/orchestrators/boss.py
```diff
@@ -462,6 +472,99 @@ def _parse_boss_args(argv: List[str]) -> argparse.Namespace:
...
+    spec_dest = os.path.join(state_dir, "spec.md")
+
+    if os.path.isfile(spec_dest) and os.path.getsize(spec_dest) > 0:
+        log(f"reusing existing spec snapshot: {spec_dest}")
+        return spec_dest, "", ""
+
+    if os.path.isfile(plan):
+        if os.path.getsize(plan) == 0:
+            die(f"--plan: file is empty: {plan}")
+        shutil.copyfile(plan, spec_dest)
```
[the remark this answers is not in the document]
xbrianh: Fixed. Wrapped the empty-size check and `shutil.copyfile` in try/except OSError so permission errors / broken symlinks surface as a clean `fatal: --plan: failed to read/copy local plan file ...` instead of an uncaught traceback.

pipeline/gh_utils.py
```diff
@@ -25,6 +26,60 @@ def get_repo() -> str:
...
+    Returns the parsed JSON dict. Raises ``RuntimeError`` when ``gh`` fails or
+    returns unparseable output.
+    """
+    r = subprocess.run(
+        [
+            "gh", "issue", "view", issue_ref, "--repo", repo,
+            "--json", "number,url,body",
+        ],
+        capture_output=True, text=True, check=False,
+    )
```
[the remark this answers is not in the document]
xbrianh: Fixed. Added a 30s `timeout` (constant `VIEW_ISSUE_TIMEOUT`) and convert `subprocess.TimeoutExpired` into a `RuntimeError` with actionable context, so both orchestrators fail fast on a hung `gh` instead of blocking chain start.

skills/bossgremlin/SKILL.md
```diff
@@ -27,18 +27,25 @@ Before invoking the launcher, compose a short (≤60 characters) human-readable
...
 ```
-~/.claude/skills/_bg/launch.sh --description "<phrase>" bossgremlin --plan <spec-path> --chain-kind <local|gh> [--model <model>]
+~/.claude/skills/_bg/launch.sh --description "<phrase>" bossgremlin --plan <spec-path|issue-ref> --chain-kind <local|gh> [--model <model>]
 ```
 
 Flags:
 
-- `--plan <spec-path>` — (required) absolute path to the top-level spec file describing the overall multi-step goal. Must be a non-empty file. This file is immutable for the life of the chain — the boss passes it to the first handoff agent and never reads it.
+- `--plan <spec-path|issue-ref>` — (required) the top-level spec describing the overall multi-step goal. Four forms, mirroring `/ghgremlin --plan`:
+    - **Local file path** — an absolute path to a non-empty spec file.
```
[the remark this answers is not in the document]
xbrianh: Fixed by relaxing the docs to "absolute or relative path" — it matches the implementation and the other gremlins.

pipeline/orchestrators/boss.py
xbrianh: Fixed — same as the Copilot comment above. `f.read().splitlines()[:50]` mirrors `gh.py`'s `_update_description_from_plan`. Added a `state.json["description"] == "Plan from issue"` assertion to the smoke test so the regression is locked in.

pipeline/orchestrators/boss.py
xbrianh: Fixed via mitigation #1: `_resolve_plan_source` now calls `patch_state(issue_url=..., issue_num=...)` immediately after the snapshot write, and the idempotent path reads them back from `state.json` instead of returning empty strings. Two new tests cover the persistence (`test_resolve_plan_source_persists_issue_metadata_on_first_fetch`) and the rescue-recovery (`test_resolve_plan_source_idempotent_recovers_issue_metadata`).

pipeline/orchestrators/boss.py
xbrianh: Fixed. Reordered `_resolve_plan_source` so `parse_issue_ref(plan, "")` runs first to classify the shape; `get_repo()` is only called for the bare-number form (`42` / `#42`). A typo like `--plan not-a-ref` now fails fast with the readable error and never shells out to `gh repo view`.

pipeline/orchestrators/gh.py
xbrianh: Fixed. `_fetch_issue_body` now routes through `view_issue` (using its returned `body` field), so both `gh issue view` callsites in this module share one implementation and pick up the new timeout for free.

Issue #58, opened 2026-04-24
Add mock tests for gremlin orchestration (unit + shell integration)


xbrianh:
## Motivation

The gremlin pipeline has no automated tests. The `--model` crash and the rescue agent scope bug both slipped through because there was no way to exercise the orchestration logic without paying for a real `claude -p` run. Two test layers would catch the class of failures we keep hitting.

## Layer 1 — orchestration unit tests (Python)

Mock `subprocess.run` in the Python orchestrators and test control logic directly:

- `gremlins.py`: state transitions (running → dead → rescued → running), land/rescue/close flows, bail-class exclusion, attempt cap enforcement, `_land_local` squash path
- `bossgremlin.py`: child sequencing, handoff → launch → wait → land loop, rescue-then-land path, bail propagation
- `handoff.py`: `next-plan` / `chain-done` / `bail` signal parsing, child plan and updated plan file writing
- `localgremlin.py`: stage ordering, reviewer parallelism, address stage

No real processes, no git, no cost.

## Layer 2 — shell integration tests (fake `claude` binary)

Put a fake `claude` script first on `PATH` that matches invocation patterns by stage and emits minimal valid `stream-json` (or `text` for `--output-format text` calls). Use a real throwaway git repo. This exercises `launch.sh`, `ghgremlin.sh`, and `localgremlin.py` end-to-end.

Coverage targets:
- `launch.sh`: worktree creation, state dir layout, `pipeline_args` storage, `--resume` reload
- `ghgremlin.sh`: `--plan <file>` path, `--model` forwarding to `CLAUDE_FLAGS`, resume from each stage, `unknown flag` crash on bad args
- `localgremlin.py`: plan → implement → review → address pipeline with fake claude outputs
- `gremlins.py rescue`: Phase A runs in scratch dir (not worktree), only `state.json` is writable, `unsalvageable` declared for anything else

The fake `claude` only needs to handle the handful of invocation patterns each script uses; these are stable and enumerable.

## Out of scope

Real `claude` invocations, real GitHub PRs, real cost.

xbrianh:
Splitting into two narrower issues since Layer 1 (Python orchestration tests) landed in #114 for the `pipeline/` package:

- #126 — Layer 1 remainder: unit tests for `skills/gremlins/gremlins.py` and `skills/handoff/handoff.py` (the parts that live outside `pipeline/`).
[...]

PR #134, opened 2026-04-27
Shell integration tests: fake claude harness for launch.sh / ghgremlin / localgremlin / rescue

### LINKED ISSUE #127: Shell integration tests: fake claude binary harness for launch.sh / ghgremlin.sh
Split out from #58 — Layer 2 of the original test plan. The Python tests stub `shutil.which`, but no harness drives the shell scripts end-to-end.

## Approach

Put a fake `claude` script first on `PATH` that matches invocation patterns by stage and emits minimal valid `stream-json` (or `text` for `--output-format text`). Use a real throwaway git repo per test.

## Coverage targets

- `_bg/launch.sh`: worktree creation, state dir layout, `pipeline_args` storage, `--resume` reload
- `ghgremlin/ghgremlin.sh`: `--plan <file>` and `--plan <issue-ref>` paths, `--model` forwarding to `CLAUDE_FLAGS`, resume from each stage, unknown-flag rejection
- `localgremlin` shim → `pipeline.cli local`: plan → implement → reviewers (parallel) → address pipeline with fake claude outputs
- `gremlins.py rescue` Phase A: runs in scratch dir (not the worktree), only `state.json` is writable, `unsalvageable` declared for anything outside that

The fake `claude` only needs to handle the handful of invocation patterns each script uses; these are stable and enumerable.

## Out of scope

Real `claude` invocations, real GitHub PRs, real cost.

xbrianh:
## Summary

Layer 2 of the test plan from #58: 28 new shell-integration tests that drive the gremlin shell scripts end-to-end with a fake `claude` (and `gh` where needed) on PATH and a real throwaway git repo per test. Catches subprocess-invocation pathway issues (CLAUDE_FLAGS, --output-format, --resume, --model forwarding) that the FakeClaudeClient Python tests can't reach.

### Fixtures (`pipeline/tests/fixtures/`)

- **`fake_claude.py`** — classifies stages by prompt content, performs the on-disk side effects each stage's post-condition checks for (write plan.md, commit, write review file, write rescue marker), and emits valid stream-json or text. Records each invocation to ` ${FAKE_CLAUDE_LOG}` so tests can assert what was called with what model in what cwd. `FAKE_CLAUDE_FAIL_AT=<stage>` forces a non-zero exit.
- **`fake_gh.py`** — stubs `gh repo view`, `gh issue create`/`view`, `gh pr edit`/`diff`/`merge`, `gh api`. Behavior configurable via env.
- **`shell_env.py`** — `setup_shell_env(tmp_path, …)` builds an isolated HOME/PATH/XDG_STATE_HOME, real git repo (with optional bare `origin` remote so ghgremlin's `git fetch origin` resolves), and symlinks `~/.claude/{skills,pipeline,agents}` at the repo so launch.sh's hardcoded `\$HOME/.claude/...` paths resolve.

### Coverage

- **`test_launch_sh.py`** (11) — state dir layout, `pipeline_args` persistence, `--resume` rehydration + `rescue_count` bump, running / finished-success refusals, `--plan` path absolutization, mutex with positional, empty-plan rejection, unknown-flag / invalid-kind rejection.
- **`test_localgremlin_shell.py`** (3) — full plan→implement→review×3→address chain via `launch.sh`, `-i <model>` forwarding, `--plan <file>` skips the plan stage.
- **`test_ghgremlin_shell.py`** (8) — unknown-flag rejection at both launch.sh and pipeline.cli levels, shim forwards to `pipeline.cli gh`, `--plan <file>` posts an issue / `--plan <issue-ref>` resolves an existing one, `--model` forwarded to every claude call, `--resume-from ghreview` / `--resume-from implement`.
- **`test_rescue_phase_a.py`** (6) — diagnosis agent's cwd is the scratch dir not the worktree, `unsalvageable` / `structural` markers write the right `bail_reason`, missing marker → `diagnosis_no_marker`, `fixed` verdict invokes `launch.sh --resume <id>`, headless rescue refuses excluded bail classes without ever spawning claude.

All 92 tests (64 existing + 28 new) pass; full suite runs in ~30s.

Closes #127

## Test plan

- [x] Run `cd pipeline && PYTHONPATH=. python -m pytest tests/` — 92 passed
- [x] Run each new test file in isolation to confirm it passes solo
- [x] Verify launch.sh / pipeline.cli / gremlins.py source files unchanged from main (only test additions)

### DIFF
--- a/pipeline/tests/fixtures/fake_claude.py
+++ b/pipeline/tests/fixtures/fake_claude.py
@@ -1,0 +1,339 @@
+#!/usr/bin/env python
+"""Drop-in fake `claude` binary for shell integration tests.
+
+Parses `claude -p [flags] <prompt>` argv, classifies the prompt to identify
+which pipeline stage spawned it, performs the side effects that stage's
+post-conditions check for (write plan.md, create a commit, write a review
+file, etc.), and emits minimal valid stream-json (or text) on stdout.
+
+Each invocation appends an entry to ``$FAKE_CLAUDE_LOG`` (one JSON line
+per call) so tests can assert what was invoked, with what model, in what
+cwd. ``$FAKE_CLAUDE_FAIL_AT`` (a stage name) makes the matching invocation
+exit non-zero — used to test rescue / resume paths.
+"""
+
+from __future__ import annotations
+
+import argparse
+import json
+import os
+import pathlib
+import re
+import subprocess
+import sys
+import time
+import uuid
+
+
+def emit_event(evt: dict) -> None:
+    sys.stdout.write(json.dumps(evt) + "\n")
+    sys.stdout.flush()
+
+
+def emit_minimal_stream(session_id: str, *, extra: list = None) -> None:
+    emit_event({
+        "type": "system", "subtype": "init",
+        "session_id": session_id,
+        "model": "fake", "cwd": os.getcwd(),
+    })
+    for evt in extra or []:
+        emit_event(evt)
+    emit_event({
+        "type": "result", "subtype": "success",
+        "num_turns": 1, "total_cost_usd": 0,
+    })
+
+
+def emit_pr_create_stream(session_id: str, pr_url: str) -> None:
+    tu_id = "tu-" + uuid.uuid4().hex[:8]
+    emit_minimal_stream(
+        session_id,
+        extra=[
+            {"type": "assistant", "message": {"content": [
+                {"type": "tool_use", "id": tu_id, "name": "Bash",
+                 "input": {"command": "gh pr create --base main"}},
+            ]}},
+            {"type": "user", "message": {"content": [
+                {"type": "tool_result", "tool_use_id": tu_id,
+                 "content": pr_url},
+            ]}},
+        ],
+    )
+
+
+def emit_issue_create_stream(session_id: str, issue_url: str) -> None:
+    tu_id = "tu-" + uuid.uuid4().hex[:8]
+    emit_minimal_stream(
+        session_id,
+        extra=[
+            {"type": "assistant", "message": {"content": [
+                {"type": "tool_use", "id": tu_id, "name": "Bash",
+                 "input": {"command": "gh issue create --title foo"}},
+            ]}},
+            {"type": "user", "message": {"content": [
+                {"type": "tool_result", "tool_use_id": tu_id,
+                 "content": issue_url},
+            ]}},
+        ],
+    )
+
+
+def log_invocation(record: dict) -> None:
+    log_path = os.environ.get("FAKE_CLAUDE_LOG")
+    if not log_path:
+        return
+    try:
+        with open(log_path, "a", encoding="utf-8") as fh:
+            fh.write(json.dumps(record) + "\n")
+    except OSError:
+        pass
+
+
+def maybe_fail_at(stage: str) -> None:
+    if os.environ.get("FAKE_CLAUDE_FAIL_AT") == stage:
+        sys.stderr.write(f"fake claude: forced failure at {stage}\n")
+        sys.exit(1)
+
+
+def find_path_in_prompt(prompt: str, pattern: str) -> str:
+    m = re.search(pattern, prompt)
+    return m.group(1) if m else ""
+
+
+def git_commit_changes(message: str) -> bool:
+    """Stage all changes and commit. Returns True if a commit was made."""
+    try:
+        subprocess.run(["git", "add", "-A"], check=False, capture_output=True)
+        r = subprocess.run(
+            ["git", "diff", "--cached", "--quiet"],
+            check=False, capture_output=True,
+        )
+        if r.returncode == 0:
+            return False  # nothing staged
+        subprocess.run(
+            ["git", "commit", "-m", message],
+            check=False, capture_output=True,
+        )
+        return True
+    except OSError:
+        return False
+
+
+def handle_plan(prompt: str, session_id: str) -> int:
+    plan_file = find_path_in_prompt(prompt, r"write it to the file `([^`]+)`")
+    if plan_file:
+        pathlib.Path(plan_file).parent.mkdir(parents=True, exist_ok=True)
+        pathlib.Path(plan_file).write_text(
+            "# Test Plan\n\n## Context\nFake claude generated plan.\n\n"
+            "## Tasks\n- [ ] Touch a file\n",
+            encoding="utf-8",
+        )
+    emit_minimal_stream(session_id)
+    return 0
+
+
+def handle_implement(prompt: str, session_id: str) -> int:
+    # Create a new file so review-code's diff is non-empty.
+    target = pathlib.Path("fake_impl.txt")
+    target.write_text("fake implementation\n", encoding="utf-8")
+    in_git = subprocess.run(
+        ["git", "rev-parse", "--git-dir"],
+        capture_output=True, check=False,
+    ).returncode == 0
+    if in_git:
+        # Commit so HEAD advances (post-impl invariant for both local and gh).
+        git_commit_changes("impl: fake implementation")
+    emit_minimal_stream(session_id)
+    return 0
+
+
+def handle_review(prompt: str, session_id: str) -> int:
+    out_file = find_path_in_prompt(prompt, r"`([^`]+\.md)`\s+is the canonical")
+    if out_file:
+        out_path = pathlib.Path(out_file)
+        out_path.parent.mkdir(parents=True, exist_ok=True)
+        out_path.write_text(
+            "# Review (fake)\n\n## Summary\nFake review.\n\n"
+            "## Findings\nNo issues.\n",
+            encoding="utf-8",
+        )
+    emit_minimal_stream(session_id)
+    return 0
+
+
+def handle_address(prompt: str, session_id: str) -> int:
+    in_git = subprocess.run(
+        ["git", "rev-parse", "--git-dir"],
+        capture_output=True, check=False,
+    ).returncode == 0
+    if in_git:
+        # Make a no-op edit so address commit lands.
+        target = pathlib.Path("fake_impl.txt")
+        if target.exists():
+            target.write_text("fake implementation (addressed)\n", encoding="utf-8")
+            git_commit_changes("Address review feedback")
+    emit_minimal_stream(session_id)
+    return 0
+
+
+def handle_commit_pr(prompt: str, session_id: str) -> int:
+    pr_url = os.environ.get(
+        "FAKE_CLAUDE_PR_URL", "https://github.com/owner/repo/pull/101"
+    )
+    emit_pr_create_stream(session_id, pr_url)
+    return 0
+
+
+def handle_ghplan(prompt: str, session_id: str) -> int:
+    issue_url = os.environ.get(
+        "FAKE_CLAUDE_ISSUE_URL", "https://github.com/owner/repo/issues/42"
+    )
+    emit_issue_create_stream(session_id, issue_url)
+    return 0
+
+
+def handle_text_title(prompt: str) -> int:
+    # `--plan <file>` flow asks for a one-line GitHub issue title.
+    sys.stdout.write("Test issue title from fake claude\n")
+    sys.stdout.flush()
+    return 0
+
+
+def handle_rescue_diagnosis(prompt: str) -> int:
+    """Rescue diagnosis-step prompt: write the marker file with status from env.
+
+    The marker path is embedded in the prompt; we extract it and write a JSON
+    object. Default verdict is "fixed" — tests override via env to exercise
+    other branches.
+    """
+    m = re.search(r"\b(/[^\s`]+\.done)\b", prompt)
+    marker_path = m.group(1) if m else ""
+    status = os.environ.get("FAKE_CLAUDE_RESCUE_VERDICT", "fixed")
+    summary = os.environ.get("FAKE_CLAUDE_RESCUE_SUMMARY", "fake diagnosis")
+    if marker_path:
+        try:
+            pathlib.Path(marker_path).parent.mkdir(parents=True, exist_ok=True)
+            pathlib.Path(marker_path).write_text(
+                json.dumps({"status": status, "summary": summary}),
+                encoding="utf-8",
+            )
+        except OSError:
+            pass
+    # Probe writability of paths the spec says should be denied. The probe
+    # records observations to FAKE_CLAUDE_LOG so tests can assert what the
+    # diagnosis agent sees from inside the scratch dir.
+    workdir_probe = os.environ.get("FAKE_CLAUDE_RESCUE_WORKDIR_PROBE")
+    if workdir_probe:
+        probe_file = pathlib.Path(workdir_probe) / ".fake_diagnosis_probe"
+        try:
+            probe_file.write_text("probe", encoding="utf-8")
+            log_invocation({
+                "stage": "rescue-probe",
+                "workdir_probe_path": str(probe_file),
+                "workdir_probe_succeeded": True,
+            })
+        except OSError as exc:
+            log_invocation({
+                "stage": "rescue-probe",
+                "workdir_probe_path": str(probe_file),
+                "workdir_probe_succeeded": False,
+                "error": str(exc),
+            })
+    return 0
+
+
+def classify_stage(prompt: str) -> str:
+    """Classify the stage from the prompt string. Order matters — earlier
+    matches win when prompts contain overlapping phrases."""
+    if "diagnosing a failed background gremlin" in prompt:
+        return "rescue-diagnosis"
+    if "Produce a concise GitHub issue title" in prompt:
+        return "plan-title"
+    if "Print ONLY the PR URL on the final line" in prompt:
+        return "commit-pr"
+    if prompt.startswith("/ghreview "):
+        return "ghreview"
+    if prompt.startswith("/ghaddress "):
+        return "ghaddress"
+    if "You are a scope reviewer for a pull request" in prompt:
+        return "scope-review-pr"
+    if prompt.startswith("/ghplan") or "/ghplan" in prompt[:20]:
+        return "ghplan"
+    if "Create a detailed implementation plan" in prompt:
+        return "plan"
+    if "Three independent code reviews" in prompt:
+        return "address"
+    if "is the canonical and required location for your review output" in prompt:
+        return "review"
+    if "Implement every task in the plan above" in prompt:
+        return "implement-local"
+    if "Implement the plan above by making the code changes" in prompt:
+        return "implement-gh"
+    return "unknown"
+
+
+def parse_argv(argv):
+    """Mimic enough of `claude -p`'s flag handling to extract what we care about.
+
+    Returns (output_format, model, prompt, extra_flags).
+    """
+    parser = argparse.ArgumentParser(add_help=False)
+    parser.add_argument("-p", action="store_true")
+    parser.add_argument("--model", default=None)
+    parser.add_argument("--output-format", default="stream-json")
+    parser.add_argument("--permission-mode", default=None)
+    parser.add_argument("--verbose", action="store_true")
+    parser.add_argument("--resume", default=None)
+    args, rest = parser.parse_known_args(argv)
+    # Last positional is the prompt; anything else is extra/unknown.
+    prompt = rest[-1] if rest else ""
+    return args, prompt
+
+
+def main(argv):
+    args, prompt = parse_argv(argv)
+    stage = classify_stage(prompt)
+    session_id = f"sess-{stage}-{uuid.uuid4().hex[:6]}"
+
+    log_invocation({
+        "stage": stage,
+        "model": args.model,
+        "output_format": args.output_format,
+        "resume_session": args.resume,
+        "cwd": os.getcwd(),
+        "prompt_head": prompt[:200],
+    })
+
+    maybe_fail_at(stage)
+
+    if args.output_format == "text":
+        if stage == "plan-title":
+            return handle_text_title(prompt)
+        sys.stdout.write("(fake text output)\n")
+        return 0
+
+    handlers = {
+        "plan": handle_plan,
+        "implement-local": handle_implement,
+        "implement-gh": handle_implement,
+        "review": handle_review,
+        "address": handle_address,
+        "commit-pr": handle_commit_pr,
+        "ghplan": handle_ghplan,
+        "ghreview": lambda p, s: (emit_minimal_stream(s), 0)[1],
+        "ghaddress": lambda p, s: (emit_minimal_stream(s), 0)[1],
+        "scope-review-pr": lambda p, s: (emit_minimal_stream(s), 0)[1],
+        "rescue-diagnosis": lambda p, s: handle_rescue_diagnosis(p),
+    }
+    h = handlers.get(stage)
+    if h is None:
+        sys.stderr.write(f"fake claude: unrecognized stage for prompt head: {prompt[:120]!r}\n")
+        emit_minimal_stream(session_id)
+        return 0
+    if stage == "rescue-diagnosis":
+        return h(prompt, session_id)
+    return h(prompt, session_id)
+
+
+if __name__ == "__main__":
+    sys.exit(main(sys.argv[1:]))
--- a/pipeline/tests/fixtures/fake_claude.py
+++ b/pipeline/tests/fixtures/fake_claude.py
@@ -21,7 +21,6 @@ import pathlib
 import re
 import subprocess
 import sys
-import time
 import uuid
 
 
@@ -82,11 +81,22 @@ def log_invocation(record: dict) -> None:
     log_path = os.environ.get("FAKE_CLAUDE_LOG")
     if not log_path:
         return
+    # Atomic append: a single os.write() of one bytes payload avoids
+    # interleaved/corrupted lines under the parallel review workers
+    # (3 concurrent claude subprocesses write to this log).
+    payload = (json.dumps(record) + "\n").encode("utf-8")
+    fd = None
     try:
-        with open(log_path, "a", encoding="utf-8") as fh:
-            fh.write(json.dumps(record) + "\n")
+        fd = os.open(log_path, os.O_APPEND | os.O_CREAT | os.O_WRONLY, 0o666)
+        os.write(fd, payload)
     except OSError:
         pass
+    finally:
+        if fd is not None:
+            try:
+                os.close(fd)
+            except OSError:
+                pass
 
 
 def maybe_fail_at(stage: str) -> None:
@@ -206,7 +216,7 @@ def handle_rescue_diagnosis(prompt: str) -> int:
     object. Default verdict is "fixed" — tests override via env to exercise
     other branches.
     """
-    m = re.search(r"\b(/[^\s`]+\.done)\b", prompt)
+    m = re.search(r"(/[^\s`]+\.done)\b", prompt)
     marker_path = m.group(1) if m else ""
     status = os.environ.get("FAKE_CLAUDE_RESCUE_VERDICT", "fixed")
     summary = os.environ.get("FAKE_CLAUDE_RESCUE_SUMMARY", "fake diagnosis")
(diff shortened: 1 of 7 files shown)


xbrianh (commented):
## Scope review

### Skipped plan item: rescue writability constraint

The plan's Phase A coverage target says:

> `gremlins.py rescue` Phase A: runs in scratch dir (not the worktree), **only `state.json` is writable, `unsalvageable` declared for anything outside that**

`test_rescue_phase_a.py` covers the scratch-dir cwd constraint and the four verdict branches (`fixed` / `structural` / `unsalvageable` / missing marker), but **no test exercises the writability restriction or the "unsalvageable when the agent writes outside `state.json`" behavior.** `test_rescue_unsalvageable_records_bail` proves the wrapper propagates an `unsalvageable` verdict the fake claude was *told* to emit via env — it doesn't prove anything about how that verdict is actually arrived at by the rescue framework.

If the writability constraint is enforced by the rescue wrapper (and is meant to be testable at this layer), the obvious test is missing. If it's enforced by Claude's permission system and not testable end-to-end with a fake binary, the plan target should be acknowledged as out of reach for this layer.

### Exceeded scope: dead probe infrastructure in `fake_claude.py`

`pipeline/tests/fixtures/fake_claude.py` `handle_rescue_diagnosis` (around the `workdir_probe` block) reads `FAKE_CLAUDE_RESCUE_WORKDIR_PROBE` and emits a `rescue-probe` log entry. **No test in this PR sets `FAKE_CLAUDE_RESCUE_WORKDIR_PROBE` or asserts on the probe log entry.** This appears to be scaffolding for the unwritten writability test above. Either delete it or write the test that consumes it — leaving it in adds noise to the fixture and rots silently as the rescue contract evolves.

### Partial: "resume from each stage" for ghgremlin

The plan calls for "resume from each stage" of the gh pipeline. The PR tests `--resume-from ghreview` and `--resume-from implement`, which are reasonable representative anchors, but `commit-pr`, `request-copilot`, `wait-copilot`, and `ghaddress` resume points are not exercised. Minor flag — reasonable interpretation of the plan is that the resume *mechanism* is what matters and two anchors prove it, but if "each stage" was meant literally, this is incomplete.

### Otherwise scoped correctly

- No unrelated refactors / cleanups outside the test harness.
- `fake_gh.py` is unmentioned in the plan but is necessary to drive the gh pipeline end-to-end without real `gh` calls; in scope.
- `_bg/launch.sh` coverage targets (worktree, state layout, `pipeline_args`, `--resume`, unknown-flag rejection) are all hit, plus a couple of adjacent guards (`--resume` refusal of running / finished-success gremlins, plan/positional mutex, empty-plan rejection) that are reasonable boundary tests for the same surface, not scope creep.
- localgremlin coverage matches the plan (plan → implement → 3 reviewers → address with stage-ordering and artifact assertions).
- Single commit for one test layer matches plan structure; not a mixed-concern commit.

pipeline/tests/fixtures/fake_claude.py
```diff
@@ -0,0 +1,339 @@
...
+
+
+def handle_rescue_diagnosis(prompt: str) -> int:
+    """Rescue diagnosis-step prompt: write the marker file with status from env.
+
+    The marker path is embedded in the prompt; we extract it and write a JSON
+    object. Default verdict is "fixed" — tests override via env to exercise
+    other branches.
+    """
+    m = re.search(r"\b(/[^\s`]+\.done)\b", prompt)
```
xbrianh: **Latent bug — passes on macOS, will fail on Linux CI.**

The `\b` word boundary cannot match between two non-word characters. In the rendered rescue prompt the marker path is preceded by whitespace (line 549 in `gremlins.py`: `       {marker_path}`), so `\b/` does not match the leading `/`. Instead, `re.search` advances and matches at the first word→`/` transition further into the path, dropping the leading component:

```python
>>> import re
>>> re.search(r'\b(/[^\s`]+\.done)\b', '       /tmp/pytest-of-runner/x/rescue-2026.done').group(1)
'/pytest-of-runner/x/rescue-2026.done'   # /tmp prefix lost
```

The tests pass today only because macOS pytest tmpdirs live under `/private/var/folders/...`, and `/var` is a symlink to `/private/var`, so `/var/folders/.../rescue.done` (the regex's truncated capture) still resolves to the right inode. On Linux this won't be the case — the truncated path won't exist, the `try/except OSError: pass` will swallow the failure, the wrapper will see no marker, and the `unsalvageable` / `structural` / `fixed` assertions in `test_rescue_phase_a.py` will all flip to `bail_reason == 'diagnosis_no_marker'`.

Fix: drop the leading `\b` (and ideally the trailing one too, since the prompt may end the path with non-word punctuation). E.g. `r'(/\S+?\.done)\b'` or, more robustly, anchor on the surrounding line context — the prompt always emits the path on its own indented line.

pipeline/tests/fixtures/fake_claude.py
```diff
@@ -0,0 +1,339 @@
...
+from __future__ import annotations
+
+import argparse
+import json
+import os
+import pathlib
+import re
+import subprocess
+import sys
+import time
```
xbrianh: `time` is imported but never used.

pipeline/tests/fixtures/fake_claude.py
```diff
@@ -0,0 +1,339 @@
...
+        "rescue-diagnosis": lambda p, s: handle_rescue_diagnosis(p),
+    }
+    h = handlers.get(stage)
+    if h is None:
+        sys.stderr.write(f"fake claude: unrecognized stage for prompt head: {prompt[:120]!r}\n")
+        emit_minimal_stream(session_id)
+        return 0
+    if stage == "rescue-diagnosis":
+        return h(prompt, session_id)
+    return h(prompt, session_id)
```
xbrianh, line 325: The `if stage == 'rescue-diagnosis':` branch on lines 333-334 returns the same expression as the fallthrough on line 335 — it can be deleted.

pipeline/tests/fixtures/shell_env.py
```diff
@@ -0,0 +1,149 @@
...
+
+from __future__ import annotations
+
+import dataclasses
+import os
+import pathlib
+import subprocess
+import sys
+import time
+from typing import Optional
```
xbrianh: `Optional` is imported but never used.

pipeline/tests/fixtures/shell_env.py
```diff
@@ -0,0 +1,149 @@
...
+        "HOME": str(home),
+        "XDG_STATE_HOME": str(state_root),
+        "PATH": f"{bin_dir}{os.pathsep}{os.environ.get('PATH', '')}",
+        "FAKE_CLAUDE_LOG": str(fake_claude_log),
+        "FAKE_GH_LOG": str(fake_gh_log),
+        # Strip caller's PYTHONPATH; launch.sh will set its own.
+    }
+    env.pop("PYTHONPATH", None)
+    # Suppress git's hint output so tests get clean stderr.
+    env["GIT_OPTIONAL_LOCKS"] = "0"
```
xbrianh, line 117: The comment is misleading: `GIT_OPTIONAL_LOCKS=0` disables optional `.lock` files for read-only commands (e.g. `git status` skipping the index lock); it does not suppress git's `hint:` output. To quiet hints you'd configure `advice.*=false`. If you only meant to suppress lock contention from concurrent test runs the env var is fine — just retitle the comment.

pipeline/tests/fixtures/shell_env.py
```diff
@@ -0,0 +1,149 @@
...
+def read_fake_claude_log(log_path: pathlib.Path) -> list:
+    if not log_path.exists():
+        return []
+    out = []
+    for line in log_path.read_text(encoding="utf-8").splitlines():
+        line = line.strip()
+        if not line:
+            continue
+        try:
+            out.append(__import__("json").loads(line))
```
xbrianh: `__import__('json').loads(line)` here and the `import json` inside `read_state` on line 148 can both be replaced with a single top-level `import json` next to the other imports.

pipeline/tests/fixtures/fake_gh.py
```diff
@@ -0,0 +1,100 @@
...
+            sys.stdout.write(repo + "\n")
+            return 0
+        sys.stdout.write(f"{repo}\n")
+        return 0
+
+    if sub == "issue" and len(argv) >= 2 and argv[1] == "create":
+        url = os.environ.get(
+            "FAKE_GH_ISSUE_URL", "https://github.com/owner/repo/issues/42"
+        )
+        sys.stdout.write(f"Creating issue\n{url}\n")
```
xbrianh: Real `gh issue create` writes only the URL to stdout — no `Creating issue` preamble. If a future caller does `URL=$(gh issue create …)` rather than `tail -1`, this stub will return `"Creating issue\n<url>"` and break. Recommend matching the real CLI: just emit `f"{url}\n"`.

pipeline/tests/test_ghgremlin_shell.py
```diff
@@ -0,0 +1,255 @@
...
+    assert r.returncode == 0, r.stderr
+    gr_id = r.stdout.strip()
+    state_dir = sh.state_root / "claude-gremlins" / gr_id
+    assert wait_for_finished(state_dir, timeout=120)
+
+    log = read_fake_claude_log(sh.fake_claude_log)
+    # Every recorded claude call must carry the model flag.
+    pipeline_stages = {"implement-gh", "commit-pr", "ghreview", "ghaddress", "scope-review-pr"}
+    for entry in log:
+        if entry["stage"] in pipeline_stages:
```
xbrianh, line 179: This loop only checks the model flag on stages that *did* run — if (e.g.) `scope-review-pr` were silently skipped, the test would still pass. Consider adding an upfront `assert pipeline_stages.issubset({e['stage'] for e in log}), …` so the test fails fast when a stage in the expected set never ran.

pipeline/tests/test_ghgremlin_shell.py
```diff
@@ -0,0 +1,255 @@
...
+            assert entry["model"] == "claude-opus-4-7", \
+                f"stage {entry['stage']} got model={entry['model']!r}"
+
+
+def test_ghgremlin_resume_from_ghreview(tmp_path):
```
xbrianh, line 184: Naming nit: this test is named `test_ghgremlin_resume_from_ghreview` but the resume itself is performed by invoking `ghgremlin.sh` directly (line 206) rather than `launch.sh --resume`. The comment on lines 201-204 acknowledges this, but the test name still suggests the launcher's resume path is exercised. Consider `test_pipeline_resume_from_ghreview` (or add a sibling test that hits `launch.sh --resume <id>` with a `--resume-from` injected via the rescue path) so the launcher's resume contract isn't left untested by this layer.

xbrianh (commented):
Solid layer of shell-integration coverage that closes a real gap (subprocess pathway, CLAUDE_FLAGS forwarding, --resume rehydration). The fixtures are well-structured. One concrete bug worth fixing before merge: the marker-path regex in `fake_claude.py` strips a leading word-prefixed path component because `\b` cannot match between whitespace and `/`. On macOS this happens to work (`/var` symlinks to `/private/var`), but on Linux CI (`/tmp/...`) the regex would yield `/pytest-of-…/…/rescue.done`, the marker write would fail silently, and every `unsalvageable`/`structural`/`fixed` rescue assertion would flip to `diagnosis_no_marker`. A few smaller cleanups noted inline.

xbrianh (commented):
## Scope review

### Skipped plan item: rescue Phase A writability constraint

The plan's Phase A target has three parts:

> runs in scratch dir (not the worktree), **only `state.json` is writable**, **`unsalvageable` declared for anything outside that**

`test_rescue_phase_a.py` covers the scratch-dir cwd assertion and the verdict-to-bail mapping for `fixed` / `structural` / `unsalvageable` / no-marker. It does **not** test that writes outside `state.json` are denied, or that such writes cause the diagnosis agent to declare `unsalvageable`. `test_rescue_unsalvageable_records_bail` only proves the wrapper propagates a verdict the fake claude was *told* to emit via env — it says nothing about how the verdict would actually be arrived at.

This is a real gap against an explicitly enumerated plan item, not a "left for later" item.

### Dead infrastructure for the missing test

`pipeline/tests/fixtures/fake_claude.py` `handle_rescue_diagnosis` reads `FAKE_CLAUDE_RESCUE_WORKDIR_PROBE`, attempts a write inside the named directory, and logs a `rescue-probe` entry to `FAKE_CLAUDE_LOG`. **No test in this PR sets that env var or asserts on the probe log.** This is scaffolding for the unwritten writability test above. Either remove it or land the test that consumes it — leaving unused branches in a fixture rots silently as the rescue contract evolves.

### Partial: "resume from each stage" for ghgremlin

Plan says "resume from each stage". PR tests `--resume-from ghreview` and `--resume-from implement` — reasonable anchors but four other stage entry points (`commit-pr`, `request-copilot`, `wait-copilot`, `ghaddress`) are not covered. Minor; flagging in case "each" was meant literally.

### Otherwise scoped correctly

- No unrelated refactors or cleanups outside the test harness.
- `fake_gh.py` is unmentioned in the plan but is necessary to drive the gh pipeline without real `gh` — in scope.
- launch.sh coverage hits all four enumerated targets (worktree, state layout, `pipeline_args`, `--resume`) plus close-by guards (`--resume` refusal on live/finished pids, plan/positional mutex, empty-plan rejection). Reasonable boundary tests on the same surface, not creep.
- localgremlin coverage matches the plan (plan → implement → 3 reviewers → address with stage-ordering and artifact assertions).
- Single test-layer commit matches the plan structure; not a mixed-concern commit.

pipeline/tests/fixtures/fake_claude.py
xbrianh: **Latent bug — passes on macOS, will fail on Linux CI.**

The `\b` word boundary cannot match between two non-word characters. In the rendered rescue prompt the marker path is preceded by whitespace (line 549 in `gremlins.py`: `       {marker_path}`), so `\b/` does not match the leading `/`. Instead, `re.search` advances and matches at the first word→`/` transition further into the path, dropping the leading component:

```python
>>> import re
>>> re.search(r'\b(/[^\s`]+\.done)\b', '       /tmp/pytest-of-runner/x/rescue-2026.done').group(1)
'/pytest-of-runner/x/rescue-2026.done'   # /tmp prefix lost
```

The tests pass today only because macOS pytest tmpdirs live under `/private/var/folders/...`, and `/var` is a symlink to `/private/var`, so `/var/folders/.../rescue.done` (the regex's truncated capture) still resolves to the right inode. On Linux this won't be the case — the truncated path won't exist, the `try/except OSError: pass` will swallow the failure, the wrapper will see no marker, and the `unsalvageable` / `structural` / `fixed` assertions in `test_rescue_phase_a.py` will all flip to `bail_reason == 'diagnosis_no_marker'`.

Fix: drop the leading `\b` (and ideally the trailing one too, since the prompt may end the path with non-word punctuation). E.g. `r'(/\S+?\.done)\b'` or, more robustly, anchor on the surrounding line context — the prompt always emits the path on its own indented line.

pipeline/tests/fixtures/fake_claude.py
xbrianh: `time` is imported but never used.

pipeline/tests/fixtures/fake_claude.py
xbrianh, line 325: The `if stage == 'rescue-diagnosis':` branch on lines 333-334 returns the same expression as the fallthrough on line 335 — it can be deleted.

pipeline/tests/fixtures/shell_env.py
xbrianh: `Optional` is imported but never used.

pipeline/tests/fixtures/shell_env.py
xbrianh, line 117: The comment is misleading: `GIT_OPTIONAL_LOCKS=0` disables optional `.lock` files for read-only commands (e.g. `git status` skipping the index lock); it does not suppress git's `hint:` output. To quiet hints you'd configure `advice.*=false`. If you only meant to suppress lock contention from concurrent test runs the env var is fine — just retitle the comment.

pipeline/tests/fixtures/shell_env.py
xbrianh: `__import__('json').loads(line)` here and the `import json` inside `read_state` on line 148 can both be replaced with a single top-level `import json` next to the other imports.

pipeline/tests/fixtures/fake_gh.py
xbrianh: Real `gh issue create` writes only the URL to stdout — no `Creating issue` preamble. If a future caller does `URL=$(gh issue create …)` rather than `tail -1`, this stub will return `"Creating issue\n<url>"` and break. Recommend matching the real CLI: just emit `f"{url}\n"`.

pipeline/tests/test_ghgremlin_shell.py
xbrianh, line 179: This loop only checks the model flag on stages that *did* run — if (e.g.) `scope-review-pr` were silently skipped, the test would still pass. Consider adding an upfront `assert pipeline_stages.issubset({e['stage'] for e in log}), …` so the test fails fast when a stage in the expected set never ran.

pipeline/tests/test_ghgremlin_shell.py
xbrianh, line 184: Naming nit: this test is named `test_ghgremlin_resume_from_ghreview` but the resume itself is performed by invoking `ghgremlin.sh` directly (line 206) rather than `launch.sh --resume`. The comment on lines 201-204 acknowledges this, but the test name still suggests the launcher's resume path is exercised. Consider `test_pipeline_resume_from_ghreview` (or add a sibling test that hits `launch.sh --resume <id>` with a `--resume-from` injected via the rescue path) so the launcher's resume contract isn't left untested by this layer.

xbrianh (commented):
Solid layer of shell-integration coverage that closes a real gap (subprocess pathway, CLAUDE_FLAGS forwarding, --resume rehydration). The fixtures are well-structured. One concrete bug worth fixing before merge: the marker-path regex in `fake_claude.py` strips a leading word-prefixed path component because `\b` cannot match between whitespace and `/`. On macOS this happens to work (`/var` symlinks to `/private/var`), but on Linux CI (`/tmp/...`) the regex would yield `/pytest-of-…/…/rescue.done`, the marker write would fail silently, and every `unsalvageable`/`structural`/`fixed` rescue assertion would flip to `diagnosis_no_marker`. A few smaller cleanups noted inline.

pipeline/tests/fixtures/shell_env.py
```diff
@@ -0,0 +1,149 @@
...
+        **os.environ,
+        "HOME": str(home),
+        "XDG_STATE_HOME": str(state_root),
+        "PATH": f"{bin_dir}{os.pathsep}{os.environ.get('PATH', '')}",
+        "FAKE_CLAUDE_LOG": str(fake_claude_log),
+        "FAKE_GH_LOG": str(fake_gh_log),
+        # Strip caller's PYTHONPATH; launch.sh will set its own.
+    }
+    env.pop("PYTHONPATH", None)
+    # Suppress git's hint output so tests get clean stderr.
```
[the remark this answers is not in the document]
xbrianh: Updated the comment to describe the actual rationale (reduce lock contention across parallel test runs).

pipeline/tests/fixtures/fake_claude.py
```diff
@@ -0,0 +1,339 @@
...
+        "ghreview": lambda p, s: (emit_minimal_stream(s), 0)[1],
+        "ghaddress": lambda p, s: (emit_minimal_stream(s), 0)[1],
+        "scope-review-pr": lambda p, s: (emit_minimal_stream(s), 0)[1],
+        "rescue-diagnosis": lambda p, s: handle_rescue_diagnosis(p),
+    }
+    h = handlers.get(stage)
+    if h is None:
+        sys.stderr.write(f"fake claude: unrecognized stage for prompt head: {prompt[:120]!r}\n")
+        emit_minimal_stream(session_id)
+        return 0
```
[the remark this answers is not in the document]
xbrianh: Adopted: unknown stage now exits 1 by default; gated permissive behavior on FAKE_CLAUDE_ALLOW_UNKNOWN_STAGE=1 for any caller that needs to opt in.

pipeline/tests/fixtures/fake_gh.py
```diff
@@ -0,0 +1,100 @@
...
+import sys
+
+
+def log_invocation(argv) -> None:
+    log_path = os.environ.get("FAKE_GH_LOG")
+    if not log_path:
+        return
+    try:
+        with open(log_path, "a", encoding="utf-8") as fh:
+            fh.write(json.dumps({"argv": list(argv)}) + "\n")
```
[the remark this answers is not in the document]
xbrianh: Switched to atomic os.write append in fake_gh log_invocation.

pipeline/tests/test_ghgremlin_shell.py
```diff
@@ -0,0 +1,255 @@
...
+    """Invoke the ghgremlin.sh shim directly (no launch.sh, no background).
+
+    The shim execs `python -m pipeline.cli gh "$@"`, so this exercises the
+    arg-forwarding contract end-to-end without the worktree+state machinery.
+    """
+    env = {
+        **sh.env,
+        "PYTHONPATH": f"{PIPELINE_PARENT}{':' + sh.env['PYTHONPATH'] if 'PYTHONPATH' in sh.env else ''}",
+    }
+    return subprocess.run(
```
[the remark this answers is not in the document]
xbrianh: Switched to os.pathsep when extending PYTHONPATH.

pipeline/tests/test_localgremlin_shell.py
```diff
@@ -0,0 +1,103 @@
...
+
+Drives the full plan → implement → review (parallel) → address chain
+via ``launch.sh localgremlin`` with a fake `claude` on PATH and a real
+git repo. Verifies stage ordering and per-stage artifacts on disk.
+"""
+
+from __future__ import annotations
+
+import json
+import pathlib
```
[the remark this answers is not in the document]
xbrianh: Removed unused json/pathlib/pytest imports.

pipeline/tests/test_launch_sh.py
```diff
@@ -0,0 +1,268 @@
...
+rejection.
+"""
+
+from __future__ import annotations
+
+import json
+import os
+import pathlib
+import subprocess
+import time
```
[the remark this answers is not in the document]
xbrianh: Removed unused time and pytest imports.

pipeline/tests/test_rescue_phase_a.py
```diff
@@ -0,0 +1,266 @@
...
+"""
+
+from __future__ import annotations
+
+import importlib.util
+import json
+import os
+import pathlib
+import subprocess
+import sys
```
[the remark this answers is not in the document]
xbrianh: Removed unused os/subprocess/sys/pytest imports.

pipeline/tests/fixtures/shell_env.py
```diff
@@ -0,0 +1,149 @@
...
+
+
+def install_fake_bin(bin_dir: pathlib.Path, name: str, target: pathlib.Path) -> None:
+    bin_dir.mkdir(parents=True, exist_ok=True)
+    wrapper = bin_dir / name
+    wrapper.write_text(
+        f"#!/usr/bin/env bash\nexec {sys.executable} {target} \"$@\"\n",
+        encoding="utf-8",
+    )
+    wrapper.chmod(0o755)
```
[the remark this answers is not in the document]
xbrianh: Wrapped both paths in shlex.quote so spaces in sys.executable or fixture targets do not break exec.

pipeline/tests/fixtures/fake_claude.py
```diff
@@ -0,0 +1,339 @@
...
+
+def log_invocation(record: dict) -> None:
+    log_path = os.environ.get("FAKE_CLAUDE_LOG")
+    if not log_path:
+        return
+    try:
+        with open(log_path, "a", encoding="utf-8") as fh:
+            fh.write(json.dumps(record) + "\n")
+    except OSError:
+        pass
```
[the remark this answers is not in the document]
xbrianh: Switched to atomic os.O_APPEND + os.write so the 3-way parallel review workers cannot interleave log lines.

pipeline/tests/test_ghgremlin_shell.py
```diff
@@ -0,0 +1,255 @@
...
+- ``--plan <issue-ref>`` resolves an existing issue and runs the chain
+- ``--model`` reaches every claude invocation
+- ``--resume-from <stage>`` skips earlier stages
+- unknown leading flags abort before any state is created
+"""
+
+from __future__ import annotations
+
+import json
+import pathlib
```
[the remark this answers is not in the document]
xbrianh: Removed unused json/pathlib/pytest imports.

pipeline/tests/test_rescue_phase_a.py
```diff
@@ -0,0 +1,266 @@
...
+        "kind": "localgremlin",
+        "stage": "implement",
+        "status": "stopped",
+        "exit_code": 1,
+        "workdir": str(workdir),
+        "project_root": str(workdir.parent),
+        "description": "test gremlin",
+        "started_at": "2026-04-27T00:00:00Z",
+        "rescue_count": 0,
+    }
```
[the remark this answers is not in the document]
xbrianh: Switched started_at to datetime.datetime.now(datetime.timezone.utc) at runtime.

pipeline/tests/fixtures/fake_claude.py
xbrianh: Fixed: dropped the leading \b so the captured marker_path is no longer truncated. Added an explicit comment about why \b cannot match between whitespace and /.

xbrianh: Fixed: dropped the leading \b so the captured marker_path is no longer truncated. Added an explicit comment about why \b cannot match between whitespace and /.

pipeline/tests/fixtures/fake_claude.py
xbrianh: Removed unused time import.

xbrianh: Removed unused time import.

pipeline/tests/fixtures/fake_claude.py
xbrianh: Removed the redundant rescue-diagnosis branch in main().

xbrianh: Removed the redundant rescue-diagnosis branch in main().

pipeline/tests/fixtures/shell_env.py
xbrianh: Removed unused Optional import.

xbrianh: Removed unused Optional import.

pipeline/tests/fixtures/shell_env.py
xbrianh: Updated the comment to describe the actual behavior (reduce lock contention across parallel test runs).

xbrianh: Updated the comment to describe the actual behavior (reduce lock contention across parallel test runs).

pipeline/tests/fixtures/shell_env.py
xbrianh: Promoted json to a top-level import; both inline forms are gone.

xbrianh: Promoted json to a top-level import; both inline forms are gone.

pipeline/tests/fixtures/fake_gh.py
xbrianh: Updated fake gh issue create to emit only the URL.

xbrianh: Updated fake gh issue create to emit only the URL.

pipeline/tests/test_ghgremlin_shell.py
xbrianh: Added an upfront pipeline_stages.issubset(observed_stages) assertion in the model-forwarding test so a silently skipped stage fails fast.

xbrianh: Added an upfront pipeline_stages.issubset(observed_stages) assertion in the model-forwarding test so a silently skipped stage fails fast.

pipeline/tests/test_ghgremlin_shell.py
xbrianh: Out of scope for this PR — naming preference rather than a defect, and the comment in-place already calls out that the launch.sh resume path is not exercised here. Filing a sibling-test idea as follow-up if it survives a triage pass; not blocking this PR on it.

xbrianh: Out of scope for this PR — naming preference rather than a defect, and the comment in-place already calls out that the launch.sh resume path is not exercised here. Filing a sibling-test idea as follow-up if it survives a triage pass; not blocking this PR on it.

PR #138, opened 2026-04-27
ci: add GitHub Actions workflow for pipeline tests

### LINKED ISSUE #137: CI: run pipeline tests on PRs and pushes to main
# CI for pipeline tests

## Context

The `pipeline/` package has ~2,800 lines of pytest coverage (unit tests, orchestrator tests, and shell integration tests that drive `launch.sh` against a fake `claude` and a throwaway git repo). Today these tests only run when someone remembers to run them locally. Regressions can land on `main` unnoticed.

We want GitHub Actions to run the pipeline test suite on every PR and every push to `main`, and to gate merges on the result.

## Goals

- Every pull request targeting `main` runs the pipeline tests automatically.
- Every push to `main` runs the pipeline tests automatically.
- A failing test run blocks the PR from being merged.
- The status check is stable enough to wire into branch protection.

## Non-goals

- Running anything other than the `pipeline/` test suite (skills, scripts, sync) — out of scope for this spec.
- Cross-version Python matrix — single interpreter only.
- Cross-OS matrix — Linux only.
- Coverage reporting, benchmarking, lint, type-check — out of scope.
- Caching/speed optimization beyond what's trivially built in.
- Auto-merge, auto-label, auto-comment behavior on PRs.

## Users

- Repo maintainers reviewing PRs — see a green/red check next to the PR before merging.
- PR authors (including background gremlins opening PRs) — get fast feedback that their change didn't break the pipeline.

## Behavior

- On every PR opened or updated against `main`, GitHub Actions runs the pipeline test suite and reports a single status check on the PR.
- On every push to `main`, the same suite runs and reports status on the commit.
- The job runs on a Linux runner with `bash`, `git`, and Python 3.9 available. Tests provide their own fake `claude`/`gh` binaries; no external network or secrets needed.
- The job installs whatever the test suite needs (pytest + the `pipeline` package itself) and runs the full suite from `pipeline/tests/`.
- A non-zero pytest exit reports the check as failed.
- The status check has a stable, predictable name that can be set as a required check in branch protection.
- The workflow runs on every PR and every push to `main` unconditionally — no path filtering, even for doc-only changes.

## Constraints

- Python 3.9 (matches `pyproject.toml`'s `requires-python`).
- Linux runner (`ubuntu-latest` or equivalent).
- Workflow definition lives in the repo and is itself subject to the sync rules in `CLAUDE.md` if it falls under a tracked path; otherwise it lives only in the repo (`.github/` is not currently tracked by `scripts/sync.sh`).
- No secrets required at runtime.
- Test runtime budget: should complete in a few minutes on a standard hosted runner; if it doesn't, that's a signal to revisit, not a blocker for this spec.

## Acceptance criteria

- Opening a PR against `main` triggers the workflow and surfaces a status check on the PR.
- Pushing a commit to `main` triggers the workflow and surfaces a status check on the commit.
- A deliberately-broken test causes the check to fail red; reverting it makes the next run pass green.
- The check appears in the branch-protection "required status checks" picker by a stable name, and once selected, a failing run blocks the PR merge button.

## Open questions

- Should the workflow also run on `workflow_dispatch` (manual trigger) for debugging? Default: yes, cheap to add.
- Branch-protection rules themselves are configured in the GitHub UI, not in code — out of scope for this spec, but the implementer should flag when the check is ready to be wired up.

xbrianh:
## Summary
- Add `.github/workflows/pipeline-tests.yml` that runs `pytest` in `pipeline/` on every PR targeting `main`, every push to `main`, and on manual `workflow_dispatch`.
- Linux runner (`ubuntu-latest`) with Python 3.9, matching `pipeline/pyproject.toml`'s `requires-python`.
- Job name is the stable string `pipeline-tests` so it can be wired into branch-protection required-checks without churn.
- Test command (`cd pipeline && PYTHONPATH=. python -m pytest`) matches what `pipeline/CLAUDE.md` documents; tests provide their own fake `claude`/`gh` binaries, so no secrets or network are required.

## Notes
- `.github/` is not in `scripts/sync.sh`'s `DIR_PAIRS`, so the workflow lives only in this repo and is not mirrored to `~/.claude/`.
- Branch-protection wiring (marking `pipeline-tests` as a required check) is a GitHub UI step and out of scope for this PR — ready to be wired up once this lands.

## Test plan
- [ ] Workflow triggers on this PR and surfaces a `pipeline-tests` status check.
- [ ] Check goes green when tests pass.
- [ ] Verify a deliberately-broken test would fail the check red (manual one-off, not part of this PR).
- [ ] After merge, push to `main` triggers the workflow on the merge commit.
- [ ] `pipeline-tests` appears in branch-protection required-checks picker.

Closes #137

### DIFF
--- a/.github/workflows/pipeline-tests.yml
+++ b/.github/workflows/pipeline-tests.yml
@@ -7,8 +7,21 @@ on:
     branches: [main]
   workflow_dispatch:
 
+# Cancel superseded PR runs; keep main-branch runs independent so push-to-main
+# CI history has no gaps.
+concurrency:
+  group: pipeline-tests-${{ github.ref }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
+
+# Least-privilege: this workflow only checks out code and runs tests.
+permissions:
+  contents: read
+
 jobs:
   pipeline-tests:
+    # Explicit `name:` locks the displayed status-check name to `pipeline-tests`
+    # for branch-protection required-checks wiring, even if the job key is
+    # later renamed.
     name: pipeline-tests
     runs-on: ubuntu-latest
     timeout-minutes: 15
@@ -20,7 +33,9 @@ jobs:
           python-version: '3.9'
 
       - name: Install pytest
-        run: python -m pip install --upgrade pip pytest
+        # Pin pytest below 8.4 — pytest 8.4+ drops Python 3.9 support, which
+        # would silently break this workflow once released.
+        run: python -m pip install --upgrade pip "pytest<8.4"
 
       - name: Run pipeline tests
         working-directory: pipeline
--- a/.github/workflows/pipeline-tests.yml
+++ b/.github/workflows/pipeline-tests.yml
@@ -1,0 +1,27 @@
+name: pipeline-tests
+
+on:
+  pull_request:
+    branches: [main]
+  push:
+    branches: [main]
+  workflow_dispatch:
+
+jobs:
+  pipeline-tests:
+    name: pipeline-tests
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: actions/setup-python@v5
+        with:
+          python-version: '3.9'
+
+      - name: Install pytest
+        run: python -m pip install --upgrade pip pytest
+
+      - name: Run pipeline tests
+        working-directory: pipeline
+        run: PYTHONPATH=. python -m pytest


xbrianh (commented):

.github/workflows/pipeline-tests.yml
```diff
@@ -0,0 +1,27 @@
+name: pipeline-tests
+
+on:
+  pull_request:
+    branches: [main]
+  push:
+    branches: [main]
+  workflow_dispatch:
+
+jobs:
```
xbrianh, line 20: Optional: consider adding a `concurrency:` block at the workflow level so a new push to a PR cancels the previous in-flight run. Without it, rapid pushes to a branch queue N independent runs that all finish even though only the latest matters. Suggestion:

```yaml
concurrency:
  group: pipeline-tests-${{ github.ref }}
  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
```

Guarding `cancel-in-progress` on `pull_request` keeps `push: main` runs from cancelling each other (which would leave gaps in main's CI history). Not blocking — just a CI-minutes / latency win.

.github/workflows/pipeline-tests.yml
```diff
@@ -0,0 +1,27 @@
...
+  pipeline-tests:
+    name: pipeline-tests
+    runs-on: ubuntu-latest
```
xbrianh, line 26: Optional hardening: set an explicit minimal `permissions:` block. With no `permissions:` declared anywhere, this job inherits the repo-default `GITHUB_TOKEN` scopes, which on many repos is `contents: write` or broader. This workflow only checks out code and runs tests — it never needs to write. GitHub's published guidance is to declare least-privilege per workflow:

```yaml
permissions:
  contents: read
```

at the workflow level (after `on:`, before `jobs:`). Defense-in-depth against a future supply-chain compromise of `setup-python` or pip. Not blocking.

.github/workflows/pipeline-tests.yml
```diff
@@ -0,0 +1,27 @@
...
+on:
+  pull_request:
+    branches: [main]
+  push:
+    branches: [main]
+  workflow_dispatch:
+
+jobs:
+  pipeline-tests:
+    name: pipeline-tests
```
xbrianh, line 25: Minor: `name: pipeline-tests` here is redundant with the job key `pipeline-tests:` on line 11 (the `name:` field defaults to the key when omitted). I'd actually **keep it** — the PR description specifically calls out wiring this into branch-protection required-checks, and having an explicit `name:` field locks the displayed status-check name to `pipeline-tests` even if someone later renames the job key. Worth a comment in the file noting that, if you want to make the intent durable. No change required.

xbrianh (commented):
Small, focused workflow that does exactly what the PR description claims: it matches `pipeline/CLAUDE.md`'s test command (`cd pipeline && PYTHONPATH=. python -m pytest`) and `pyproject.toml`'s `requires-python = '>=3.9'`. Tests already vendor fake `claude`/`gh` binaries via fixtures, so no secrets/network needed — good. Action versions are current (`checkout@v4`, `setup-python@v5`) and the 15-minute timeout is reasonable.

No blockers. Three non-blocking suggestions inline: a `concurrency:` group to cancel superseded PR runs, an explicit minimal `permissions:` block (defense-in-depth — this job needs only `contents: read`), and a note on the cosmetically-redundant `name:` field on the job (which I'd actually keep, see comment).

.github/workflows/pipeline-tests.yml
xbrianh: Added the workflow-level `concurrency:` block as suggested, with `cancel-in-progress` guarded on `pull_request` so push-to-main runs stay independent.

.github/workflows/pipeline-tests.yml
xbrianh: Added `permissions: contents: read` at the workflow level.

.github/workflows/pipeline-tests.yml
xbrianh: Kept the explicit `name:` field as you suggested, and added an inline comment recording the branch-protection-pinning intent so it survives future edits.

.github/workflows/pipeline-tests.yml
```diff
@@ -0,0 +1,27 @@
...
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: actions/setup-python@v5
```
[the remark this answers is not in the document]
xbrianh: Filed as #139 — out of scope here since it's a policy that should apply to every workflow we add (and the suggested SHAs need to be re-resolved against the actions' published tags before applying), so handling as a follow-up sweep.

.github/workflows/pipeline-tests.yml
```diff
@@ -0,0 +1,27 @@
...
+        with:
+          python-version: '3.9'
+
+      - name: Install pytest
+        run: python -m pip install --upgrade pip pytest
```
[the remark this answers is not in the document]
xbrianh: Pinned to `pytest<8.4` and added an inline comment noting that pytest 8.4+ drops Python 3.9 support.

.github/workflows/pipeline-tests.yml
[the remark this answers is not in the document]
xbrianh: Added `permissions: contents: read` at the workflow level.

PR #145, opened 2026-04-27
pipeline/gh: default --model to sonnet

### LINKED ISSUE #144: ghgremlin orchestrator defaults --model to None, falls through to caller's runtime model (opus)
## Symptom

A `ghgremlin` invoked with no explicit `--model` runs every stage on the calling Claude Code session's default model — **not** sonnet, despite that being the convention everywhere else in the pipeline.

Captured live on gremlin `fix-gr-id-leak-from-pytest-in-pipeline-a0433e` (and prior gremlins). Every stage init in the log shows:

```
[implement]       init session=... model=claude-opus-4-7[1m]
[commit-pr]       init session=... model=claude-opus-4-7[1m]
[scope-review-pr] init session=... model=claude-opus-4-7[1m]
[ghreview]        init session=... model=claude-opus-4-7[1m]
[ghaddress]       init session=... model=claude-opus-4-7[1m]
```

The user's expectation — and the convention enforced by the other orchestrators — is sonnet.

## Root cause

`pipeline/orchestrators/gh.py` is the only orchestrator that doesn't pin a default model:

| Orchestrator | `--model` default | Source |
|---|---|---|
| `local.py` | `"sonnet"` (per-stage: plan / impl / address / holistic / detail / scope) | lines 90–95, 295–297, 386 |
| `boss.py` | `"sonnet"` | line 489 |
| `handoff.py` | `"sonnet"` | line 236 |
| **`gh.py`** | **`None`** | **line 90** |

When `args.model` is `None`, `gh.py` propagates `None` into `pipeline.clients.claude.SubprocessClaudeClient.run(..., model=None, ...)`, which at `claude.py:215` does:

```python
if model is not None:
    cmd += ["--model", model]
```

— so `claude -p` is invoked with **no `--model` argument at all**, and claude inherits the calling session's model. For sessions defaulting to opus 4.7, every stage runs on opus.

## Why this is a regression

The shell-original `skills/ghgremlin/ghgremlin.sh` (added in PR #52) treated `MODEL=""` the same way (no default), so the Python port preserved the existing behavior literally. But the user-visible expectation, set by `localgremlin` and `bossgremlin` defaulting to sonnet, was that `ghgremlin` would too. The default-`None` is now exposed as a real cost regression because the calling session defaults to opus.

This is also a cross-orchestrator inconsistency: a `bossgremlin --chain-kind gh` (which defaults `--model` to sonnet) will spawn ghgremlin children that happily run on opus when launched standalone.

## Proposed fix

Change `pipeline/orchestrators/gh.py:90` from:

```python
parser.add_argument("--model", dest="model", default=None)
```

to:

```python
parser.add_argument("--model", dest="model", default="sonnet")
```

Side-effects to check:

1. **Resume path** — `gh.py:299–303` reloads `model` from `state.json` when `args.model is None`. With the new default, `args.model` is never None, so resume would always overwrite with `"sonnet"` even when the original run had a different model persisted. Fix: change the sentinel to `"sonnet"` for the resume check, **or** keep `default=None` in argparse and inject the `"sonnet"` fallback only after the resume-restore step. The latter is the cleaner fix and matches existing semantics.
2. **`--model` plumbing into title generation** — verify the title-gen call site at `gh.py` (and the `_resolve_plan_source` call) gets the resolved model, not the raw argparse value.

## Regression test

Add a test to `pipeline/tests/test_orchestrator_gh.py` mirroring the existing local-orchestrator pattern at `test_orchestrator_local.py:123`:

```python
def test_gh_main_defaults_model_to_sonnet(tmp_path, monkeypatch):
    """Regression (#<this issue>): ghgremlin must default --model to sonnet,
    not fall through to claude's runtime default."""
    # Standard fakes: no GH network, no real git push, etc.
    _common_patches(monkeypatch)
    monkeypatch.setenv("XDG_STATE_HOME", str(tmp_path / "xdg"))

    client = FakeClaudeClient(fixtures={
        "implement":   MINIMAL_EVENTS,
        "commit-pr":   MINIMAL_EVENTS,
        "ghreview":    MINIMAL_EVENTS,
        "scope-review-pr": MINIMAL_EVENTS,
        "ghaddress":   MINIMAL_EVENTS,
        # plus whatever the existing happy-path test in this file already wires.
    })

    # Invoke with NO --model. Use a --plan local-file source so we don't
    # need to fake `gh issue view`.
    plan = tmp_path / "plan.md"
    plan.write_text("# Plan\n\nDo the thing.\n")
    rc = gh_main(["--plan", str(plan)], client=client)
    assert rc == 0

    # Every recorded run must have model == "sonnet". Asserting on every call
    # (not just calls[0]) catches the case where one stage is fixed but
    # another is overlooked.
    assert client.calls, "expected at least one client call"
    bad = [c for c in client.calls if c.model != "sonnet"]
    assert not bad, (
        f"{len(bad)} stage(s) ran on a non-sonnet model: "
        f"{[(c.label, c.model) for c in bad]}"
    )
```

Two failure modes this catches:

- The argparse default reverts to `None` — every recorded call comes through with `model=None`, test fails with the model-mismatch assertion.
- A new stage is added later that accidentally hard-codes a different model or drops the orchestrator's resolved model — test fails on that specific stage.

The regression test should also have a positive companion that confirms `--model opus` *does* propagate (so we don't accidentally pin sonnet so hard that the override is broken).

## Out of scope

- Whether `localgremlin`'s six per-stage model dials should also exist on `ghgremlin` (e.g. cheap model for `commit-pr`, more powerful for `implement`). Separate config-surface decision.
- Whether the default should be `"haiku"` instead of `"sonnet"` for cost reasons — separate.

xbrianh:
## Summary

- `pipeline/orchestrators/gh.py` was the only orchestrator without a `--model` default, so `claude -p` was invoked with no `--model` flag and inherited the calling session's model (typically opus). Every stage of a `ghgremlin` run silently ran on opus, contrary to the sonnet convention used by `localgremlin` and `bossgremlin`.
- Argparse keeps `default=None` so the resume path can still detect "user didn't pass `--model`" and prefer the persisted value over the fresh default; the `"sonnet"` fallback is injected only after the resume-restore step.
- Adds a regression test mirroring the local-orchestrator pattern: invokes `gh_main` without `--model` and asserts every recorded `client.run` call has `model == "sonnet"`. The pre-existing `test_model_forwarded_to_all_stages` already covers the positive companion (explicit `--model` propagates).

## Test plan

- [x] `pytest pipeline/tests/test_orchestrator_gh.py` (20 passed, including the new `test_gh_main_defaults_model_to_sonnet`)
- [x] Full `pipeline/tests/` suite (164 passed)

Closes #144

### DIFF
--- a/pipeline/tests/test_orchestrator_gh.py
+++ b/pipeline/tests/test_orchestrator_gh.py
@@ -464,6 +464,48 @@ def test_model_forwarded_to_all_stages(tmp_path, monkeypatch):
         assert call.model == "claude-opus-4-7", f"stage {call.label!r} got model={call.model!r}"
 
 
+def test_gh_main_defaults_model_to_sonnet(tmp_path, monkeypatch):
+    """Regression: ghgremlin must default --model to sonnet, not fall through
+    to claude's runtime default (which inherits the calling session's model
+    and silently runs every stage on opus when launched from an opus session).
+    """
+    _init_git_repo(tmp_path)
+    monkeypatch.chdir(tmp_path)
+
+    session_dir, state_file = _patch_common(monkeypatch, tmp_path)
+
+    monkeypatch.setattr(
+        subprocess, "run",
+        _make_gh_subprocess(issue_body="# Plan\nDo stuff.\n"),
+    )
+    monkeypatch.setattr("pipeline.orchestrators.gh.run_ghreview_stage", lambda **kw: None)
+    monkeypatch.setattr("pipeline.orchestrators.gh.run_wait_copilot_stage", lambda **kw: "APPROVED")
+    monkeypatch.setattr("pipeline.orchestrators.gh.run_request_copilot_stage", lambda **kw: None)
+    monkeypatch.setattr("pipeline.orchestrators.gh.run_ghaddress_stage", lambda **kw: None)
+
+    client = _CommittingClient(
+        git_dir=tmp_path,
+        fixtures={
+            "implement": IMPL_EVENTS,
+            "commit-pr": _pr_events(),
+        },
+    )
+
+    # Invoke with NO --model.
+    result = gh_main(["--plan", "42"], client=client)
+    assert result == 0
+
+    # Every recorded run must have model == "sonnet". Asserting on every call
+    # (not just calls[0]) catches the case where one stage is fixed but
+    # another is overlooked.
+    assert client.calls, "expected at least one client call"
+    bad = [c for c in client.calls if c.model != "sonnet"]
+    assert not bad, (
+        f"{len(bad)} stage(s) ran on a non-sonnet model: "
+        f"{[(c.label, c.model) for c in bad]}"
+    )
+
+
 def test_resume_from_implement(tmp_path, monkeypatch):
     """--resume-from implement reloads issue_url from state.json and runs implement onward."""
     _init_git_repo(tmp_path)
--- a/pipeline/tests/test_orchestrator_gh.py
+++ b/pipeline/tests/test_orchestrator_gh.py
@@ -506,6 +506,67 @@ def test_gh_main_defaults_model_to_sonnet(tmp_path, monkeypatch):
     )
 
 
+def test_gh_main_resume_prefers_persisted_model_over_sonnet_default(tmp_path, monkeypatch):
+    """Regression: on resume with no --model, a persisted state.json.model must
+    win over the fresh-launch "sonnet" fallback. Locks in the other half of the
+    invariant called out in gh.py — a future refactor that switched argparse to
+    default="sonnet" would silently break this precedence and only the
+    fresh-launch test would still pass.
+    """
+    _init_git_repo(tmp_path)
+    monkeypatch.chdir(tmp_path)
+
+    state_data = {
+        "issue_url": "https://github.com/owner/repo/issues/99",
+        "issue_num": "99",
+        "model": "claude-opus-4-7",
+    }
+    session_dir, state_file = _patch_common(monkeypatch, tmp_path, state_data=state_data)
+
+    monkeypatch.setattr(
+        subprocess, "run",
+        _make_gh_subprocess(issue_body="# Resumed Plan\nDo more stuff.\n"),
+    )
+    monkeypatch.setattr("pipeline.orchestrators.gh.run_ghreview_stage", lambda **kw: None)
+    monkeypatch.setattr("pipeline.orchestrators.gh.run_wait_copilot_stage", lambda **kw: "APPROVED")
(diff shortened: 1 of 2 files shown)


xbrianh (commented):

pipeline/tests/test_orchestrator_gh.py
```diff
@@ -464,6 +464,48 @@ def test_model_forwarded_to_all_stages(tmp_path, monkeypatch):
...
+
+    # Every recorded run must have model == "sonnet". Asserting on every call
+    # (not just calls[0]) catches the case where one stage is fixed but
+    # another is overlooked.
+    assert client.calls, "expected at least one client call"
+    bad = [c for c in client.calls if c.model != "sonnet"]
+    assert not bad, (
+        f"{len(bad)} stage(s) ran on a non-sonnet model: "
+        f"{[(c.label, c.model) for c in bad]}"
+    )
```
xbrianh, line 506: Nice positive companion to `test_model_forwarded_to_all_stages`. Consider adding one more case to lock in the other half of the invariant the comment in `gh.py` calls out — i.e. that on resume, a persisted `state.json.model` is preferred over the new `"sonnet"` fallback.

Something like: launch with `--resume-from implement` (no `--model`) and a `_fake_read` that returns `"claude-opus-4-7"` for the `model` field, then assert every recorded call's `model == "claude-opus-4-7"`. Without that, a future refactor that switches argparse to `default="sonnet"` would silently break the resume-restore precedence and only `test_gh_main_defaults_model_to_sonnet` would still pass.

xbrianh (commented):
Fix is correct and surgical: keeping `default=None` in argparse and injecting `"sonnet"` after the resume-restore step preserves the resume-prefers-state ordering while closing the silent-opus-on-fresh-launch gap. The new comment explains the non-obvious invariant clearly, and the test mirrors the local-orchestrator pattern (asserting on every recorded call's `model`, not just `calls[0]`).

One minor observation: the new comment specifically claims the resume path prefers the persisted state value over the fresh-launch fallback, but no test currently locks that in — only the fresh-launch half is covered. Inline comment with a suggested companion test.

pipeline/tests/test_orchestrator_gh.py
xbrianh: Done in fd2b38c — added `test_gh_main_resume_prefers_persisted_model_over_sonnet_default`. Launches with `--resume-from implement` and no `--model`, with state.json + `_fake_read` returning `model="claude-opus-4-7"`, and asserts every recorded `client.run` call's `model == "claude-opus-4-7"`. Pairs with the fresh-launch test to lock in both halves of the invariant — a future refactor switching argparse to `default="sonnet"` would now fail this test. All 165 pipeline tests pass.

PR #146, opened 2026-04-27
ci: pin actions to full commit SHAs for supply-chain hardening

### LINKED ISSUE #139: Pin GitHub Actions to commit SHAs for supply-chain hardening
The `pipeline-tests` workflow currently pins third-party actions to major version tags (`actions/checkout@v4`, `actions/setup-python@v5`). A compromised upstream tag could affect CI runs. GitHub's hardening guidance recommends pinning to full commit SHAs (with a comment naming the tag for readability) so the action contents are content-addressed.

This applies to any future workflows we add as well, so worth handling as a small policy sweep rather than a one-line tweak in the original PR.

Reviewer comment (Copilot):
> For supply-chain hardening, consider pinning `actions/checkout` and `actions/setup-python` to full commit SHAs (instead of only major version tags). This reduces the risk of a compromised upstream action tag affecting CI runs.
> ```yaml
>       - uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
>       - uses: actions/setup-python@82c7e631bb3cdc910f68e0081d67478d79c6982d # v5.1.0
> ```

(Note: the SHAs in the suggestion above are reviewer-provided; resolve them against the action repo's published tags before applying.)

Ref #138
https://github.com/xbrianh/claude/pull/138#discussion_r3148706835

xbrianh:
Pins `actions/checkout` and `actions/setup-python` to full commit SHAs so a compromised upstream tag cannot affect CI runs. Tag names are kept as inline comments for readability.

- `actions/checkout` → `34e114876b0b11c390a56381ad16ebd13914f8d5` (v4.3.1)
- `actions/setup-python` → `a26af69be951a213d495a4c3e4e4022e16d87065` (v5.6.0)

Closes #139

### DIFF
--- a/.github/workflows/pipeline-tests.yml
+++ b/.github/workflows/pipeline-tests.yml
@@ -26,9 +26,9 @@ jobs:
     runs-on: ubuntu-latest
     timeout-minutes: 15
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
 
-      - uses: actions/setup-python@v5
+      - uses: actions/setup-python@a26af69be951a213d495a4c3e4e4022e16d87065 # v5.6.0
         with:
           python-version: '3.10'
 


xbrianh (commented):

xbrianh (commented):
Both SHAs verified against the upstream tags — `34e114876b0b11c390a56381ad16ebd13914f8d5` is exactly the commit `actions/checkout` points to for `v4.3.1`, and `a26af69be951a213d495a4c3e4e4022e16d87065` is the correct commit for `actions/setup-python` `v5.6.0`. The inline `# vX.Y.Z` comments preserve readability. Combined with the existing `permissions: contents: read` least-privilege block, this workflow is in good shape for supply-chain hardening. No issues found.

PR #159, opened 2026-04-29
fix: child gremlins inherit project_root from parent via --parent

### LINKED ISSUE #158: Fix child gremlins inheriting boss worktree as `project_root`, breaking `--here`
# Fix: child gremlins inherit boss's worktree as project_root, breaking `--here`

## Bug

Boss-launched child gremlins do not appear in `gremlins --here` (or any
other view that filters by the current repo). Their `state.json` records
`project_root` as the boss's worktree path (e.g. `/var/folders/.../aibg-bossgremlin.XXXXXX.xyz`)
instead of the user's actual repo (e.g. `/Users/.../my-project`).

Concretely, with a boss + child chain running while the user is in
`/Users/.../my-project`:

- `gremlins` (no flags) shows the boss and child.
- `gremlins --here` shows only the boss; the child is silently filtered out.

## Root cause

`skills/_bg/launch.sh` resolves `PROJECT_ROOT` from
`git rev-parse --show-toplevel` of `pwd` at the time launch.sh runs:

```bash
if PROJECT_ROOT=$(git -C "$(pwd)" rev-parse --show-toplevel 2>/dev/null); then
    IS_GIT=1
else
    PROJECT_ROOT=$(pwd)
    IS_GIT=0
fi
```

When the boss orchestrator spawns a child via
`gremlins/orchestrators/boss.py::launch_child`:

```python
proc = subprocess.Popen(cmd, stdout=subprocess.PIPE, text=True)
```

…there is no `cwd=` argument, so the child's `launch.sh` inherits the
boss's cwd. The boss runs inside its own worktree (launch.sh `cd`s into
`$WORKDIR` before exec'ing the gremlin entry point), so
`git -C "$(pwd)" rev-parse --show-toplevel` resolves to the boss's
worktree.

Then `gremlins --here` matches `state.project_root == git_toplevel(user_cwd)`
(see `gremlins/fleet.py::collect_rows`, the `--here filter` block), so the
child — whose `project_root` is the boss's worktree — gets filtered out.

## Fix

In `skills/_bg/launch.sh`, when `--parent <id>` is supplied, inherit
`project_root` from the parent's `state.json` rather than resolving via
`git` in the current cwd. The parent's `state.json` lives at:

```
${XDG_STATE_HOME:-$HOME/.local/state}/claude-gremlins/<parent-id>/state.json
```

…and its `.project_root` field is the user's actual repo (because the
boss was originally launched from the user's cwd). Using that as the
child's `project_root` correctly captures the semantic that a child's
project is its parent's project, and it works for any caller that uses
`--parent`, not only the boss orchestrator.

Fall back to the existing `git rev-parse` behavior when:

- `--parent` is not set (top-level launches keep current behavior); or
- the parent's `state.json` is missing / unreadable / has no
  `project_root` (defensive fallback so a corrupt parent state doesn't
  block child launches).

After resolving `PROJECT_ROOT` (whether inherited or computed),
re-derive `IS_GIT` by checking whether the resolved path is a git
repo. Don't assume the inherited path is a git repo just because the
parent recorded it — keep the determination consistent.

## Where to make the edit

`skills/_bg/launch.sh` around line 344 (the
`if PROJECT_ROOT=$(git -C "$(pwd)" rev-parse --show-toplevel ...)` block).
This file is mirrored to `~/.claude/skills/_bg/launch.sh` via
`scripts/sync.sh`, so edit the repo file and let sync handle propagation.

`PARENT_ID` is already parsed earlier in the script (line 47, populated
from `--parent` at line 64), so the inheritance check can read it
directly.

## Tests

Add a shell-integration test that exercises the inheritance path. Two
scenarios are sufficient:

1. **Inheritance happens.** Set up a fake parent state directory with a
   `state.json` containing `project_root = "/some/path"` (must be a real
   directory the test can create). Invoke launch.sh in a way that
   triggers the path that records `project_root` into the child's state
   (or short of running a real gremlin: refactor the resolution into a
   shape that's testable in isolation, e.g. expose it via a flag or a
   helper script — but only if needed; a black-box launch.sh test is
   preferred).
2. **Fallback works when parent state is missing.** With a `--parent`
   pointing at a non-existent gremlin id, launch.sh should fall back to
   `git rev-parse` and not die.

If a black-box launch.sh test turns out to be heavy to set up (it
requires populating GR_ID, the worktree base, etc.), an alternative is
to add a Python-level test that asserts the boss's `launch_child`
produces children whose `project_root` matches the boss's
`project_root`. That test would still need to cover the launch.sh
codepath end-to-end to actually catch the bug, so prefer the shell test
if feasible.

## Acceptance

- A boss + child scenario, running with the user in their repo: both
  the boss and the child appear in `gremlins --here`.
- The fleet-status row for the child reports the user's repo as its
  project (not the boss's worktree).
- Top-level (non-child) launches behave exactly as before — no change
  to `project_root` resolution when `--parent` is unset.
- Existing tests pass; the new test covers both the inheritance path
  and the fallback.

## Out of scope

- Refactoring `boss.py::launch_child` to pass `cwd=` to Popen. The
  centralized fix in launch.sh is preferred (works for any `--parent`
  caller, not just boss).
- Backfilling `project_root` on already-running or already-finished
  gremlins. This fix only affects newly-launched children.
- Reworking `--here`'s matching logic in `fleet.py`. The bug is at the
  child-launch end, not the filter end.

xbrianh:
## Summary

- When `launch.sh` is given `--parent <id>`, read `project_root` from the parent's `state.json` instead of resolving via `git rev-parse` in the current working directory
- The boss orchestrator runs inside its own worktree, so children were recording the boss's worktree path as `project_root`, causing them to be filtered out by `gremlins --here`
- Falls back to `git rev-parse` when `--parent` is unset, the parent state is missing, or the parent has no `project_root` field

## Test plan

- [ ] New test `test_launch_sh_child_inherits_project_root_from_parent`: verifies child's `project_root` matches parent's when parent state.json is present
- [ ] New test `test_launch_sh_child_falls_back_when_parent_state_missing`: verifies fallback to `git rev-parse` when parent id is non-existent
- [ ] All 13 `test_launch_sh.py` tests pass

Closes #158

### DIFF
--- a/gremlins/tests/test_launch_sh.py
+++ b/gremlins/tests/test_launch_sh.py
@@ -279,16 +279,50 @@ def test_launch_sh_child_inherits_project_root_from_parent(tmp_path):
         encoding="utf-8",
     )
 
+    # Invoke launch.sh from outside the repo so a broken implementation cannot
+    # accidentally match the parent's project_root via cwd-based fallback.
+    launch_cwd = tmp_path / "outside-repo"
+    launch_cwd.mkdir()
+
     r = _run_launch(
         sh.env, "--print-id", "--parent", parent_id,
         "localgremlin", "test child inheritance",
-        cwd=sh.repo, timeout=30,
+        cwd=launch_cwd, timeout=30,
     )
     assert r.returncode == 0, f"launch.sh failed: {r.stderr}"
     gr_id = r.stdout.strip()
 
     state = read_state(sh.state_root / "claude-gremlins" / gr_id / "state.json")
     assert state["project_root"] == str(parent_root)
+
+    wait_for_finished(sh.state_root / "claude-gremlins" / gr_id, timeout=60)
+
+
+def test_launch_sh_child_falls_back_when_parent_root_nonexistent(tmp_path):
+    """Parent state.json has project_root but the path no longer exists; falls back to git rev-parse."""
+    sh = setup_shell_env(tmp_path)
+
+    # Create a parent state with a project_root that points to a deleted directory.
+    parent_id = "fake-parent-gone-aabbcc"
+    parent_state_dir = sh.state_root / "claude-gremlins" / parent_id
+    parent_state_dir.mkdir(parents=True)
+    gone_root = tmp_path / "deleted-repo"  # deliberately never created
+    (parent_state_dir / "state.json").write_text(
+        json.dumps({"id": parent_id, "project_root": str(gone_root)}),
+        encoding="utf-8",
+    )
+
+    r = _run_launch(
+        sh.env, "--print-id", "--parent", parent_id,
+        "localgremlin", "test nonexistent root fallback",
+        cwd=sh.repo, timeout=30,
+    )
+    assert r.returncode == 0, f"launch.sh failed: {r.stderr}"
+    gr_id = r.stdout.strip()
+
+    state = read_state(sh.state_root / "claude-gremlins" / gr_id / "state.json")
+    # Should fall back to git rev-parse of sh.repo (the launch cwd).
+    assert state["project_root"] == str(sh.repo)
 
     wait_for_finished(sh.state_root / "claude-gremlins" / gr_id, timeout=60)
 
--- a/gremlins/tests/test_launch_sh.py
+++ b/gremlins/tests/test_launch_sh.py
@@ -263,3 +263,50 @@ def test_launch_sh_localgremlin_empty_plan_file_rejected(tmp_path):
     )
     assert r.returncode != 0
     assert "empty" in (r.stderr + r.stdout)
+
+
+def test_launch_sh_child_inherits_project_root_from_parent(tmp_path):
+    """--parent causes the child's project_root to come from the parent's state.json."""
+    sh = setup_shell_env(tmp_path)
+
+    # Create a fake parent state directory with a known project_root.
+    parent_id = "fake-parent-aabbcc"
+    parent_state_dir = sh.state_root / "claude-gremlins" / parent_id
+    parent_state_dir.mkdir(parents=True)
+    parent_root = sh.repo  # a real directory so IS_GIT check works
+    (parent_state_dir / "state.json").write_text(
+        json.dumps({"id": parent_id, "project_root": str(parent_root)}),
+        encoding="utf-8",
+    )
+
+    r = _run_launch(
+        sh.env, "--print-id", "--parent", parent_id,
+        "localgremlin", "test child inheritance",
+        cwd=sh.repo, timeout=30,
+    )
+    assert r.returncode == 0, f"launch.sh failed: {r.stderr}"
+    gr_id = r.stdout.strip()
+
+    state = read_state(sh.state_root / "claude-gremlins" / gr_id / "state.json")
+    assert state["project_root"] == str(parent_root)
+
+    wait_for_finished(sh.state_root / "claude-gremlins" / gr_id, timeout=60)
+
+
+def test_launch_sh_child_falls_back_when_parent_state_missing(tmp_path):
+    """--parent with a non-existent parent id falls back to git rev-parse and succeeds."""
+    sh = setup_shell_env(tmp_path)
+
+    r = _run_launch(
+        sh.env, "--print-id", "--parent", "nonexistent-parent-id",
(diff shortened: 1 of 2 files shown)


skills/_bg/launch.sh
```diff
@@ -341,10 +341,25 @@ case "$KIND" in
...
+
+if [[ -n "$_parent_root" && -d "$_parent_root" ]]; then
+    PROJECT_ROOT="$_parent_root"
+elif PROJECT_ROOT=$(git -C "$(pwd)" rev-parse --show-toplevel 2>/dev/null); then
+    : # PROJECT_ROOT set
 else
     PROJECT_ROOT=$(pwd)
+fi
+
+if git -C "$PROJECT_ROOT" rev-parse --show-toplevel &>/dev/null; then
```
xbrianh, line 360: The `IS_GIT` check now correctly runs against `$PROJECT_ROOT` rather than `$(pwd)`. The original code implicitly coupled git-detection to the working directory, so if `PROJECT_ROOT` came from the parent (a different path), `IS_GIT` would have reflected the child's cwd, not the actual project root. This separation is the right shape.

One unreachable-but-harmless note: the `|| die` guard at line 365 can't fire in practice because the `else` branch always assigns `$(pwd)`, which is always a valid directory. Not wrong to have it.

gremlins/tests/test_launch_sh.py
```diff
@@ -263,3 +263,50 @@ def test_launch_sh_localgremlin_empty_plan_file_rejected(tmp_path):
...
+    r = _run_launch(
+        sh.env, "--print-id", "--parent", parent_id,
+        "localgremlin", "test child inheritance",
+        cwd=sh.repo, timeout=30,
+    )
+    assert r.returncode == 0, f"launch.sh failed: {r.stderr}"
+    gr_id = r.stdout.strip()
+
+    state = read_state(sh.state_root / "claude-gremlins" / gr_id / "state.json")
+    assert state["project_root"] == str(parent_root)
```
xbrianh, line 296: State is read before `wait_for_finished` — this is safe because `state.json` is written synchronously in launch.sh before the pipeline is forked to the background. Consistent with the existing test pattern.

One gap worth noting: there's no test for the case where the parent's `state.json` *has* a `project_root` field but the path no longer exists on disk. The code handles it correctly (the `-d` check falls through to `git rev-parse`), but a test for that path would give complete coverage of the three-branch logic.

xbrianh (commented):
Clean fix. The `IS_GIT` refactor (now checking `$PROJECT_ROOT` instead of `$(pwd)`) is a correctness improvement that fell out naturally from the change. No blockers — one minor gap in test coverage noted below.

skills/_bg/launch.sh
xbrianh: Agreed — the IS_GIT refactor is a correctness improvement that fell out naturally. On the `|| die` guard: noted, it's harmless defensive code and I've left it in place.

gremlins/tests/test_launch_sh.py
xbrianh: Added `test_launch_sh_child_falls_back_when_parent_root_nonexistent` — it creates a parent state.json with a `project_root` pointing to a deleted directory, then verifies the child falls back to `git rev-parse` of the launch cwd. All three branches of the logic now have test coverage.

gremlins/tests/test_launch_sh.py
```diff
@@ -263,3 +263,50 @@ def test_launch_sh_localgremlin_empty_plan_file_rejected(tmp_path):
...
+    parent_root = sh.repo  # a real directory so IS_GIT check works
+    (parent_state_dir / "state.json").write_text(
+        json.dumps({"id": parent_id, "project_root": str(parent_root)}),
+        encoding="utf-8",
+    )
+
+    r = _run_launch(
+        sh.env, "--print-id", "--parent", parent_id,
+        "localgremlin", "test child inheritance",
+        cwd=sh.repo, timeout=30,
```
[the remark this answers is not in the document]
xbrianh: Fixed — the inheritance test now launches from `tmp_path / 'outside-repo'` (a non-git directory), so cwd-based fallback would produce a different `project_root` than the parent's `sh.repo`. The test is now meaningful.

PR #163, opened 2026-04-29
feat: detect and recover from host-terminated gremlins

### LINKED ISSUE #160: Gremlins should be recoverable after external process termination
## Problem

When a gremlin's host environment is terminated externally (e.g. a VM is killed, machine reboots, OOM kill of the launcher), the pipeline process dies without writing a terminal `status` to `state.json`. If the worktree directory also disappears (ephemeral disk, manual cleanup, or the boss already cleaned it up), there is no path back: `/gremlins rescue` requires the worktree (the headless-rescue agent must edit files there and resume from a stage), and `liveness.sh` only classifies the gremlin as `dead:crashed (pid gone)` or `stalled:no log update` without distinguishing the worktree-gone case.

I just hit this — a batch of gremlins was running in a VM that was terminated and they were left wedged: not finished, can't be rescued, can't be resumed.

## What we want

The fleet should detect and recover from external termination as a first-class failure mode, not just as an unrescuable `structural`/`unsalvageable` dead end. Concrete capabilities to consider:

1. **Detect the case explicitly.** When `pid` is gone and the worktree path no longer exists (or the branch is missing), surface a distinct liveness label like `dead:host-terminated` or `dead:worktree-gone` so it isn't conflated with a normal crash.
2. **Allow resumption from artifacts alone.** All durable state (plan, review markdown, state.json) already lives under `~/.local/state/claude-gremlins/<id>/`. A recovery path could re-create the worktree from the recorded `branch` + `base_ref` (or fall back to `base_ref` if the branch is gone) and restart the failed stage, since the stages are designed to be idempotent under `--resume`.
3. **Graceful `rm`/`close` for unrecoverable cases.** If the branch and any in-progress commits are also gone, make it obvious to the operator and let `/gremlins rm` clean up without manual intervention. Today `rm` works, but the operator first has to figure out that recovery isn't possible.

## Reproduction

1. Start one or more gremlins via `/localgremlin` or `/ghgremlin`.
2. While they're mid-stage, kill the launcher process and `rm -rf` the worktree directory (simulating VM teardown).
3. `/gremlins` shows them as dead/stalled. `/gremlins rescue <id>` fails because the worktree is gone.

## Notes

- See `gremlins/fleet.py:116` (`liveness_of_state_file`) and the rescue prompt around `gremlins/fleet.py:540` — the current docstring already calls "missing worktree" out as `unsalvageable`, which is the gap this issue asks to close.
- Boss/child chains make this worse: a terminated boss leaves orphaned children with no parent to advance them.

xbrianh:
## Summary

- Adds `dead:host-terminated` liveness label when pid is gone and the worktree directory no longer exists, distinguishing external host termination from a normal in-process crash
- Adds `_recreate_worktree` helper that reconstructs a missing worktree from the gremlin's recorded branch (preserving in-progress commits) or falls back to a detached checkout of `worktree_base`/HEAD
- `do_rescue` now handles `dead:host-terminated` by attempting worktree recreation before proceeding with normal diagnosis/relaunch; headless rescue bails with `host_terminated_unrecoverable` when recovery isn't possible
- Both the Python (`fleet/state.py`) and bash (`liveness.sh`) classifiers are updated in lockstep per their documented invariant

## Test plan

- [ ] `liveness_of_state_file` returns `dead:host-terminated` when pid is gone and workdir doesn't exist
- [ ] Returns `dead:crashed` when pid is gone but workdir exists (no regression)
- [ ] Headless rescue bails with `host_terminated_unrecoverable` when project_root is also gone
- [ ] Headless rescue bails with `host_terminated_unrecoverable` when worktree recreation fails
- [ ] Rescue proceeds to diagnosis step when worktree recreation succeeds
- [ ] Full test suite: 190 passed

Closes #160

### DIFF
--- a/gremlins/fleet/rescue.py
+++ b/gremlins/fleet/rescue.py
@@ -447,6 +447,13 @@ def _recreate_worktree(state: dict) -> tuple:
     if not os.path.isdir(project_root):
         return False, f"project_root {project_root!r} does not exist"
 
+    # Prune stale worktree entries so that git worktree add doesn't fail
+    # with "is already registered" when the directory was deleted by the host.
+    subprocess.run(
+        ["git", "worktree", "prune"],
+        capture_output=True, cwd=project_root,
+    )
+
     if branch:
         r = subprocess.run(
             ["git", "worktree", "add", workdir, branch],
--- a/gremlins/fleet/rescue.py
+++ b/gremlins/fleet/rescue.py
@@ -430,6 +430,49 @@ def _run_headless_diagnosis(workdir: str, prompt: str, marker_path: str):
     return _read_rescue_marker(marker_path)
 
 
+def _recreate_worktree(state: dict) -> tuple:
+    """Attempt to recreate a missing worktree from the gremlin's branch or base ref.
+
+    Tries the named branch first (preserves in-progress commits for localgremlin),
+    then falls back to a detached checkout of worktree_base or HEAD.
+    Returns (success: bool, detail: str).
+    """
+    workdir = state.get("workdir") or ""
+    branch = state.get("branch") or ""
+    worktree_base = state.get("worktree_base") or ""
+    project_root = state.get("project_root") or ""
+
+    if not workdir or not project_root:
+        return False, "workdir or project_root not recorded in state"
+    if not os.path.isdir(project_root):
+        return False, f"project_root {project_root!r} does not exist"
+
+    if branch:
+        r = subprocess.run(
+            ["git", "worktree", "add", workdir, branch],
+            capture_output=True, text=True, cwd=project_root,
+        )
+        if r.returncode == 0:
+            return True, f"recreated from branch {branch!r}"
+        branch_err = r.stderr.strip()
+    else:
+        branch_err = ""
+
+    ref = worktree_base or "HEAD"
+    r = subprocess.run(
+        ["git", "worktree", "add", "--detach", workdir, ref],
+        capture_output=True, text=True, cwd=project_root,
+    )
+    if r.returncode == 0:
+        suffix = f" (branch {branch!r} was gone)" if branch_err else ""
+        return True, f"recreated detached at {ref!r}{suffix}"
+    fallback_err = r.stderr.strip()
+
+    if branch_err:
+        return False, f"branch {branch!r}: {branch_err}; fallback {ref!r}: {fallback_err}"
+    return False, f"worktree add --detach {ref!r}: {fallback_err}"
+
+
 def do_rescue(target: str, headless: bool = False) -> bool:
     match = resolve_gremlin(target)
     if match is None:
@@ -464,6 +507,30 @@ def do_rescue(target: str, headless: bool = False) -> bool:
     if not workdir:
         print(f"error: no workdir recorded in state for {gr_id} — cannot rescue")
         return False
+
+    if live == "dead:host-terminated":
+        project_root_check = state.get("project_root") or ""
+        print(f"gremlin {gr_id}: worktree is gone (host likely terminated externally)")
+        if not project_root_check or not os.path.isdir(project_root_check):
+            detail = (
+                f"project_root {project_root_check!r} is also gone — "
+                "worktree cannot be recreated; use 'gremlins rm' to clean up"
+            )
+            print(f"  {detail}")
+            if headless:
+                _write_bail(sf, wdir, "host_terminated_unrecoverable", detail)
+            return False
+        print(f"  attempting to recreate worktree at {workdir}...")
+        recreated, detail = _recreate_worktree(state)
+        if not recreated:
+            msg = f"worktree recreation failed: {detail}"
+            print(f"  {msg}")
+            print(f"  use 'gremlins rm {gr_id}' to clean up")
+            if headless:
+                _write_bail(sf, wdir, "host_terminated_unrecoverable", msg)
+            return False
+        print(f"  worktree recreated: {detail}; resuming rescue...")
+        live = liveness_of_state_file(sf, state)
 
     rescue_count_raw = state.get("rescue_count") or 0
     try:
--- a/gremlins/fleet/state.py
+++ b/gremlins/fleet/state.py
@@ -94,6 +94,9 @@ def liveness_of_state_file(sf: str, state=None) -> str:
             try:
                 os.kill(int(gr_pid), 0)
             except (OSError, ValueError):
+                workdir = state.get("workdir") or ""
+                if workdir and not os.path.exists(workdir):
+                    return "dead:host-terminated"
                 return f"dead:crashed (pid {gr_pid} gone)"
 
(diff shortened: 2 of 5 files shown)


skills/_bg/liveness.sh
```diff
@@ -32,17 +32,18 @@ liveness_of_state_file() {
...
     # bash treats tab as IFS-whitespace and collapses consecutive empty
     # columns, so a future 5th field could silently lose a value. US is
     # non-whitespace.
-    IFS=$'\x1f' read -r gr_status gr_pid gr_exit_code gr_bail_reason < <(
+    IFS=$'\x1f' read -r gr_status gr_pid gr_exit_code gr_bail_reason gr_workdir < <(
         jq -r '[.status, (.pid // "" | tostring),
                 (.exit_code // "" | tostring),
-                (.bail_reason // "")] | join("\u001f")' "$sf" 2>/dev/null || true
+                (.bail_reason // ""),
+                (.workdir // "")] | join("")' "$sf" 2>/dev/null || true
```
xbrianh: **Bug: `join("")` should be `join("\u001f")`.**

The `IFS` is set to `$'\x1f'` (ASCII unit separator), but the `jq` join now uses an empty string as the delimiter. All five fields are concatenated into a single string with no separators, so `read` can't split them — everything goes into `gr_status` and `gr_workdir` is always empty. The `dead:host-terminated` branch at line 67 will never fire in bash regardless of state.

Fix:
```bash
                (.workdir // "")] | join("\u001f")' "$sf" 2>/dev/null || true
```

The same `join("\u001f")` was used for the original four fields; the fifth needs it too.

gremlins/fleet/rescue.py
```diff
@@ -430,6 +430,49 @@ def _run_headless_diagnosis(workdir: str, prompt: str, marker_path: str):
...
+    worktree_base = state.get("worktree_base") or ""
+    project_root = state.get("project_root") or ""
+
+    if not workdir or not project_root:
+        return False, "workdir or project_root not recorded in state"
+    if not os.path.isdir(project_root):
+        return False, f"project_root {project_root!r} does not exist"
+
+    if branch:
+        r = subprocess.run(
```
xbrianh, line 458: **Missing `git worktree prune` before `git worktree add` will cause recreation to consistently fail.**

When a gremlin runs, it registers a worktree entry in `.git/worktrees/`. After host termination the directory is gone but the entry persists. When `_recreate_worktree` runs `git worktree add <workdir> <branch>`, git sees the path still has a registered (stale) entry and rejects it:

```
fatal: '<workdir>' is already registered as a worktree
```

The `--detach` fallback on line 462 hits the same error. Both attempts fail, `do_rescue` bails with `host_terminated_unrecoverable`, and the worktree is never recovered — even though the branch and commits are intact.

Fix: run `git worktree prune` in `project_root` before the first add attempt:
```python
subprocess.run(["git", "worktree", "prune"], cwd=project_root, capture_output=True)
```

This removes stale entries for directories that no longer exist, which is exactly the host-terminated case.

gremlins/fleet/state.py
```diff
@@ -94,6 +94,9 @@ def liveness_of_state_file(sf: str, state=None) -> str:
             try:
                 os.kill(int(gr_pid), 0)
             except (OSError, ValueError):
+                workdir = state.get("workdir") or ""
+                if workdir and not os.path.exists(workdir):
```
xbrianh: **Minor lockstep inconsistency: `os.path.exists` vs bash's `-d`.**

The comment in `liveness.sh` (and the CLAUDE.md invariant) says the two classifiers must be kept in lockstep. The bash side uses `-d` (true only for directories); Python uses `os.path.exists` (true for files, symlinks to files, etc.). A `workdir` path that exists as a non-directory would yield `dead:host-terminated` in bash but `dead:crashed` in Python.

Fix:
```python
if workdir and not os.path.isdir(workdir):
```

xbrianh (commented):
Three issues found: one is a correctness bug in `liveness.sh` that silently disables the new detection path in bash, one is a functional gap in `_recreate_worktree` that will cause recreation to consistently fail for normal gremlins, and one is a minor lockstep inconsistency between the Python and bash classifiers. The test suite exercises the Python path correctly but doesn't catch the bash regression.

skills/_bg/liveness.sh
xbrianh: Fixed —  →  so the `\x1f` IFS separator correctly splits the five fields.

gremlins/fleet/rescue.py
xbrianh: Fixed — added `subprocess.run(["git", "worktree", "prune"], ...)` before the first `git worktree add` call so stale entries from host-terminated gremlins are cleared first.

gremlins/fleet/state.py
xbrianh: Fixed — `os.path.exists` → `os.path.isdir` to match bash's `-d` test and keep the two classifiers in lockstep.

skills/_bg/liveness.sh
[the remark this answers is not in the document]
xbrianh: Fixed — now using the explicit escape sequence.

gremlins/fleet/state.py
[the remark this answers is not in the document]
xbrianh: Fixed — changed to os.path.isdir.

gremlins/fleet/rescue.py
```diff
@@ -430,6 +430,49 @@ def _run_headless_diagnosis(workdir: str, prompt: str, marker_path: str):
...
+            return True, f"recreated from branch {branch!r}"
+        branch_err = r.stderr.strip()
+    else:
+        branch_err = ""
+
+    ref = worktree_base or "HEAD"
+    r = subprocess.run(
+        ["git", "worktree", "add", "--detach", workdir, ref],
+        capture_output=True, text=True, cwd=project_root,
+    )
```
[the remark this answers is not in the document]
xbrianh: Fixed — added git worktree prune before the first worktree add call.

skills/_bg/liveness.sh
```diff
@@ -32,17 +32,18 @@ liveness_of_state_file() {
     [[ -f "$sf" ]] || return 0
     # Note: avoid `status` as a local name — it's a special/readonly in zsh,
     # and this file is intended to be sourced from either bash or zsh hooks.
-    local wdir gr_status gr_pid gr_exit_code gr_bail_reason
+    local wdir gr_status gr_pid gr_exit_code gr_bail_reason gr_workdir
     wdir=$(dirname "$sf")
 
     # US (\x1f) separator, matching session-summary.sh and pipeline/fleet.py:
     # bash treats tab as IFS-whitespace and collapses consecutive empty
     # columns, so a future 5th field could silently lose a value. US is
```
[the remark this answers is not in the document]
xbrianh: Fixed — updated to "future 6th field".

Issue #169, opened 2026-04-29
handoff: rolling plan still names completed work despite design rule


xbrianh:
## Symptom

The \`handoff\` agent's rolling plan output names completed work in prose, even though the design rule says the plan should contain only forward work.

## Design rule (from \`gremlins/handoff.py\` build_prompt)

> The rolling plan describes only **remaining** work. Completed tasks are removed entirely — no \`[x]\` markers, no struck-through entries, no "completed" appendix. The chain of versioned plan files plus git history is the audit trail; the rolling plan does not repeat it. Do not propagate the overarching goal of the chain forward into the rolling plan — that lives upstream, in the original spec.

> **\`chain-done\`**: minimal output. A short note that the chain is complete is enough — no leftover task list, no carried-over context.

## Evidence

\`~/.local/state/claude-gremlins/extract-pipeline-package-via-phased-d04390/handoff-005.md\` (a \`next-plan\` exit from a real chain) opens with:

\`\`\`
# Rolling plan: extract pipeline/ package via phased migration

Phases 0, 1, 2, and 3 have landed (#106, #110, #114, #117). Phases 4–5 remain.
\`\`\`

…and \`handoff-006.md\` (chain-done) opens with:

\`\`\`
# Rolling plan: extract pipeline/ package via phased migration

Chain complete. All six phases (0–5) have landed:

- Phase 0 (#106): \`pipeline/DESIGN.md\` design doc
- Phase 1 (#110): package scaffold + localgremlin migration
- Phase 2 (#114): test suite against the new seam
- Phase 3 (#117): ghgremlin migration
- Phase 4 (#119): bossgremlin migration
- Phase 5 (cleanup): \`_core.py\` absent, CLAUDE.md and README.md stale references cleared, all skills directories contain only SKILL.md + thin shims
\`\`\`

Both cases violate the rule. The first leaks completed work into the prose preamble; the second is a full enumeration of completed phases under chain-done. The chain-done case also retains the overarching goal in its H1 title, which the rule explicitly forbids ("Do not propagate the overarching goal … into the rolling plan").

## Why the prompt isn't catching it

The prompt enumerates three forms it forbids — \`[x]\` markers, struck-through entries, "completed" appendix — but doesn't explicitly forbid:

- prose statements about what has landed (e.g. "Phases 0, 1, 2, and 3 have landed (#106, …)")
- bulleted lists of completed phases under \`chain-done\`
[...]

PR #172, opened 2026-04-29
feat: add sanitize pass to handoff rolling plan output

### LINKED ISSUE #171: handoff: add sanitize pass to enforce rolling plan format rules
## Context

The handoff agent's rolling plan output consistently violates the format rules defined in `build_prompt` (ref #169). The root cause is architectural: the main agent simultaneously reasons about the diff, decides the exit state, and writes the output document — format rules compete with content decisions for attention. Prompt-tightening alone is fragile because the agent treats the listed prohibited forms as an exhaustive set and finds compliant-sounding alternatives (prose preambles, bullet lists of completed phases, overarching-goal H1 titles).

The fix is two-pronged: (1) tighten the rules in `build_prompt` to be exhaustive rather than exemplary, and (2) add a lightweight sanitize pass — a second `claude -p` call with no git context, no diff, no spec — that enforces format rules mechanically after the main agent finishes.

## Approach

Add `build_sanitize_prompt` and `sanitize_rolling_plan` helpers to `gremlins/handoff.py`. After the main agent succeeds and the signal file is verified, `main()` calls `sanitize_rolling_plan(out_path, ...)` which reads the rolling plan, runs a focused `claude -p` with only the plan text and a tight "never include X" ruleset, and overwrites `out_path` in-place. No exit-state knowledge is needed in the sanitize call — the rules are universal. Sanitize failure is non-fatal: log a warning and keep the original.

The sanitize call hardcodes haiku — format enforcement is a mechanical task that doesn't require a more capable model.

Alongside this, tighten `build_prompt` rule language: change the three-item list of prohibited forms to a comprehensive statement covering prose landing summaries, bullet lists of completed items, H1 titles that name the chain goal, and in-line parentheticals referencing merged PRs.

## Tasks

- [ ] Task 1: Add `build_sanitize_prompt(rolling_plan_text: str, out_path: pathlib.Path) -> str` to `gremlins/handoff.py`. Rules must explicitly prohibit: prose statements about what landed/shipped/merged (at any document position), bullet lists of completed items, `[x]` markers, struck-through entries, and H1 titles that name the overall chain goal. Prompt must instruct the agent to output only the rewritten document to `out_path` with no explanation.
- [ ] Task 2: Add `sanitize_rolling_plan(out_path: pathlib.Path, timeout: Optional[int]) -> None` to `gremlins/handoff.py`. Reads `out_path`, skips with a warning if missing, runs `claude -p --model haiku` with the sanitize prompt, logs a warning (does not raise) on non-zero return.
- [ ] Task 3: Call `sanitize_rolling_plan` from `main()` after signal-file verification and before returning, for all exit states.
- [ ] Task 4: Tighten `build_prompt` in `gremlins/handoff.py`: rewrite the "rolling plan" format rules to be exhaustive — no prose summaries of completed work, no bullets listing completed phases, no H1 that names the chain goal — rather than listing three exemplary prohibited forms.
- [ ] Task 5: Update `_stub_happy_main` in `gremlins/tests/test_handoff.py` to handle the second `subprocess.run` call (sanitize pass). The updated helper should write a dummy `out_path` during the main agent call and treat the second call as a no-op (return 0).
- [ ] Task 6: Add `gremlins/tests/fixtures/handoff_bad_next_plan.md` and `handoff_bad_chain_done.md` containing the real violating outputs from #169 (the "Phases 0–3 have landed" preamble and the "Chain complete. All six phases" enumeration).
- [ ] Task 7: Add tests to `test_handoff.py`: (a) `test_build_sanitize_prompt_rules` — assert the sanitize prompt explicitly names each prohibited pattern; (b) `test_sanitize_rolling_plan_rewrites_file` — fake claude writes a cleaned version, assert `out_path` is overwritten; (c) `test_sanitize_rolling_plan_nonzero_is_nonfatal` — fake claude returns 1, assert `main()` still returns 0 and logs a warning; (d) two regression tests loading the fixture files and asserting `build_sanitize_prompt` output would reject each violation pattern.

## Open questions

(none)

Ref #169

xbrianh:
## Summary

- Adds `build_sanitize_prompt` and `sanitize_rolling_plan` to `gremlins/handoff.py`: after the main agent writes the rolling plan, a second `claude -p --model haiku` call strips prose-about-landing, bullet lists of completed items, `[x]` markers, struck-through entries, and goal-naming H1 titles
- Sanitize failure is non-fatal — logs a warning and keeps the original
- Tightens `build_prompt` format rules from three exemplary prohibited forms to an exhaustive five-point list
- Adds 7 new tests and two fixture files reproducing the real violation patterns from #169

## Test plan

- [ ] `cd gremlins && PYTHONPATH=. python -m pytest tests/test_handoff.py` — all 34 tests pass

Closes #171

### DIFF
--- a/gremlins/handoff.py
+++ b/gremlins/handoff.py
@@ -245,7 +245,7 @@ def build_sanitize_prompt(rolling_plan_text: str, out_path: pathlib.Path) -> str
 
 ## What to keep
 
-Keep all remaining task lists (`- [ ] ...`), open questions, context relevant to what is still to be done, and operator follow-ups. Do not add any content that was not in the original.
+Keep all remaining task lists (`- [ ] ...`), open questions, context relevant to what is still to be done, and operator follow-ups. You may make minimal wording changes only when needed to satisfy the rules above, such as replacing a too-broad H1 with one scoped to the remaining work or rephrasing surrounding context so it refers only to unfinished work. Do not invent new tasks, requirements, decisions, or factual claims that are not supported by the original.
 
 ## Output
 
@@ -268,16 +268,37 @@ def sanitize_rolling_plan(out_path: pathlib.Path, timeout: Optional[int]) -> Non
     except OSError as exc:
         sys.stderr.write(f"warning: sanitize skipped — could not read rolling plan: {exc}\n")
         return
+    def restore_original(reason: str) -> None:
+        try:
+            out_path.write_text(plan_text, encoding="utf-8")
+        except OSError as exc:
+            sys.stderr.write(
+                f"warning: {reason} — failed to restore original rolling plan: {exc}\n"
+            )
+            return
+        sys.stderr.write(f"warning: {reason} — restored original rolling plan\n")
+
+    sanitize_timeout = min(timeout, 60) if timeout is not None else 60
     prompt = build_sanitize_prompt(plan_text, out_path)
     cmd = ["claude", "-p", "--model", "haiku", *CLAUDE_FLAGS, prompt]
     print("==> sanitizing rolling plan", flush=True)
     try:
-        result = subprocess.run(cmd, timeout=timeout)
+        result = subprocess.run(cmd, timeout=sanitize_timeout, capture_output=True, text=True)
     except subprocess.TimeoutExpired:
-        sys.stderr.write("warning: sanitize pass timed out — keeping original\n")
+        restore_original("sanitize pass timed out")
         return
+    if result.stderr:
+        sys.stderr.write(f"warning: sanitize stderr:\n{result.stderr}")
     if result.returncode != 0:
-        sys.stderr.write(f"warning: sanitize pass exited {result.returncode} — keeping original\n")
+        restore_original(f"sanitize pass exited {result.returncode}")
+        return
+    try:
+        sanitized_text = out_path.read_text(encoding="utf-8")
+    except OSError as exc:
+        restore_original(f"sanitize pass completed but output could not be read: {exc}")
+        return
+    if not sanitized_text.strip():
+        restore_original("sanitize pass completed but output was empty")
 
 
 def parse_args(argv: List[str]) -> argparse.Namespace:
@@ -290,7 +311,7 @@ def parse_args(argv: List[str]) -> argparse.Namespace:
     parser.add_argument("--base", dest="base", default=None)
     parser.add_argument("--model", dest="model", default="sonnet")
     parser.add_argument("--timeout", dest="timeout", type=int, default=None,
-                        help="timeout in seconds for the inner claude agent (default: no timeout)")
+                        help="timeout in seconds for the main handoff agent (default: no timeout); the sanitize pass is capped at min(timeout, 60)s")
     parser.add_argument("--rev", dest="rev", default=None,
                         help="inspect this ref instead of HEAD (e.g. a target branch name)")
     return parser.parse_args(argv)
--- a/gremlins/handoff.py
+++ b/gremlins/handoff.py
@@ -182,7 +182,14 @@ If the spec author wrote operator-flavoured language inline with implementation 
 
 5. Write an **updated plan document** (the "rolling plan") to: `{out_path}`
 
-   The rolling plan describes only **remaining** work. Completed tasks are removed entirely — no `[x]` markers, no struck-through entries, no "completed" appendix. The chain of versioned plan files plus git history is the audit trail; the rolling plan does not repeat it. Do not propagate the overarching goal of the chain forward into the rolling plan — that lives upstream, in the original spec.
+   The rolling plan describes only **remaining** work. These forms are **never** allowed anywhere in the document, at any position:
+   - Prose statements about what has landed, shipped, merged, or been completed — e.g. "Phases 0–3 have landed", "X was merged in PR #N", "the following work is complete", "all tasks in this phase are done"
+   - Bullet lists enumerating completed phases or items
+   - `[x]` checkboxes or checked markers of any kind
+   - Struck-through entries (~~text~~)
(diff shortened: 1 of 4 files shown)


gremlins/handoff.py
```diff
@@ -225,6 +232,54 @@ def build_prompt(
...
+    try:
+        plan_text = out_path.read_text(encoding="utf-8")
+    except OSError as exc:
+        sys.stderr.write(f"warning: sanitize skipped — could not read rolling plan: {exc}\n")
+        return
+    prompt = build_sanitize_prompt(plan_text, out_path)
+    cmd = ["claude", "-p", "--model", "haiku", *CLAUDE_FLAGS, prompt]
+    print("==> sanitizing rolling plan", flush=True)
+    try:
+        result = subprocess.run(cmd, timeout=timeout)
```
xbrianh: Both subprocess calls use the same `timeout` value, so `--timeout N` is a per-call ceiling, not a budget for the whole handoff. Worst case is 2N seconds of wall time. The `--timeout` help text still says "for the inner claude agent" (singular) -- worth updating it, or capping the sanitize timeout to something shorter like `min(timeout, 60)` since Haiku is fast.

xbrianh (commented):
Clean implementation -- the sanitize pass is well-scoped, non-fatal by design, and the test coverage is solid. Two things worth a look:

1. **Shared timeout**: the sanitize pass uses the same `--timeout` value as the main agent, so the ceiling is per-call, not for the whole handoff. A caller who sets `--timeout 120` may see up to 240s total.

2. **Spurious warning in `test_main_next_plan_signal`**: that test replaces `fake_run` entirely with a lambda that never writes `out_path`. As a result, `sanitize_rolling_plan` emits "warning: sanitize skipped -- rolling plan not found" to stderr on every run. The test does not assert on stderr so it passes, but it leaks noise and diverges from the comment added to the docstring on line 217.

gremlins/handoff.py
```diff
@@ -225,6 +232,54 @@ def build_prompt(
...
+Keep all remaining task lists (`- [ ] ...`), open questions, context relevant to what is still to be done, and operator follow-ups. Do not add any content that was not in the original.
+
+## Output
+
+Write ONLY the rewritten document to: {out_path}
+Do not print the document to stdout. Do not explain what you changed.
+
+## Rolling plan to rewrite
+
+~~~~
```
xbrianh, line 257: The rolling plan is fenced with `~~~~`. If the plan itself contains a bare `~~~~` line, the fence closes early and the model receives a malformed document. Low probability in practice (the main agent prompt uses the same convention, so well-formed plans unlikely to contain it), but worth a note as a known fragility.

gremlins/handoff.py
```diff
@@ -225,6 +232,54 @@ def build_prompt(
...
+
+1. Prose statements about what has landed, shipped, merged, or been completed at any document position — e.g. "Phases 0–3 have landed", "X was merged in PR #N", "the following work is complete", "all tasks in this phase are done". Remove such sentences entirely.
+2. Bullet lists enumerating completed phases or items — any bullet that describes something already done. Remove them.
+3. `[x]` checkboxes or checked markers of any kind. Remove the entire line.
+4. Struck-through entries (~~text~~). Remove the entire line.
+5. An H1 title (`# ...`) that names the overall chain goal or summarizes the completed chain — e.g. `# Implement Feature X` or `# Claude Config Personal Setup`. Replace it with a short H1 scoped only to the remaining work.
+
+## What to keep
+
+Keep all remaining task lists (`- [ ] ...`), open questions, context relevant to what is still to be done, and operator follow-ups. Do not add any content that was not in the original.
```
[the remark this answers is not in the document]
xbrianh: Fixed. Rewrote the 'What to keep' paragraph to explicitly allow minimal wording changes when needed to satisfy the rules (e.g. replacing a too-broad H1, rephrasing context to refer only to unfinished work), while still forbidding inventing new tasks, requirements, or factual claims.

gremlins/handoff.py
```diff
@@ -225,6 +232,54 @@ def build_prompt(
...
+    prompt = build_sanitize_prompt(plan_text, out_path)
+    cmd = ["claude", "-p", "--model", "haiku", *CLAUDE_FLAGS, prompt]
+    print("==> sanitizing rolling plan", flush=True)
+    try:
+        result = subprocess.run(cmd, timeout=timeout)
+    except subprocess.TimeoutExpired:
+        sys.stderr.write("warning: sanitize pass timed out — keeping original\n")
+        return
+    if result.returncode != 0:
+        sys.stderr.write(f"warning: sanitize pass exited {result.returncode} — keeping original\n")
```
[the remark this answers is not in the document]
xbrianh: Fixed. The function now saves `plan_text` before invoking the subprocess, then calls `restore_original()` on timeout, non-zero exit, unreadable output, or empty output. Post-run it reads back the file and restores the original if the result is empty.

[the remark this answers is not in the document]
xbrianh: Fixed. Added `capture_output=True, text=True` to the subprocess call. Stderr from the sanitize agent is now surfaced as a warning; stdout is discarded.

gremlins/handoff.py
xbrianh: Fixed. Sanitize timeout is now `min(timeout, 60)` (defaulting to 60s when timeout is None), so the main agent keeps the full `--timeout` budget and the sanitize pass is bounded separately. Updated the `--timeout` help text to reflect this.

gremlins/handoff.py
xbrianh: Acknowledged as a known fragility. Low probability since well-formed plans generated by the main agent shouldn't contain a bare `~~~~` line. No fix in this PR.

PR #173, opened 2026-04-29
fix: land boss children into boss worktree, not project_root

### LINKED ISSUE #170: boss-launched local children land squash commits on user's current branch in project_root
## Symptom

While a boss gremlin chain is running, each local child gremlin's `fleet land` step squash-merges onto whatever branch the user happens to have checked out in the user's actual repo. This blocks the user from doing any work in the repo while the boss spins, and silently rewrites the user's branch with chain commits.

## Reproduction (mental walkthrough)

1. `cd ~/Desktop/amorphous-industries/claude` (some feature branch checked out, say `main`).
2. Launch a boss chain: `/bossgremlin --plan some-spec.md --chain-kind local`.
3. Boss creates a detached worktree at HEAD (`skills/_bg/launch.sh`:599) and spawns local children.
4. Each child's `project_root` is set to the user's actual repo via the `--parent` inheritance fix (`skills/_bg/launch.sh`:344-356, added in #159 / 865327b).
5. When a child finishes, the boss calls `python -m gremlins.cli fleet land <child>` (`gremlins/orchestrators/boss.py:471-475`).
6. `_land_local` (`gremlins/fleet/land.py:547-579`) reads `project_root` from the child's state — which is the user's actual repo — and runs `_squash_land` with `cwd=project_root`.
7. `_squash_land` (`gremlins/fleet/land.py:455-508`) runs `git merge --squash <branch>` followed by `git commit`, on the current branch in `project_root`.
8. After landing, `advance_boss_workdir` (`gremlins/orchestrators/boss.py:478-491`, added in #168 / 89519f2) does `git reset --hard <new_head>` in the boss worktree to mirror the user's repo.

Net effect: each chain step writes a squash commit onto the user's checked-out branch.

## Why this is wrong

- The user can't switch branches, do unrelated work, or even keep an uncommitted edit during a chain — `_preflight_land` requires a clean working tree, so a child's land will fail if the user has uncommitted changes.
- The user's branch silently grows chain commits with no visible signal in the boss state dir.
- If the user switches branches mid-chain, subsequent children land on the new branch rather than where the chain started.
- It violates the implicit promise of `gremlins` as background workers: the user's working repo should be untouched until they explicitly opt in to landing.

## Likely cause

This is a regression introduced (or made worse) by recent disaster-recovery / project_root threading work:

- **#159 (865327b):** Made boss-launched children inherit `project_root` from the parent's state.json. Before this, children recorded the boss's worktree as `project_root`, so a land operation would have hit the boss worktree, not the user's repo. Whatever land semantics existed before, they didn't write to the user's checked-out branch. The fix solved a `gremlins --here` filter problem but inadvertently re-pointed every land at the user's repo.
- **#168 (89519f2):** Reinforces this architecture by explicitly advancing the boss worktree to mirror `project_root` after each land. The commit message is candid: "fleet land squash-merges into project_root but leaves the boss's own worktree frozen." That sentence describes the bug: it shouldn't be landing into project_root at all.

## Suggested direction (not a fix — for the implementer)

The boss's own worktree is the natural landing zone: it's already detached at the chain's start SHA, isolated from the user's checkout, and its contents are exactly what the user would want to inspect / merge at chain end. Possible shapes (for the implementer to evaluate):

1. **Land into the boss worktree, not project_root.** Either thread a `--into <path>` argument through `fleet land` for boss-launched children, or have the boss orchestrator pass its own `workdir` as the landing cwd. After landing, no `advance_boss_workdir` is needed because the boss workdir *is* where the commits live.
2. **Give the boss its own real branch (not detached).** Create `bg/bossgremlin/<GR_ID>` at chain start, land children onto it inside the boss worktree, and let the user `gremlins land <boss-id>` at the end to merge that branch into their actual checkout.
3. **At minimum, hard-fail the chain start if the user's working repo isn't clean / on the expected branch**, so the chain can't silently surprise the user with commits on a feature branch they've moved on from.

The architectural goal: the user's `project_root` checkout should be read-only from the boss's perspective until the user explicitly runs `fleet land` against the boss.

xbrianh:
## Summary

- Local child gremlins launched by a boss chain were squash-landing onto the user's checked-out branch in `project_root`, blocking the user from switching branches or keeping uncommitted edits during a chain run.
- Threads `--into <path>` through `fleet land` → `_land_local` so the boss directs each child's squash-merge into its own isolated detached-HEAD worktree instead.
- Removes `advance_boss_workdir`, which existed only to mirror `project_root`'s HEAD back into the boss worktree — no longer needed since children land there directly.

## Test plan

- [ ] `python -m pytest gremlins/tests/` — 194 tests pass
- [ ] `test_land_child_uses_boss_workdir_for_local_chain` — verifies `land_child` receives `into_dir=boss_workdir` for local chains
- [ ] `test_land_child_no_into_dir_for_gh_chain` — verifies `into_dir=""` for gh chains
- [ ] `test_land_local_into_dir_lands_in_worktree` — end-to-end: squash commit lands in boss worktree, not `project_root`
- [ ] `test_land_local_into_dir_nonexistent_fails` — error path when `--into` dir is missing

Closes #170

### DIFF
--- a/gremlins/fleet/cli.py
+++ b/gremlins/fleet/cli.py
@@ -143,6 +143,9 @@ def _dispatch_subcommand(argv: List[str]):
             into_idx = raw.index("--into")
             if into_idx + 1 < len(raw):
                 into_dir = raw[into_idx + 1]
+            else:
+                print("error: --into requires a directory argument")
+                sys.exit(1)
         ok = do_land(target, force=force, mode=mode, into_dir=into_dir)
     else:
         headless = "--headless" in raw
--- a/gremlins/fleet/cli.py
+++ b/gremlins/fleet/cli.py
@@ -138,7 +138,12 @@ def _dispatch_subcommand(argv: List[str]):
             print("error: --squash and --ff are mutually exclusive")
             sys.exit(1)
         mode = "squash" if squash_flag else ("ff" if ff_flag else None)
-        ok = do_land(target, force=force, mode=mode)
+        into_dir = ""
+        if "--into" in raw:
+            into_idx = raw.index("--into")
+            if into_idx + 1 < len(raw):
+                into_dir = raw[into_idx + 1]
+        ok = do_land(target, force=force, mode=mode, into_dir=into_dir)
     else:
         headless = "--headless" in raw
         ok = do_rescue(target, headless=headless)
--- a/gremlins/fleet/land.py
+++ b/gremlins/fleet/land.py
@@ -544,7 +544,7 @@ def _ff_land(gr_id: str, sf: str, wdir: str, state: dict, cwd,
     return True
 
 
-def _land_local(gr_id: str, sf: str, wdir: str, state: dict, mode: str) -> bool:
+def _land_local(gr_id: str, sf: str, wdir: str, state: dict, mode: str, into_dir: str = "") -> bool:
     """Land a local gremlin branch onto the current branch (mode: 'squash' or 'ff')."""
     setup_kind = state.get("setup_kind", "")
     if setup_kind != "worktree-branch":
@@ -557,7 +557,13 @@ def _land_local(gr_id: str, sf: str, wdir: str, state: dict, mode: str) -> bool:
         return False
 
     project_root = state.get("project_root") or ""
-    cwd = project_root if project_root and os.path.isdir(project_root) else None
+    if into_dir:
+        if not os.path.isdir(into_dir):
+            print(f"error: --into directory does not exist: {into_dir!r}")
+            return False
+        cwd = into_dir
+    else:
+        cwd = project_root if project_root and os.path.isdir(project_root) else None
 
     r = subprocess.run(
         ["git", "show-ref", "--verify", "--quiet", f"refs/heads/{branch}"],
@@ -735,7 +741,7 @@ def _land_gh(gr_id: str, sf: str, wdir: str, state: dict, force: bool = False) -
     return True
 
 
-def do_land(target: str, force: bool = False, mode: str = None) -> bool:
+def do_land(target: str, force: bool = False, mode: str = None, into_dir: str = "") -> bool:
     match = resolve_gremlin(target)
     if match is None:
         return False
@@ -756,7 +762,7 @@ def do_land(target: str, force: bool = False, mode: str = None) -> bool:
         if live != "dead:finished":
             print(f"gremlin {gr_id} is not finished (liveness: {live})")
             return False
-        return _land_local(gr_id, sf, wdir, state, mode or "squash")
+        return _land_local(gr_id, sf, wdir, state, mode or "squash", into_dir=into_dir)
     elif kind == "bossgremlin":
         if live != "dead:finished":
             print(f"gremlin {gr_id} is not finished (liveness: {live})")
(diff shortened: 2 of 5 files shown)


gremlins/fleet/land.py
```diff
@@ -544,7 +544,7 @@ def _ff_land(gr_id: str, sf: str, wdir: str, state: dict, cwd,
     return True
 
 
-def _land_local(gr_id: str, sf: str, wdir: str, state: dict, mode: str) -> bool:
+def _land_local(gr_id: str, sf: str, wdir: str, state: dict, mode: str, into_dir: str = "") -> bool:
     """Land a local gremlin branch onto the current branch (mode: 'squash' or 'ff')."""
```
xbrianh: Docstring is now stale: landing with `into_dir` set doesn't go "onto the current branch". Could be: `"""Land a local gremlin branch (mode: 'squash' or 'ff'). If into_dir is given, land there instead of project_root."""`

gremlins/fleet/cli.py
```diff
@@ -138,7 +138,12 @@ def _dispatch_subcommand(argv: List[str]):
             print("error: --squash and --ff are mutually exclusive")
             sys.exit(1)
         mode = "squash" if squash_flag else ("ff" if ff_flag else None)
-        ok = do_land(target, force=force, mode=mode)
+        into_dir = ""
+        if "--into" in raw:
+            into_idx = raw.index("--into")
+            if into_idx + 1 < len(raw):
```
xbrianh, line 144: If `--into` is the last token (no value follows), `into_dir` silently stays `""` and the land falls back to `project_root` — likely not what the caller intended. Could print an error and `sys.exit(1)` here instead of silently ignoring it:
```python
if into_idx + 1 < len(raw):
    into_dir = raw[into_idx + 1]
else:
    print("error: --into requires a directory argument")
    sys.exit(1)
```
Low-risk since this is an internal API, but worth the guard.

xbrianh (commented):
Clean simplification — threading `into_dir` through the call chain and landing directly into the boss worktree is cleaner than the old `advance_boss_workdir` workaround. Test coverage is solid, including the real-worktree end-to-end case. Two minor nits below.

gremlins/fleet/land.py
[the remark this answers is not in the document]
xbrianh: Fixed: updated the `_land_local` docstring to describe the `into_dir` parameter and clarify it redirects landing away from project_root.

xbrianh: Fixed: docstring now reads: 'Land a local gremlin branch (mode: "squash" or "ff"). If into_dir is given, land there instead of project_root.'

gremlins/fleet/cli.py
```diff
@@ -138,7 +138,12 @@ def _dispatch_subcommand(argv: List[str]):
...
+                into_dir = raw[into_idx + 1]
+        ok = do_land(target, force=force, mode=mode, into_dir=into_dir)
```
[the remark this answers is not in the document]
xbrianh: Fixed: added an `else` branch that prints 'error: --into requires a directory argument' and calls `sys.exit(1)` instead of silently falling back to an empty `into_dir`.

gremlins/fleet/cli.py
xbrianh: Fixed: added the error + exit path.

PR #175, opened 2026-04-29
feat: rewrite state.set_stage and emit_bail as pure Python

### LINKED ISSUE #174: Rewrite state-helpers in Python, remove subprocess from state.py
# Phase 1: State-helpers Python rewrite

etc. over reimplementation.*

## Goal

`gremlins/state.py`'s `set_stage` and `emit_bail` write `state.json`
in pure Python. `boss.py`'s local `set_stage` is gone. No
`subprocess` import in `state.py`. Tests no longer monkeypatch
`SET_STAGE_SH` / `SET_BAIL_SH` or install fake bash executables —
they assert state.json contents directly.

### `set-stage.sh <gr_id> <stage> [sub_stage_json]`

- Set `.stage = <stage>`.
- Set `.stage_updated_at = <now ISO-8601 UTC, second precision>`.
- If `sub_stage` provided: set `.sub_stage = <parsed JSON>`. If not:
  delete the `.sub_stage` key.
- Atomic via tempfile + rename. No-op on missing state.json. No-op
  on missing `gr_id` or `stage`.

### `set-bail.sh <gr_id> <bail_class> [bail_detail]`

- Set `.bail_class = <bail_class>`.
- If `bail_detail` non-empty: set `.bail_detail = <bail_detail>`. If
  empty: delete the `.bail_detail` key.
- Atomic via tempfile + rename. No-op on missing state.json. No-op
  on missing `gr_id` or `bail_class`.

### `gremlins/state.py`

Remove:
- `import subprocess`
- `SET_STAGE_SH` constant (line 27)
- `SET_BAIL_SH` constant (line 28)

Rewrite the module docstring (lines 1–15) — drop the bash carve-out
paragraph; describe the module as state.json writers with the no-op-
without-`GR_ID` and never-raise contracts.

Rewrite `set_stage(stage, sub_stage=None)`:
- No-op when `GR_ID` is unset.
- Resolve `state.json` via `resolve_state_file()` (existing helper)
  and no-op if it's missing.
- Build the patch dict: `{"stage": stage, "stage_updated_at":
  <now-iso-Z>}`. ISO timestamp format must match the bash script:
  `"%Y-%m-%dT%H:%M:%SZ"` in UTC, second precision.
- If `sub_stage is not None`: include `"sub_stage": sub_stage` in
  the patch. If `None`: include `"sub_stage": <SENTINEL>` and have
  the writer treat the sentinel as "delete this key".
- Write atomically (tempfile + `os.replace`).
- Catch and swallow all `Exception`. Never raise.

Rewrite `emit_bail(bail_class, bail_detail="")`:
- No-op when `GR_ID` is unset, when `bail_class` is empty, or when
  `state.json` is missing.
- Patch `{"bail_class": bail_class}`; include
  `"bail_detail": bail_detail` if non-empty, otherwise delete the
  key.
- Atomic write. Swallow all `Exception`.

Recommended: factor a single private writer that takes a dict of
fields-to-set and a list of keys-to-delete, used by both
`set_stage` and `emit_bail` (and replacing the existing
`patch_state` if a clean unification falls out — `patch_state`
currently only sets, never deletes; expanding it to optionally
accept a `_delete=()` argument is the natural shape).

### `gremlins/orchestrators/boss.py`

Delete the local `set_stage` function (lines 111–114).

Update the imports near line 30 to add:
```python
from ..state import set_stage
```

Update all five call sites (lines 253, 714, 794, 806, 836) from
`set_stage(gr_id, "<stage>")` to `set_stage("<stage>")`. The
`gr_id` argument disappears — `state.set_stage` reads `GR_ID` from
env, which `boss_main` already requires at line 627.

The boss-specific stage names (`handoff`, `done`, `waiting`,
`landing`, `rescuing`) are vocabulary the state writer doesn't
care about — no validation needed in `state.set_stage`.

### `gremlins/tests/test_state_isolation.py`

This file (per its docstring at lines 4–29) was specifically built
around the shell-out behavior. Rework rather than patch:

- The test installing a fake `set-stage.sh` executable
  (lines 135–149) becomes a test that calls `state.set_stage(...)`
  with `GR_ID` pointing at a fixture state.json and asserts the
  on-disk contents (`stage`, `stage_updated_at`, `sub_stage`
  presence/absence) post-call.
- The "GR_ID unset → no-op" coverage (line 22) stays — same
  contract, same assertion shape, no fake executable needed.
- Drop the `monkeypatch.setattr("gremlins.state.SET_STAGE_SH", ...)`
  line entirely.
- Update the docstring to drop the "shells out to" framing.

Add coverage for the keys-to-delete behavior:
- `set_stage("foo", sub_stage={"k":1})` then `set_stage("bar")`
  should leave no `sub_stage` key in state.json.
- `emit_bail("other", "detail")` then `emit_bail("other")` should
  leave no `bail_detail` key in state.json.

### `gremlins/tests/test_orchestrator_boss.py`

Line 52: `monkeypatch.setattr(boss_mod, "set_stage", lambda *a: None)`.
After the import change in boss.py, `set_stage` is no longer a
module-level attribute on `boss_mod` — it's an imported name. The
monkeypatch target either becomes `gremlins.state.set_stage` (if
the test wants to suppress all set_stage calls package-wide) or
the test drops the monkeypatch entirely if the fixture state.json
makes the calls harmless.

The arg shape changes too: `lambda *a: None` swallows any args, so
the signature shift (two args → one arg) doesn't break it as
written, but if the test asserts on call args, update accordingly.

### `gremlins/tests/conftest.py`

Line 55 mentions `gremlins.state.set_stage` in a comment. Read the
surrounding fixture to confirm whether anything functional changes
— likely just a comment update.

## Verification

- `grep -n "subprocess\|SET_STAGE_SH\|SET_BAIL_SH" gremlins/state.py`
  returns nothing.
- `grep -n "skills" gremlins/state.py gremlins/orchestrators/boss.py`
  returns nothing for the lines this phase touches (other skills
  references survive in those files until later phases).
- Spot-check: invoke `set_stage("foo")` against a temp state.json
  with `GR_ID` set — observe `.stage`, `.stage_updated_at` set,
  `.sub_stage` absent. Repeat with `sub_stage={"k":1}` — observe
  `.sub_stage` present. Repeat with `sub_stage=None` — observe
  `.sub_stage` removed.
- Boss orchestrator integration: a synthetic boss run that hits all
  five stage transitions (`handoff` / `waiting` / `landing` /
  `rescuing` / `done`) writes the expected stage values to
  state.json without invoking any bash.

## Non-goals for this phase

- Touching `address_code.py`'s prompt fragment (the agent-facing
  `set-bail.sh` invocation lives in a prompt string; phase 3
  rewrites it to use `python -m gremlins.cli bail`).
- Touching the rescue prompt, `GREMLIN_SCRIPTS`, or any docstring
  that references skills — phase 4.
- Deleting `skills/_bg/{set-stage,set-bail}.sh` from disk —
  phase 5 (the bash readers `liveness.sh` /
  `session-summary.sh` keep working off whatever the gremlin
  package now writes; they're not affected by the writer change).
- The launcher, `finish.sh`, or anything spawning a gremlin —
  phase 2.

xbrianh:
## Summary

- Replaces subprocess shell-outs to `set-stage.sh` / `set-bail.sh` with atomic in-process writes to `state.json`
- Expands `patch_state` to accept a `_delete=()` argument so both helpers can remove keys (`sub_stage`, `bail_detail`) without a sentinel
- Removes the local `set_stage` from `boss.py` and imports from `state` instead; updates all five call sites to drop the `gr_id` argument
- Rewrites `test_state_isolation` to assert on-disk `state.json` contents directly and adds coverage for the key-deletion behavior

## Test plan

- [ ] `cd gremlins && PYTHONPATH=. python -m pytest` — all 208 tests pass
- [ ] `grep -n "subprocess\|SET_STAGE_SH\|SET_BAIL_SH" gremlins/state.py` — returns nothing
- [ ] Verify `set_stage("foo")` with `GR_ID` set writes `stage` + `stage_updated_at`, no `sub_stage`
- [ ] Verify `set_stage("foo", sub_stage={...})` writes `sub_stage`; subsequent `set_stage("bar")` removes it

Closes #174

### DIFF
--- a/gremlins/state.py
+++ b/gremlins/state.py
@@ -1,16 +1,7 @@
-"""Stage and bail bookkeeping wrappers.
+"""Stage and bail bookkeeping for gremlin state.json.
 
-Both helpers shell out to the canonical bash scripts under
-``$HOME/.claude/skills/_bg/`` rather than reimplementing the state.json
-patching logic in Python. The bash scripts are also invoked by non-pipeline
-code paths (``session-summary.sh`` is a hook that sources ``liveness.sh``
-to classify gremlins for the at-startup summary), so leaving them as the
-single source of truth keeps the on-disk vocabulary stable across pipeline
-and non-pipeline writers. ``pipeline/fleet.py`` reimplements the liveness
-classifier inline rather than sourcing ``liveness.sh``; the two must be
-kept in lockstep by hand.
-
-Both helpers are no-ops outside a gremlin context (no ``GR_ID``) and never
+Both helpers write state.json atomically in pure Python. Both are no-ops
+outside a gremlin context (no ``GR_ID`` or missing state.json) and never
 raise — stage/bail bookkeeping must not break a running gremlin.
 """
 
@@ -22,10 +13,6 @@ import os
 import pathlib
 import re
 import secrets
-import subprocess
-
-SET_STAGE_SH = pathlib.Path.home() / ".claude" / "skills" / "_bg" / "set-stage.sh"
-SET_BAIL_SH = pathlib.Path.home() / ".claude" / "skills" / "_bg" / "set-bail.sh"
 
 GR_ID_RE = re.compile(r"^[A-Za-z0-9._-]+$")
 
@@ -40,31 +27,17 @@ BAIL_CLASS_OTHER = "other"
 
 
 def set_stage(stage: str, sub_stage=None) -> None:
-    """Shell out to set-stage.sh. No-op without GR_ID or when the helper is
-    missing/non-executable."""
-    gr_id = os.environ.get("GR_ID")
-    if not gr_id:
+    """Write stage and stage_updated_at to state.json. No-op without GR_ID or missing state.json."""
+    if not os.environ.get("GR_ID"):
         return
-    try:
-        if not SET_STAGE_SH.exists() or not os.access(str(SET_STAGE_SH), os.X_OK):
-            return
-    except Exception:
+    sf = resolve_state_file()
+    if sf is None or not sf.exists():
         return
-    args = [str(SET_STAGE_SH), gr_id, stage]
+    now = datetime.datetime.utcnow().strftime("%Y-%m-%dT%H:%M:%SZ")
     if sub_stage is not None:
-        try:
-            args.append(json.dumps(sub_stage))
-        except Exception:
-            return
-    try:
-        subprocess.run(
-            args,
-            stdout=subprocess.DEVNULL,
-            stderr=subprocess.DEVNULL,
-            check=False,
-        )
-    except Exception:
-        pass
+        patch_state(stage=stage, stage_updated_at=now, sub_stage=sub_stage)
+    else:
+        patch_state(_delete=("sub_stage",), stage=stage, stage_updated_at=now)
 
 
 def resolve_session_dir() -> pathlib.Path:
@@ -97,27 +70,16 @@ def resolve_session_dir() -> pathlib.Path:
 
 
 def emit_bail(bail_class: str, bail_detail: str = "") -> None:
-    """Shell out to set-bail.sh to record a bail_class (and optional detail)
-    on the running gremlin's state.json. No-op without GR_ID or when the
-    helper is missing — never raises."""
-    gr_id = os.environ.get("GR_ID")
-    if not gr_id:
+    """Write bail_class (and optional bail_detail) to state.json. No-op without GR_ID, empty bail_class, or missing state.json."""
+    if not os.environ.get("GR_ID") or not bail_class:
         return
-    try:
-        if not SET_BAIL_SH.exists() or not os.access(str(SET_BAIL_SH), os.X_OK):
-            return
-    except Exception:
+    sf = resolve_state_file()
+    if sf is None or not sf.exists():
         return
-    try:
-        subprocess.run(
-            [str(SET_BAIL_SH), gr_id, bail_class, bail_detail],
-            stdout=subprocess.DEVNULL,
-            stderr=subprocess.DEVNULL,
-            check=False,
-            timeout=5,
-        )
-    except Exception:
-        pass
+    if bail_detail:
+        patch_state(bail_class=bail_class, bail_detail=bail_detail)
+    else:
+        patch_state(_delete=("bail_detail",), bail_class=bail_class)
 
 
 def resolve_state_file() -> "Optional[pathlib.Path]":
@@ -132,8 +94,8 @@ def resolve_state_file() -> "Optional[pathlib.Path]":
     return state_root / gr_id / "state.json"
 
 
-def patch_state(**fields) -> None:
-    """Merge keyword fields into state.json atomically.
+def patch_state(_delete=(), **fields) -> None:
+    """Merge keyword fields into state.json atomically, deleting any keys in _delete.
 
     No-op when GR_ID is unset, when state.json doesn't exist, or when the
     write fails — stage bookkeeping must not crash a running gremlin.
@@ -143,6 +105,8 @@ def patch_state(**fields) -> None:
         return
     try:
         data = json.loads(sf.read_text(encoding="utf-8"))
+        for key in _delete:
+            data.pop(key, None)
         data.update(fields)
         tmp = sf.with_name(
             f"{sf.name}.{os.getpid()}.{secrets.token_hex(8)}.tmp"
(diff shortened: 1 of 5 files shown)


gremlins/state.py
```diff
@@ -40,31 +27,17 @@
...
         return
-    try:
-        if not SET_STAGE_SH.exists() or not os.access(str(SET_STAGE_SH), os.X_OK):
-            return
-    except Exception:
+    sf = resolve_state_file()
+    if sf is None or not sf.exists():
         return
-    args = [str(SET_STAGE_SH), gr_id, stage]
+    now = datetime.datetime.utcnow().strftime("%Y-%m-%dT%H:%M:%SZ")
```
xbrianh: `datetime.datetime.utcnow()` is deprecated since Python 3.12 and scheduled for removal in 3.14. Use:
```python
now = datetime.datetime.now(datetime.timezone.utc).strftime("%Y-%m-%dT%H:%M:%SZ")
```

xbrianh (commented):
Clean rewrite — the subprocess elimination is straightforward and the test expansion is good coverage. Three findings:

1. **`gremlins/CLAUDE.md` stale section** (not in the diff, but important): lines 73–84 still contain a **"Bash-script carve-outs"** section that explicitly says `state.set_stage` and `state.emit_bail` shell out to bash scripts and "Don't reimplement that bookkeeping in pure Python — forks would drift." This directly contradicts the new implementation and will actively mislead future AI agents reading the CLAUDE.md. The section should be removed or replaced with a note describing the pure-Python approach.

2. `datetime.datetime.utcnow()` deprecation — see inline.

3. The `resolve_state_file()` / `sf.exists()` guard in `set_stage` and `emit_bail` is unprotected by try/except, partially undermining the "never raise" contract the module docstring promises — see inline.

gremlins/state.py
```diff
@@ -40,31 +27,17 @@
...
-    if not gr_id:
+    """Write stage and stage_updated_at to state.json. No-op without GR_ID or missing state.json."""
+    if not os.environ.get("GR_ID"):
         return
-    try:
-        if not SET_STAGE_SH.exists() or not os.access(str(SET_STAGE_SH), os.X_OK):
-            return
-    except Exception:
+    sf = resolve_state_file()
+    if sf is None or not sf.exists():
```
xbrianh: `resolve_state_file()` and `sf.exists()` are called outside any try/except. `resolve_state_file()` internally calls `pathlib.Path.home()`, which can raise on misconfigured systems (no home dir, `/etc/passwd` issues, etc.). The module docstring promises "never raise" but only `patch_state`'s body is wrapped — the guard preamble here (and the equivalent at lines 76–78 in `emit_bail`) isn't. Wrapping each function body in `try/except Exception: pass` would close the gap.

gremlins/state.py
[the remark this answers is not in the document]
xbrianh: Fixed — added `if not stage: return` as the first guard in `set_stage`.

xbrianh: Fixed — replaced with `datetime.datetime.now(datetime.timezone.utc).strftime(...)`.

gremlins/tests/test_orchestrator_boss.py
```diff
@@ -49,7 +49,6 @@ def _common_boss_patches(monkeypatch, tmp_path, gr_id):
     monkeypatch.setenv("GR_ID", gr_id)
     monkeypatch.setattr(boss_mod, "STATE_ROOT", str(tmp_path))
     monkeypatch.setattr(boss_mod, "_stop_requested", False)
-    monkeypatch.setattr(boss_mod, "set_stage", lambda *a: None)
     monkeypatch.setattr(boss_mod, "get_head_ref", lambda p: "abc123def456abc1")
```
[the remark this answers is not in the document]
xbrianh: Fixed — re-added `boss_mod.set_stage` stub in `_common_boss_patches` so the real `state.set_stage` is never called from boss tests. (Used attribute monkeypatching on the imported name in `boss_mod` rather than `XDG_STATE_HOME`, to avoid interfering with the one test that sets its own `XDG_STATE_HOME` for a specific assertion.)

gremlins/state.py
xbrianh: Fixed — wrapped both `set_stage` and `emit_bail` bodies in `try/except Exception: pass` so any exception from `resolve_state_file()`, `sf.exists()`, or anywhere else is silently swallowed, matching the module-level 'never raise' contract.

PR #177, opened 2026-04-30
feat: port launch.sh/finish.sh to Python (Phase 2)

### LINKED ISSUE #176: Port launcher and finish scripts from bash to Python
# Phase 2: Launcher port

etc. over reimplementation.*

## Goal

The launcher is a Python module. `gremlins/launcher.py` exposes
`launch(...)` and `resume(...)` functions. A private
`_run-pipeline` CLI subcommand is the spawn boundary, replacing
`finish.sh`'s terminal-state bookkeeping with a `try/finally`. A
public `launch` CLI subcommand wraps the same functions for direct
human invocation. `boss.launch_child` and `rescue` relaunch import
and call the Python API instead of `subprocess.run`-ing a bash
script. `skills/_bg/launch.sh` and `skills/_bg/finish.sh` are
deleted. The shell-integration tests under `gremlins/tests/` are
either ported to drive the Python launcher or replaced with
Python-native tests.

## What `launch.sh` does (port targets)

Source: `skills/_bg/launch.sh` (701 lines). Read it once before
starting; the high-level shape is below.

### CLI surface (forward path)

```
launch.sh [--description <phrase>] [--parent <boss-id>] [--print-id]
          <kind> [pipeline-args...] [<instructions>]
```

- `<kind>` ∈ `{ghgremlin, localgremlin, bossgremlin}`.
- `[pipeline-args...]` are everything before the first positional
  (typically `--plan <path|ref>` or model-selector flags like
  `-a opus`). The launcher captures and persists them so `--resume`
  rehydrates the same config.
- `<instructions>` is the trailing positional (mutually exclusive
  with `--plan` in the pipeline args; the launcher enforces this).
- `--description <phrase>` is an explicit human-readable description
  (≤60 char). Falls back through the chain below if absent.
- `--parent <boss-id>` links a child to its boss.
- `--print-id` writes only the gremlin id to stdout, sends all
  other output to stderr.

### CLI surface (resume path)

```
launch.sh --resume <gremlin-id> [--print-id]
```

Re-spawns an existing gremlin from its recorded stage via
`--resume-from <stage>` on the pipeline. Used by
`/gremlins rescue`'s relaunch step.

### Setup work (forward path)

1. **Argument validation**: kind in the allowed set; jq + claude on
   PATH; gh on PATH if kind is ghgremlin; --plan and positional
   instructions mutually exclusive.
2. **Slug + description resolution**:
   - Slugify the priority-first source: explicit `--description` →
     `--plan` file's H1 → `--plan` file's basename → first
     positional file's H1/basename → `--plan` issue ref → first
     80 chars of remaining args. Result is `[a-z0-9-]+`, max 40
     chars, prefer trim at hyphen boundary, falls back to literal
     `gremlin` if empty.
   - Description fallback: explicit `--description` → `--plan` H1
     (truncated to 60) → ghgremlin/bossgremlin issue ref leaves
     description empty (the orchestrator fills it in after
     resolving the issue) → first 60 chars of raw instructions.
3. **GR_ID = `<slug>-<6 hex>`**.
4. **State dir creation** at
   `${XDG_STATE_HOME:-$HOME/.local/state}/claude-gremlins/<GR_ID>/`.
5. **Spec copy**: if the first positional arg is a readable file,
   copy it to `<state_dir>/artifacts/spec.md`.
6. **Worktree setup** via `gremlins/git.py`. Three modes:
   - `localgremlin`: `git worktree add -b bg/localgremlin/<GR_ID>
     <workdir> HEAD`. `setup_kind = "worktree-branch"`.
   - `ghgremlin`: resolve `origin`'s default branch via
     `gh repo view --json defaultBranchRef`; `git fetch origin
     <default>` (explicit refspec — narrow-refspec clones must
     fetch the right branch); `git worktree add --detach
     <workdir> origin/<default>`. `setup_kind = "worktree"`,
     `worktree_base = "origin/<default>"`.
   - `bossgremlin`: `git worktree add --detach <workdir> HEAD`.
     `setup_kind = "worktree"`.
   - Non-git: `mktemp -d`, `cp -a <project_root>/. <workdir>/`.
     `setup_kind = "copy"`.
7. **Instructions sidecar**: write the full untruncated instructions
   to `<state_dir>/instructions.txt`.
8. **Initial state.json** (atomic write, fields all required by
   the rest of the pipeline):
   - `id`, `kind`, `project_root`, `workdir`, `setup_kind`,
     `branch` (empty unless localgremlin), `worktree_base` (empty
     unless ghgremlin), `status: "running"`, `started_at`,
     `instructions` (200-char truncated summary), `description`,
     `description_explicit` (bool), `parent_id`, `pipeline_args`
     (JSON array of all pre-positional flags + values),
     `stage: "starting"`, `pid: null`.
9. **Spawn the pipeline detached** (see "Spawn boundary" below).
10. **Patch state.json with PID**.
11. **Print summary** (or just GR_ID under --print-id).

### Setup work (resume path)

1. Read state.json. Validate kind, workdir exists, gh available
   if ghgremlin.
2. Refuse if status="running" with a live pid, or if `finished`
   marker exists with exit_code=0.
3. Read `instructions.txt` for the prompt; fall back to
   state.json's truncated `instructions` if the sidecar is missing.
4. Read `pipeline_args` from state.json; rehydrate as the leading
   args. If `pipeline_args` contains `--plan`, omit the trailing
   instructions positional (it would just be flag-echo garbage).
5. Patch state.json:
   - `status = "running"`
   - `stage = <recorded stage>` (rewind to "plan" if stage was
     `""` or `"starting"` — the gremlin crashed before the first
     `set_stage` call)
   - delete `exit_code`, `ended_at`, `sub_stage`,
     `stage_updated_at`, `bail_class`, `bail_reason`, `bail_detail`
   - `rescue_count = (rescue_count or 0) + 1`
   - `rescued_at = now`
   - `resumed_from_stage = <stage>`
   - `pid = null`
6. Append a resume header to `<state_dir>/log`.
7. Remove `finished` and `summarized` marker files.
8. Spawn the pipeline detached with `--resume-from <stage>`
   prepended to the pipeline-arg array.
9. Patch state.json with PID.

### Spawn boundary

Bash currently does:

```bash
nohup bash -c 'python3 -m gremlins.cli "$@"; EC=$?; finish.sh "$GR_ID" "$EC"' \
  -- <kind_subcommand> <pipeline_args...> </dev/null >"$STATE_DIR/log" 2>&1 &
```

Python equivalent: the launcher spawns
`python -m gremlins.cli _run-pipeline <gr_id> <kind_subcommand>
<pipeline_args...>` via `subprocess.Popen` with:

- `start_new_session=True` — detach from the caller's session group
  so it survives parent exit.
- `stdin=subprocess.DEVNULL`.
- `stdout` and `stderr` redirected to a log file opened at
  `<state_dir>/log`.
- `cwd=<workdir>`.
- `env` extending current env with: `GR_ID=<gr_id>`,
  `PYTHONSAFEPATH=1`, `PYTHONPATH=$HOME/.claude` (prepended if
  already set).

After Popen, write `proc.pid` to `<state_dir>/pid` and patch
state.json.

## What `finish.sh` does (port targets)

Source: `skills/_bg/finish.sh` (63 lines).

When the spawned pipeline exits:

1. Touch `<state_dir>/finished` first — that's the marker the
   session-summary hook watches; it must precede the visible status
   change so a failed jq update can't wedge the gremlin in
   "running" forever.
2. Patch state.json: `status = "done" if exit_code==0 else "stopped"`,
   `ended_at = <now>`, `exit_code = <int>`.
3. On success only (exit_code==0), and only for ghgremlin /
   localgremlin (not bossgremlin — its detached HEAD holds the
   chain's squash commits with no branch ref), best-effort remove
   the worktree:
   - `git -C <project_root> worktree remove --force <workdir>`
   - `git -C <project_root> worktree prune`

This logic moves entirely into the `_run-pipeline` subcommand's
`try/finally`. No separate exit hook.

### New: `gremlins/launcher.py`

Public functions:

```python
def launch(kind: str, *,
           instructions: str | None = None,
           plan: str | None = None,
           description: str | None = None,
           parent_id: str | None = None,
           pipeline_args: tuple[str, ...] = (),
           ) -> str:
    """Set up state dir + worktree, spawn the pipeline detached,
    return the gremlin id. Synchronous through spawn; does not
    wait for the pipeline to finish."""

def resume(gr_id: str) -> None:
    """Re-spawn the pipeline for an existing gremlin from its
    recorded stage."""
```

Internal helpers (pick names that read well — these are
suggestions):
- `_resolve_description_and_slug(...)` — single helper covering the
  priority chain; returns `(description, description_explicit, slug)`.
- `_build_initial_state(...)` — assembles the state.json blob.
- `_spawn_pipeline(state_dir, workdir, gr_id, kind_subcommand,
  pipeline_args, env_extras)` — the Popen call.

The slugify/description chain in `launch.sh` accumulated
incrementally and contains a fair amount of dead code (e.g. the
"first positional file H1" branch is rarely hit). Simplify
aggressively — backward compat is not required.

### New: `gremlins/cli.py` `_run-pipeline` subcommand

```
python -m gremlins.cli _run-pipeline <gr_id> <kind_subcommand>
                                     <pipeline_args...>
```

Body:

```python
def _run_pipeline_main(argv):
    gr_id, kind_subcommand, *args = argv
    rc = 1
    try:
        rc = main([kind_subcommand, *args])  # dispatch to the orchestrator
    except SystemExit as e:
        rc = e.code if isinstance(e.code, int) else 1
    except BaseException:
        rc = 1
        traceback.print_exc()
    finally:
        _write_terminal_state(gr_id, exit_code=rc)
    sys.exit(rc)
```

`_write_terminal_state` does what `finish.sh` does: touch the
`finished` marker, patch state.json terminal fields, conditionally
remove the worktree. Importable from `gremlins.launcher` (or a
private `gremlins/_finish.py` if `launcher.py` gets unwieldy).

The leading underscore on `_run-pipeline` signals it's not for
direct human invocation. Document in `cli.py`'s top-of-file
docstring, omit from the dispatch usage line.

### New: `gremlins/cli.py` `launch` subcommand

```
python -m gremlins.cli launch <kind> [--plan <path|ref>]
       [--description <phrase>] [--parent <boss-id>] [--print-id]
       [pipeline-args...] [<instructions>]
```

Argparse front-end → `launcher.launch(...)` → print summary or just
GR_ID.

A `resume` subcommand (`python -m gremlins.cli resume <gr_id>`) is
the cleanest shape for the resume path — separate from `launch` so
neither subcommand has to mode-switch on a flag. (The bash version
overloaded `--resume` onto `launch.sh` for shell-script reasons
that don't apply.)

### Extended: `gremlins/git.py`

Add the three worktree-setup branches. Suggested API:

```python
def setup_worktree_branch(project_root, gr_id, branch_prefix="bg/localgremlin")
    -> (workdir, branch_name)
    # localgremlin: git worktree add -b <prefix>/<gr_id> <tmp> HEAD

def setup_detached_worktree(project_root, base_ref) -> workdir
    # bossgremlin or ghgremlin: git worktree add --detach <tmp> <base>

def setup_copy(project_root) -> workdir
    # non-git fallback: mktemp -d + cp -a

def remove_worktree(project_root, workdir) -> None
    # finish-time cleanup
```

Plus a helper to resolve `origin`'s default branch via
`gh repo view --json defaultBranchRef -q .defaultBranchRef.name`
(needed by ghgremlin's setup). `git.py` already shells out to
`git`/`gh`; this fits.

### Rewired: `gremlins/orchestrators/boss.py`

`launch_child` (lines 346–380, current):

```python
def launch_child(gr_id: str, launch_kind: str, child_plan: str) -> str:
    launcher = os.path.expanduser("~/.claude/skills/_bg/launch.sh")
    if not os.access(launcher, os.X_OK):
        die(f"launch.sh not executable at {launcher}")
    cmd = [launcher, "--parent", gr_id, "--print-id", launch_kind,
           "--plan", child_plan]
    proc = subprocess.Popen(cmd, stdout=subprocess.PIPE, text=True)
    # ... wait, parse stdout, return id ...
```

Becomes:

```python
from ..launcher import launch as _launch

def launch_child(gr_id: str, launch_kind: str, child_plan: str) -> str:
    return _launch(launch_kind, plan=child_plan, parent_id=gr_id)
```

The `_current_proc` global tracking and signal forwarding around
the Popen call go away — `launch()` returns synchronously after
spawning, so there's no long-running child to track at this layer.
The detached pipeline gets its own SIGTERM handling internally;
the boss tracks gremlin lifecycle via state.json + fleet stop, not
via an open Popen handle.

While here, drop the docstring's references to `launch.sh`
(lines 6–11 — "Receives pipeline args forwarded by launch.sh",
"GR_ID env var is set by launch.sh"). Update to "the launcher" or
"the launch entry point".

### Rewired: `gremlins/fleet/rescue.py`

`do_rescue`'s relaunch step (around line 850, current):

```python
launcher = os.path.expanduser("~/.claude/skills/_bg/launch.sh")
if not os.access(launcher, os.X_OK):
    # bail with relaunch_launcher_missing
proc = subprocess.run([launcher, "--resume", gr_id, "--print-id"], ...)
```

Becomes:

```python
from gremlins.launcher import resume as _resume
try:
    _resume(gr_id)
except Exception as e:
    # bail with relaunch_failed (or similar)
```

The `relaunch_launcher_missing` bail class becomes irrelevant (the
launcher is in the same Python process now). Keep the
`relaunch_failed` path for genuine spawn failures.

### Deleted

- `skills/_bg/launch.sh`
- `skills/_bg/finish.sh`

After deletion, `skills/_bg/` still contains `set-stage.sh`,
`set-bail.sh`, `liveness.sh`, `session-summary.sh` (phase 5
retires those).

### `gremlins/tests/test_launch_sh.py` (346 lines)

Currently tests `launch.sh` directly via subprocess. Two options:

1. **Port** — rewrite to drive `gremlins.launcher.launch` /
   `resume` directly. Most assertions are about state.json fields,
   worktree shape, and detached-process behavior — all reachable
   from Python without spawning bash.
2. **Replace** — write fresh tests in `test_launcher.py` that
   exercise the same surfaces. Probably faster than translating
   line-by-line.

Pick replace unless port is straightforwardly mechanical. The new
test file should cover:
- All three setup_kind branches (worktree-branch, worktree,
  copy) plus the non-git fallback.
- Slug and description fallback chain (cover each priority
  level).
- pipeline_args round-trip through state.json.
- The resume path: state.json patches, `finished`/`summarized`
  marker removal, rescue_count increment.
- Refusal to resume a still-running or already-finished gremlin.
- The `_run-pipeline` subcommand: terminal state on success, on
  failure (non-zero exit), on uncaught exception. Worktree
  cleanup on success-for-non-bossgremlin.

### `gremlins/tests/test_localgremlin_shell.py` (140 lines)

Same disposition — these test the launcher's invocation of
`localgremlin`. After the port, they're either ported to drive
`launcher.launch("localgremlin", ...)` or replaced.

### `gremlins/tests/test_ghgremlin_shell.py` (376 lines)

Same. The ghgremlin path needs `gh` available; the existing tests
likely stub `gh repo view`. Mirror that approach with the Python
launcher.

### Other tests

The `launcher` import will surface in tests that previously
monkeypatched `skills/_bg/launch.sh` paths. Search for `launch.sh`
references across `gremlins/tests/`; update each.

`tests/fixtures/shell_env.py` lines 50/53 mention `skills` for
symlink setup — those exist for the shell-integration tests.
Remove if no remaining test needs them; otherwise leave for
phase 5.

### Test-suite-level

Add a `tests/test_launcher.py` (or expand the replacement file
chosen above) that asserts:
- `launch()` returns a well-formed GR_ID.
- The spawned subprocess has `start_new_session=True` semantics
  (its pgid differs from the parent's).
- Concurrent launches don't collide on GR_ID.
- `_run-pipeline` writes terminal state even when the orchestrator
  raises.

## Verification

- `python -m gremlins.cli launch localgremlin --plan
  /tmp/test-plan.md` spawns a detached background gremlin,
  returns its id on stdout, and the gremlin runs through plan →
  implement → review-code → address-code writing terminal state on
  exit. Manual confirmation that the parent shell can close
  without killing the gremlin.
- `python -m gremlins.cli resume <id>` of a deliberately-failed
  gremlin (kill the pipeline mid-stage to leave it in a bailable
  shape) successfully re-spawns it and the rescue_count
  increments.
- Boss chain end-to-end: a 2-step boss spec runs through both
  child gremlins and lands cleanly.
- `ls skills/_bg/` shows only the four scripts that survive
  (`set-stage.sh`, `set-bail.sh`, `liveness.sh`,
  `session-summary.sh`).
- `grep -n "launch.sh\|finish.sh" gremlins/` returns nothing.
- `scripts/sync.sh push` propagates the deletions cleanly. Two
  files at `DELETE_THRESHOLD=5` is fine without `--force`.

## Optional sub-slicing

If the diff is too large for one PR, split into:

- **2a**: Stand up `launcher.py` + `_run-pipeline` subcommand +
  new tests, leaving `launch.sh` / `finish.sh` in place. Existing
  callers still use the bash launcher.
- **2b**: Switch `boss.launch_child` and `rescue.do_rescue` to the
  Python API.
- **2c**: Add the public `launch` (and `resume`) CLI subcommands.
  Delete `launch.sh` and `finish.sh`. Retire the
  shell-integration tests.

The implementing gremlin can decide based on diff size after 2a
whether to continue or bail to a separate child.

## Non-goals for this phase

- Touching `gremlins/state.py`, `set-stage.sh`, `set-bail.sh`,
  or any `set_stage`/`emit_bail` call site (phase 1, complete).
- Touching the `address_code` prompt's `set-bail.sh` invocation
  (phase 3).
- Rewriting the rescue prompt's `~/.claude/skills/` narration
  (phase 4).
- Removing `GREMLIN_SCRIPTS` from `fleet/resolve.py` (phase 4).
- Stripping `skills/` references from module docstrings or
  CLAUDE.md beyond the boss/rescue files this phase touches —
  phase 4 sweeps the rest.
- Porting `liveness.sh` or `session-summary.sh` (phase 5).
- Renaming any state.json fields. The on-disk vocabulary is
  preserved.

xbrianh:
## Summary

- Introduces `gremlins/launcher.py` with `launch()`, `resume()`, and `write_terminal_state()`, replacing the bash launcher and finish scripts entirely
- Adds worktree/copy setup helpers to `gremlins/git.py` (`setup_worktree_branch`, `setup_detached_worktree`, `setup_copy`, `remove_worktree`, etc.)
- Adds `_run-pipeline` (spawn boundary), `launch`, and `resume` subcommands to `gremlins/cli.py`
- Rewires `boss.launch_child` and `rescue.do_rescue` relaunch step to call the Python API directly instead of shelling out to `launch.sh`
- Deletes `skills/_bg/launch.sh` and `skills/_bg/finish.sh`; replaces three shell integration test files with `test_launcher.py` (19 Python-native tests)

## Test plan

- [ ] `cd gremlins && PYTHONPATH=. python -m pytest` — all 197 tests pass
- [ ] `skills/_bg/` contains only `liveness.sh`, `session-summary.sh`, `set-stage.sh`, `set-bail.sh`
- [ ] `python -m gremlins.cli launch localgremlin --plan /tmp/test-plan.md` spawns a detached gremlin and returns its id
- [ ] `python -m gremlins.cli resume <id>` of a failed gremlin re-spawns it with incremented rescue_count
- [ ] Boss chain end-to-end smoke test passes

Closes #176


gremlins/cli.py
```diff
@@ -54,9 +59,128 @@ def main(argv: Optional[List[str]] = None) -> int:
...
+    p.add_argument("--plan", default=None)
+    p.add_argument("--description", default=None)
+    p.add_argument("--parent", dest="parent_id", default=None)
+    p.add_argument("--print-id", action="store_true")
+    args, rest = p.parse_known_args(argv)
+
+    # Separate trailing instructions positional from pipeline flags
+    instructions = None
+    pipeline_flags = list(rest)
+    if pipeline_flags and not pipeline_flags[-1].startswith("-"):
```
xbrianh: **Bug**: this heuristic misparses when a pipeline flag's *value* (not instructions) is the last `rest` element and doesn't start with `-`.

Example: `launch ghgremlin --model claude-opus-4-7` (no instructions) leaves `rest = ["--model", "claude-opus-4-7"]`. The check sees `"claude-opus-4-7"` and promotes it to `instructions`, leaving `"--model"` as an orphaned pipeline flag with no value.

This is undetected because every test in `test_launcher.py` calls `launcher.launch()` directly — nothing exercises `_launch_main`.

Suggested fix: add a proper `--instructions` / `-c` argument to the argparse spec rather than inferring it from positional position, or walk `rest` checking for flag/value pairs to confirm the last item is truly standalone.

xbrianh (commented):
Solid port overall — the state contracts are preserved, the test suite covers the launcher API well, and the boss/rescue rewiring is clean. Three issues worth addressing before merge:

1. **Breaking**: All three SKILL.md files (`skills/ghgremlin/SKILL.md`, `skills/localgremlin/SKILL.md`, `skills/bossgremlin/SKILL.md`) still have `allowed-tools: Bash(~/.claude/skills/_bg/launch.sh:*)` and tell Claude to invoke `launch.sh`. Since `launch.sh` is deleted by this PR, `/ghgremlin`, `/localgremlin`, and `/bossgremlin` will fail immediately after merge. These need to be updated to reference `python -m gremlins.cli launch` (and their `allowed-tools` updated accordingly). The comment in `skills/ghgremlin/ghgremlin.sh` lines 4–5 is also stale ("this file exists only so launch.sh can still resolve...").

2. **Bug in `_launch_main`**: The trailing-positional detection heuristic silently misparses when the last `rest` element is a flag value rather than instructions (e.g. `--model claude-opus-4-7` with no instructions → "claude-opus-4-7" becomes `instructions`, `--model` becomes an orphaned pipeline flag). No tests exercise `_launch_main` directly — all tests call `launcher.launch()` — so this goes undetected.

3. **Missing regression test**: `test_pipeline_survives_worktree_pipeline_rename` (PYTHONSAFEPATH guard that the pipeline survives a worktree rename during implement) was in the deleted `test_localgremlin_shell.py` and has no equivalent in `test_launcher.py`.

gremlins/launcher.py
```diff
@@ -0,0 +1,466 @@
...
+    exit_code = state.get("exit_code")
+
+    if kind not in VALID_KINDS:
+        raise RuntimeError(f"invalid kind in state.json: {kind!r}")
+    if not workdir or not os.path.isdir(workdir):
+        raise RuntimeError(f"worktree missing: {workdir}")
+    if kind == "ghgremlin" and shutil.which("gh") is None:
+        raise RuntimeError("gh CLI not found on PATH (required for ghgremlin)")
+
+    if status == "running" and old_pid is not None and old_pid != "null":
```
xbrianh: `old_pid != "null"` is dead code: `json.loads` converts JSON `null` to Python `None`, which the preceding `old_pid is not None` already catches. The string `"null"` can never appear here. Safe to remove.

gremlins/orchestrators/boss.py
```diff
@@ -344,29 +344,14 @@ def run_handoff(gr_id: str, state_dir: str, boss_state: dict,
...
-
-    child_id = stdout.strip()
-    if not child_id:
-        die("launch.sh --print-id produced no output")
+        child_id = _launch(launch_kind, plan=child_plan, parent_id=gr_id)
+    except (ValueError, RuntimeError) as exc:
+        die(f"launcher failed: {exc}")
 
     log(f"child launched: {child_id}")
     return child_id
```
xbrianh, line 358: The old `launch_child` called `check_stop()` after `proc.communicate()` to handle a SIGTERM that arrived while waiting for the launch subprocess to print the child ID. That call is gone here. Since `_launch()` now returns immediately, the window is tiny, but `_stop_requested` could be set between `_launch()` returning and the next `check_stop()` in the chain. Consider adding `check_stop()` here for consistency with the rest of the boss loop.

gremlins/tests/test_launcher.py
```diff
@@ -0,0 +1,458 @@
...
+        f"{(state_dir / 'log').read_text(errors='replace')[-2000:]}"
+    assert state["exit_code"] == 0
+    # Check stage ordering from fake claude log
+    if lenv.fake_claude_log.exists():
+        log = [json.loads(line) for line in lenv.fake_claude_log.read_text().splitlines() if line.strip()]
+        stages = [e["stage"] for e in log]
+        assert stages[0] == "plan", stages
+        assert "implement-local" in stages, stages
+        assert "review" in stages, stages
+        assert "address" in stages, stages
```
xbrianh, line 472: The deleted `test_localgremlin_shell.py` included `test_pipeline_survives_worktree_pipeline_rename`, which verified that `PYTHONSAFEPATH=1` lets the pipeline survive when the worktree's `gremlins/` directory is renamed during the implement stage (a real regression). That test had no dependency on the bash launcher and would port cleanly here.

gremlins/cli.py
```diff
@@ -54,9 +59,128 @@ def main(argv: Optional[List[str]] = None) -> int:
...
+        instructions = pipeline_flags[-1]
+        pipeline_flags = pipeline_flags[:-1]
```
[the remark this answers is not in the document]
xbrianh: Fixed: replaced the trailing-positional heuristic with an explicit `--instructions` / `-c` argument in `_launch_main`. No more risk of a flag value being promoted to instructions.

gremlins/launcher.py
```diff
@@ -0,0 +1,466 @@
...
+            cwd=workdir,
+            stdin=subprocess.DEVNULL,
+            stdout=log_fh,
+            stderr=log_fh,
+            start_new_session=True,
+            env=env,
+        )
+    finally:
+        log_fh.close()
+    return proc
```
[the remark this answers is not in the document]
xbrianh: Fixed: `_spawn_pipeline` now accepts a `log_mode` parameter (default `"w"` for first launch). `resume()` passes `log_mode="a"` so the prior run output and the resume header appended just before spawn are preserved.

gremlins/launcher.py
```diff
@@ -0,0 +1,466 @@
...
+                capture_output=True, text=True,
+            )
+            if fr.returncode != 0:
+                raise RuntimeError(
+                    f"git fetch origin {default_branch} failed: {fr.stderr.strip()}"
+                )
+            worktree_base = f"origin/{default_branch}"
+            setup_kind = "worktree"
+            workdir = _git_mod.setup_detached_worktree(project_root, worktree_base)
+        else:  # bossgremlin
```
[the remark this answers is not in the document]
xbrianh: Added `test_launch_ghgremlin_state_layout` which uses `lenv_with_gh` to launch a ghgremlin and asserts `setup_kind == "worktree"`, `worktree_base == "origin/main"`, and that the worktree directory exists. Also fixed `lenv_with_gh` to re-chdir into the repo after recreating it (was leaving cwd pointing at the deleted directory).

gremlins/cli.py
xbrianh: Fixed: replaced the trailing-positional heuristic with an explicit `--instructions` / `-c` argument. The remaining `rest` items are all passed through as pipeline flags without any guessing.

gremlins/launcher.py
xbrianh: Removed. `json.loads` converts JSON null to Python `None`, so the preceding `old_pid is not None` is the only guard needed.

gremlins/orchestrators/boss.py
xbrianh: Added `check_stop()` at the end of `launch_child` itself so the function is self-contained. The call site already has belt-and-suspenders coverage (pre-launch `check_stop()`, explicit `_stop_requested` guard, post-launch `check_stop()`), but having it in the function makes the contract explicit.

gremlins/tests/test_launcher.py
xbrianh: Ported as `test_pipeline_survives_worktree_pipeline_rename`. It commits a `gremlins/` stub to the repo so the worktree shadows `$HOME/.claude/gremlins/`, sets `FAKE_CLAUDE_RENAME_GREMLINS=1` so the fake claude renames that directory during implement, then asserts the full pipeline still reaches exit 0.

PR #187, opened 2026-04-30
feat: port session-summary hook to Python, retire skills/_bg/ (Phase 5b)

### LINKED ISSUE #186: feat(gremlins): port session-summary to Python, retire skills/_bg/
# Phase 5b: Session-summary port + bash retirement

etc. over reimplementation.*

## Goal

`~/.claude/skills/_bg/` is empty. The `SessionStart` and
`UserPromptSubmit` hooks in `settings.json` invoke
`python -m gremlins.cli session-summary` instead of the bash
script. The Python session-summary entry point consumes the
existing `gremlins.fleet.state.liveness_of_state_file` — no
duplicate classifier remains.

## What `session-summary.sh` does

Source: `skills/_bg/session-summary.sh` (218 lines). Read it once
before starting; the high-level shape:

1. **Hook input parsing.** Reads JSON from stdin (the hook payload)
   and pulls `hook_event_name` and `cwd`. Project root resolution
   priority: `$CLAUDE_PROJECT_DIR` → input `cwd` → `pwd`. Normalize
   to git toplevel when possible.
2. **State walk.** For each state.json under
   `${XDG_STATE_HOME:-$HOME/.local/state}/claude-gremlins/*/`,
   parse fields (`project_root`, `id`, `kind`, `status`, `workdir`,
   `pid`, `exit_code`, `stage`, `description` or `instructions`).
   Skip gremlins whose `project_root` doesn't match the resolved
   project root.
3. **Bucketing.** Each gremlin lands in one of:
   - **finished**: `finished` marker exists, no `summarized` marker,
     no `closed` marker. Will be announced and then `summarized`
     marker written.
   - **running**: `status == "running"` and no `finished` marker.
     Liveness classified via `liveness_of_state_file`.
4. **Per-hook gating.**
   - `SessionStart`: show running + finished blocks.
   - `UserPromptSubmit`: show only finished. If no new finishes,
     emit nothing (don't spam every prompt).
   - Unknown event: show both.
5. **Output.** Build a markdown summary block. Emit two channels:
   - `stdout`: a JSON object
     `{hookSpecificOutput: {hookEventName, additionalContext}}`
     with the full summary prefixed by an `IMPORTANT:` directive
     telling Claude to surface it verbatim.
   - `stderr`: the raw summary (Claude Code routes non-blocking
     hook stderr into the transcript).
6. **Marker writes.** After successful display, touch
   `<state_dir>/summarized` for each newly-announced finished
   gremlin so future hook firings skip it.
7. **Pruning.** Best-effort: remove state dirs whose `closed` marker
   is older than 14 days. Same for `$STATE_ROOT/direct/*` dirs
   older than 14 days (these come from direct CLI invocations and
   never get a `closed` marker).
8. **Opt-out.** `GREMLIN_SKIP_SUMMARY=1` short-circuits the whole
   thing — used by `claude -p` callers (e.g. fleet manager's `land`
   step) that need clean stdout.
9. **Always exits 0.** Hooks must never break a session.

## What `liveness.sh` does

Source: `skills/_bg/liveness.sh` (101 lines). The bash classifier.

Already mirrored in Python at
`gremlins/fleet/state.py:liveness_of_state_file`. Phase 5a's grep
sweep brought production code to zero skills refs, but this
function is still the duplicate the lockstep-warning describes.
Once `liveness.sh` is deleted in this slice, the warning becomes
moot.

### New: `gremlins/fleet/session_summary.py`

Public entry point:

```python
def main(argv) -> int:
    """Run the session-summary hook. Reads hook input from stdin,
    walks state files for the current project, prints the summary
    block (stdout JSON + stderr raw), writes summarized markers,
    runs prune sweeps. Always returns 0."""
```

Internal structure (suggested):

- `_resolve_project_root(hook_input) -> str` — `$CLAUDE_PROJECT_DIR`
  → input `cwd` → `os.getcwd()`, normalized to git toplevel via
  `gremlins.git`'s existing helpers.
- `_collect_gremlins(state_root, project_root)` — walk the state
  files, classify into running / finished, return two lists with
  the fields the renderer needs.
- `_render_summary(running, finished, hook_event)` — build the
  markdown block, return `(directive_summary, raw_summary)`.
- `_emit(hook_event, directive_summary, raw_summary)` — write
  stdout JSON and stderr raw.
- `_mark_summarized(state_dirs)` — touch markers.
- `_prune_old_state(state_root)` — closed-marker sweep + direct/
  sweep.

Reuse:
- `gremlins.fleet.state.liveness_of_state_file` for the classifier.
  No duplicate implementation.
- `gremlins.fleet.state.iter_state_files` for the walk (or a
  smaller helper if iter_state_files does more than this hook
  needs).
- `gremlins.git.git_toplevel` (or whatever the equivalent is) for
  project-root normalization.
- `gremlins.fleet.state.kind_short` if that maps long kind names
  to display abbreviations the hook should show.

Behavior the port preserves byte-for-byte:

- Exact output format. The summary blocks are user-facing at
  session start; format drift breaks user expectations.
- The `summarized` marker semantics (touch after first display).
- The `GREMLIN_SKIP_SUMMARY=1` opt-out.
- Exit-0-always (a hook crash interferes with session start).
- The `IMPORTANT:` directive prefix in the JSON `additionalContext`.
- The two-channel emission (stdout JSON + stderr raw).
- Prune behaviors (closed >14 days, direct/ >14 days).

### New: `session-summary` CLI subcommand in `gremlins/cli.py`

Add alongside the existing `bail` / `launch` / `resume` /
`_run-pipeline` subcommands. Mirror the pattern phase 3's `bail`
established:

```python
if sub == "session-summary":
    from .fleet.session_summary import main as _session_summary_main
    return _session_summary_main(rest)
```

Update the module docstring's subcommand list and the `usage`
string in `main()`.

### `settings.json`: repoint hook entries

Lines 60–79 currently:

```json
"SessionStart": [
  {
    "hooks": [
      {
        "type": "command",
        "command": "$HOME/.claude/skills/_bg/session-summary.sh"
      }
    ]
  }
],
"UserPromptSubmit": [
  {
    "hooks": [
      {
        "type": "command",
        "command": "$HOME/.claude/skills/_bg/session-summary.sh"
      }
    ]
  }
]
```

Both `command` lines change to:

```
PYTHONPATH=$HOME/.claude python -m gremlins.cli session-summary
```

(`PYTHONPATH=$HOME/.claude` ensures the `gremlins` package is
importable when the hook fires from any cwd.)

The user has a `python` not `python3` shell hook, so use `python`
explicitly. Verify by reading
`~/.claude/settings.json` ↔ this repo's `settings.json` and
confirming the binary name in use.

### Deleted

All four remaining files in `skills/_bg/`:
- `set-stage.sh`
- `set-bail.sh`
- `liveness.sh`
- `session-summary.sh`

After deletion, `skills/_bg/` should be empty. Either leave the
empty directory or remove it entirely — pick removal (simpler;
`rsync --delete` will handle the mirror).

Before deleting, repo-wide grep confirmation that no caller
outside `skills/_bg/` itself references `_bg/` paths:

```
grep -rn "_bg/" /Users/xath/Desktop/amorphous-industries/claude/ \
    --exclude-dir=skills --exclude-dir=.git
```

If anything reaches outside this slice's scope, surface it before
deleting.

### New: `gremlins/tests/test_session_summary.py`

Fixture-driven coverage. Suggested cases:

- **No gremlins for this project**: empty state root → no output,
  exit 0.
- **One running gremlin matching project_root**: SessionStart hook
  emits the running block; UserPromptSubmit emits nothing.
- **One newly-finished gremlin**: both hooks emit the finished
  block; `summarized` marker is touched after.
- **Already-summarized finished gremlin**: skipped on subsequent
  calls (no double-announce).
- **Closed gremlin**: never appears.
- **Stalled gremlin (log mtime > BG_STALL_SECS)**: liveness
  reports stalled, summary line says "stalled?".
- **Crashed gremlin (PID gone, no finished marker)**: liveness
  reports `dead:crashed`.
- **`GREMLIN_SKIP_SUMMARY=1`**: short-circuit, no output.
- **Project-root mismatch**: gremlins for a different project_root
  filtered out.
- **Prune behavior**: state dir with `closed` marker older than 14
  days is removed.

Use `tmp_path` for state root; monkeypatch `XDG_STATE_HOME` and
`CLAUDE_PROJECT_DIR`.

### Existing tests

Run the full suite. Anything that monkeypatched
`session-summary.sh` directly should be updated to drive the
Python entry point instead.

## Verification

- `ls /Users/xath/Desktop/amorphous-industries/claude/skills/_bg/`
  returns no files (or the directory doesn't exist).
- `grep -rn "_bg/" /Users/xath/Desktop/amorphous-industries/claude/
  --exclude-dir=.git` returns nothing.
- `grep -ri skills /Users/xath/Desktop/amorphous-industries/claude/gremlins/
  --include="*.py" --include="*.md"` (excluding `DESIGN.md`)
  returns zero hits.
- Manual smoke (after merge):
  - `scripts/sync.sh push --force --yes` propagates the deletions
    (six-file deletion exceeds `DELETE_THRESHOLD=5` — `--dry-run`
    first to confirm scope).
  - Open a fresh Claude Code session in the project. The
    SessionStart hook should fire and report any in-flight
    gremlins via the new Python entry point. (Test by launching a
    short-lived gremlin first if no in-flight gremlins exist.)
  - Submit a prompt. The UserPromptSubmit hook should emit nothing
    if no gremlins finished since session start.
  - Submit a prompt while a gremlin finishes. Both stderr (raw
    block) and the model's reply (verbatim block) should show the
    finished announcement.
- Sanity check the `python` vs `python3` hook: an erroneous
  `python3` invocation gets blocked by the user's PreToolUse hook;
  the `settings.json` lines must use `python`.

## Risks

- **Hook breakage is user-visible every session.** Misformat the
  output, fail to exit 0, or spawn `python` with a missing
  PYTHONPATH and the user sees a broken or empty hook block at
  every session start until reverted. Test thoroughly before
  pushing.
- **`scripts/sync.sh push` deletion threshold.** Six file
  deletions exceeds `DELETE_THRESHOLD=5`. Use `--dry-run` to
  confirm the deletion list matches the slice's scope, then
  `--force --yes`.
- **Out-of-tree callers of the bash scripts.** Repo-wide grep
  catches in-tree callers; anything outside this repo is on its
  own. Document the change in the PR description so anyone with
  external invocations notices.

## Non-goals for this phase

- Touching `DESIGN.md`. It's a migration artifact; whether to
  retire it is a separate decision (out of this consolidation
  chain entirely).
- Renaming any state.json fields. The on-disk vocabulary stays.
- Replacing `claude -p` with a different backend. This was the
  whole motivation, but the consolidation is the *prerequisite* —
  the backend swap is its own work.
- Refactoring the existing fleet manager listing
  (`gremlins/fleet/cli.py` + `views.py`). They already share the
  liveness classifier; this slice doesn't change them.

xbrianh:
## Summary

- Ports `skills/_bg/session-summary.sh` to `gremlins/fleet/session_summary.py`, reusing `liveness_of_state_file` from `fleet/state.py` — eliminates the duplicate liveness classifier warned about in the design doc.
- Adds `session-summary` subcommand to `gremlins/cli.py`.
- Updates both hook entries in `settings.json` to `PYTHONPATH=$HOME/.claude python -m gremlins.cli session-summary`.
- Deletes all four remaining `skills/_bg/` files (`liveness.sh`, `session-summary.sh`, `set-bail.sh`, `set-stage.sh`) and the now-empty directory.
- Adds 23 fixture-driven tests covering running/finished/stalled/crashed/opt-out/prune/project-root-mismatch/JSON-envelope behavior.

## Test plan

- [ ] `cd gremlins && PYTHONPATH=. python -m pytest` — 230/230 pass
- [ ] `ls skills/_bg/` — directory does not exist
- [ ] `grep -rn "_bg/" . --exclude-dir=.git --exclude-dir=skills` — only `settings.json` (old path) and docs; no code callers
- [ ] After `scripts/sync.sh push --dry-run --force`, confirm the 6 deletions are in scope, then push with `--force --yes`
- [ ] Open a fresh Claude Code session; SessionStart hook fires via the Python entry point

Closes #186


gremlins/fleet/session_summary.py
```diff
@@ -0,0 +1,284 @@
...
+        }
+    })
+    sys.stdout.write(output + "\n")
+    sys.stdout.flush()
+
+
+def _mark_summarized(state_dirs: list) -> None:
+    for d in state_dirs:
+        try:
+            open(os.path.join(d, "summarized"), "a").close()
```
xbrianh: `open(..., "a").close()` works (`.close()` is called synchronously), but `pathlib.Path(os.path.join(d, "summarized")).touch()` is the idiomatic Python idiom for an empty marker file. Requires adding `import pathlib` (or `from pathlib import Path`).

gremlins/fleet/session_summary.py
```diff
@@ -0,0 +1,284 @@
...
+            pass
+    return root
+
+
+def _collect_gremlins(state_root: str, project_root: str):
+    """Walk state files and split into running / newly-finished lists.
+
+    Returns (running, finished, newly_summarized_dirs).
+    """
+    from gremlins.fleet.state import liveness_of_state_file
```
xbrianh: Deferred import inside the function body. By the time `_collect_gremlins` runs, `gremlins.fleet` is already in `sys.modules` — it was loaded when `session_summary.py` was imported as part of the fleet package. There's no circular-import reason to defer this; a top-level import at the module level would be clearer.

gremlins/fleet/session_summary.py
```diff
@@ -0,0 +1,284 @@
...
+        else:
+            running_block += (
+                f"- `{g['id']}` ({g['kind']}): running"
+                f" (stage: {stage_disp}, pid {pid_str}){desc_suffix} — log: {g['log']}\n"
+            )
+
+    finished_block = ""
+    for g in finished:
+        desc_suffix = f" — _{g['description']}_" if g["description"] else ""
+        exit_suffix = f" (exit {g['exit_code']})" if g["exit_code"] else ""
```
xbrianh, line 206: `if g["exit_code"]` — the string `"0"` is truthy in Python, so a clean-exit gremlin (exit code 0) renders `(exit 0)` in the finished block. This matches the shell's `${exit_code:+ (exit $exit_code)}` behavior exactly (non-empty string is non-empty), so it's a faithful port. Just noting it's intentional — if only non-zero exits should annotate, the guard would be `g["exit_code"] not in ("", "0")`.

xbrianh (commented):
Clean port — the Python replicates the shell's logic faithfully and the 23-test suite covers the interesting edge cases well. Three minor observations inline; none are blockers.

gremlins/fleet/session_summary.py
xbrianh: Fixed — now uses `pathlib.Path(d, "summarized").touch()`.

gremlins/fleet/session_summary.py
xbrianh: Fixed — moved to a top-level import.

gremlins/fleet/session_summary.py
xbrianh: Intentional — matches the shell's `${exit_code:+ (exit $exit_code)}` behavior exactly (any non-empty string, including "0", renders the suffix). Thanks for the callout.

gremlins/fleet/session_summary.py
```diff
@@ -0,0 +1,284 @@
...
+    )
+    return os.path.join(xdg, "claude-gremlins")
+
+
+def _read_stdin() -> dict:
+    try:
+        if not sys.stdin.isatty():
+            raw = sys.stdin.read()
+            if raw:
+                return json.loads(raw)
```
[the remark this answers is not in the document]
xbrianh: Fixed — `_read_stdin` now checks `isinstance(parsed, dict)` before returning, falling back to `{}` for any non-object JSON.

settings.json
```diff
@@ -62,7 +62,7 @@
         "hooks": [
           {
             "type": "command",
-            "command": "$HOME/.claude/skills/_bg/session-summary.sh"
+            "command": "PYTHONPATH=$HOME/.claude python -m gremlins.cli session-summary"
           }
```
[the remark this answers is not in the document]
xbrianh: Fixed — updated `skills/ghreview/SKILL.md` and `skills/ghaddress/SKILL.md` to use `PYTHONPATH=$HOME/.claude python -m gremlins.cli bail <class> "..."` and updated the `allowed-tools` frontmatter accordingly.

settings.json
```diff
@@ -72,7 +72,7 @@
...
```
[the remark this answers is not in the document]
xbrianh: Fixed — updated README.md to reference `python -m gremlins.cli launch` / `gremlins/fleet/session_summary.py` instead of the retired `skills/_bg/` scripts. DESIGN.md references are historical (describing the migration phases and the pre-migration state), so left as-is.

PR #191, opened 2026-04-30
fix(clients): suppress session-summary hook in SubprocessClaudeClient

### LINKED ISSUE #165: SubprocessClaudeClient.run() should set GREMLIN_SKIP_SUMMARY=1 to prevent hook corruption of text_result
etc. over reimplementation.*

## Problem

`SubprocessClaudeClient.run()` (`gremlins/clients/claude.py:306`)
spawns `claude -p` via `subprocess.Popen(cmd, ...)` with no `env=`
override. Whatever `GREMLIN_SKIP_SUMMARY` happens to be in the
caller's environment is what the child sees. By default, the hook
fires in the child session and injects a "surface this verbatim"
directive that prepends the gremlin status block to the model's
reply, corrupting `CompletedRun.text_result` and `result` events
in stream-json output.

`gremlins/fleet/land.py:346-347` already sets
`GREMLIN_SKIP_SUMMARY=1` for its own one-shot `claude -p` for
exactly this reason. `SubprocessClaudeClient.run()` should do the
same.

## Fix

In `SubprocessClaudeClient.run()`, build a copy of `os.environ`
with `GREMLIN_SKIP_SUMMARY=1` overlaid and pass it via `env=` to
`subprocess.Popen`:

```python
env = os.environ.copy()
env["GREMLIN_SKIP_SUMMARY"] = "1"
p = subprocess.Popen(
    cmd,
    stdout=subprocess.PIPE,
    stderr=None,
    start_new_session=False,
    env=env,
)
```

Always-set is the right shape — every consumer of this client
parses model output programmatically (text_result for plan-title,
stream-json events for stage outcomes, etc.); none of them want
the hook directive. No client-option for it; backward
compatibility is not a concern, simplification is the preference.

While here, look for opportunities to **deduplicate with
`fleet/land.py`**: both call sites overlay the same one-key env
delta on `os.environ`. A small helper
(`gremlins/clients/claude.py` or a new `gremlins/_subprocess.py`)
that returns the suppressed env would be the cleaner shape if
adopted by both. Judge based on what reads well — if the
duplication is the only DRY win, inlining is fine.

## Files

- `gremlins/clients/claude.py` (the fix)
- `gremlins/fleet/land.py` (only if extracting a shared helper)

## Tests

Add to `gremlins/tests/test_clients_claude.py` (or wherever
`SubprocessClaudeClient` tests live; create the file if not
present):

- Spawn `SubprocessClaudeClient.run(...)` against a stub `claude`
  binary in `tests/bin/` that records its env to a file.
- Assert `GREMLIN_SKIP_SUMMARY=1` appears in the recorded env.
- Assert other vars from the parent process pass through unchanged.

If a stub-binary fixture pattern already exists in
`tests/test_launcher.py` or similar, reuse it.

## Verification

- `grep -n "GREMLIN_SKIP_SUMMARY" gremlins/clients/claude.py`
  returns at least one hit.
- Manual: launch a localgremlin, watch
  `<state_dir>/artifacts/stream-*.jsonl` for any `result` event
  whose text begins with the gremlin status block. Should not
  occur after the fix.

## Non-goals

- Reworking the hook itself (`gremlins/fleet/session_summary.py`).
  The opt-out mechanism (`GREMLIN_SKIP_SUMMARY=1`) is the contract
  every claude-parsing caller in this repo should use.
- Adding a configurable opt-in for callers that *want* the hook
  directive. None exist; if one ever does, add it then.

Closes #165.
Ref #164 (where the regression was originally surfaced as a
review comment).

xbrianh:
## Summary

- `SubprocessClaudeClient.run()` now builds `env = os.environ.copy()` with `GREMLIN_SKIP_SUMMARY=1` overlaid and passes it via `env=` to `subprocess.Popen`, matching the pattern already used in `fleet/land.py`.
- Adds `gremlins/tests/test_clients_claude.py` with two tests: one asserting `GREMLIN_SKIP_SUMMARY=1` is present in the child env, one asserting other parent env vars pass through unchanged.

## Test plan

- [ ] `cd gremlins && PYTHONPATH=. python -m pytest tests/test_clients_claude.py` — both new tests pass
- [ ] `cd gremlins && PYTHONPATH=. python -m pytest` — full suite (235 tests) passes with no regressions
- [ ] `grep -n "GREMLIN_SKIP_SUMMARY" gremlins/clients/claude.py` returns a hit

Closes #165

### DIFF
--- a/gremlins/clients/claude.py
+++ b/gremlins/clients/claude.py
@@ -16,6 +16,7 @@ same client without a second pass.
 from __future__ import annotations
 
 import json
+import os
 import pathlib
 import subprocess
 import sys
@@ -303,11 +304,14 @@ class SubprocessClaudeClient:
         # or EOF, it doesn't block for the buffer to fill) and throughput
         # on the big implement-stage stream-json traces jumps by orders of
         # magnitude.
+        env = os.environ.copy()
+        env["GREMLIN_SKIP_SUMMARY"] = "1"
         p = subprocess.Popen(
             cmd,
             stdout=subprocess.PIPE,
             stderr=None,
             start_new_session=False,
+            env=env,
         )
         self._track(p)
 
--- a/gremlins/tests/test_clients_claude.py
+++ b/gremlins/tests/test_clients_claude.py
@@ -1,0 +1,85 @@
+"""Tests for SubprocessClaudeClient (gremlins/clients/claude.py)."""
+
+from __future__ import annotations
+
+import json
+import os
+import shlex
+import sys
+import pathlib
+
+import pytest
+
+from gremlins.clients.claude import SubprocessClaudeClient
+
+TESTS_DIR = pathlib.Path(__file__).resolve().parent
+
+# ---------------------------------------------------------------------------
+# Stub claude binary
+# ---------------------------------------------------------------------------
+
+_STUB_CLAUDE_SRC = """\
+#!/usr/bin/env python3
+import json, os, sys
+
+env_out = os.environ.get("STUB_ENV_OUT")
+if env_out:
+    with open(env_out, "w", encoding="utf-8") as f:
+        json.dump(dict(os.environ), f)
+
+sys.stdout.write(json.dumps({"type": "system", "subtype": "init", "session_id": "stub-sess"}) + "\\n")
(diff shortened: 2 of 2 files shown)


xbrianh (commented):
Clean, minimal fix. The env-copy pattern is consistent with the existing `fleet/land.py` approach, the tests are well-structured, and the stub technique avoids relying on a real `claude` binary. No issues found.

PR #193, opened 2026-04-30
fix(commit-pr): derive prompt from diff instead of resuming implement session

### LINKED ISSUE #181: ghgremlin: commit-pr resume after crash is broken (lost IMPL_SESSION → EmptyImpl loop)
etc. over reimplementation.*

## Symptom

When a ghgremlin crashes at \`commit-pr\` (e.g. transient auth-403,
org usage limit, network flake), \`python -m gremlins.cli resume\`
cannot recover it. The orchestrator silently rewinds
\`--resume-from commit-pr\` → \`--resume-from implement\`
(\`gremlins/orchestrators/gh.py:95-101\`); the implement agent
sees the work is already on the impl-handoff branch and produces
no new commits; \`_run_implement_gh\` raises \`EmptyImpl: refusing
to open empty PR\`. Stuck in a rescue loop.

The other gh stages (\`request-copilot\`, \`ghreview\`,
\`wait-copilot\`, \`ghaddress\`) **do not have this problem** —
each opens its own claude session and resumes cleanly. Confirmed
in the wild on 2026-04-30: gremlin
\`boss-child-base-ref-current-head-a08507\` died at \`ghreview\`
(org usage limit) and resumed cleanly via
\`python -m gremlins.cli resume\`. Only \`commit-pr\` is affected.

## Root cause

\`gremlins/stages/commit_pr.py:87\` calls
\`client.run(..., resume_session=impl_session_id, ...)\` — i.e.
\`claude --resume <session_id>\`. The session id is in-memory
state from the just-completed implement stage, never persisted.
On a crash, that state is gone.

The orchestrator's rewind to \`implement\` is a workaround for
the missing session. It works for fresh runs but loops forever
when the impl-handoff branch already has the diff.

## Fix: drop the cross-stage session dependency

Reimplement \`commit_pr.py\` to compose its prompt from the
**diff on disk**, not from a resumed session.

The information the resumed session carries is the diff, the
commit message intent, and the agent's understanding of "what
was just implemented." All three are recoverable from the
impl-handoff branch:

- diff: \`git diff <base_ref>..<impl_handoff_branch>\` or
  \`git log -p <base_ref>..HEAD\`
- summary: included in the prompt (the action clauses already
  vary by \`ImplOutcome\` — \`HeadAdvanced\`, \`DirtyOnly\`,
  empty-handoff fallback — and that classification is enough
  context for a fresh agent to write a commit message).
- intent: the spec / issue body, also on disk in
  \`<state_dir>/spec.md\` or via \`gh issue view\`.

The new \`run_commit_pr_stage\` signature:

\`\`\`python
def run_commit_pr_stage(
    *,
    client: ClaudeClient,
    model: str,
    impl_outcome: ImplOutcome,        # was: also impl_session_id
    impl_handoff_branch: str,
    base_ref: str,
    issue_url: str,
    cwd: str,
    session_dir: pathlib.Path,
) -> str:
    # No resume_session= argument.
    ...
\`\`\`

The stage prompt now begins with the diff and a short
classification note, and asks the agent to produce a commit
message + open the PR.

\`stages/implement.py\` no longer needs to return \`session_id\`
in \`ImplStageResult\` (or the caller can stop passing it
through). One less piece of cross-stage in-memory state.

The orchestrator's \`commit-pr\` rewind hack at
\`gh.py:95-101\` is deleted. \`--resume-from commit-pr\` now
means exactly what it says.

## Files

- \`gremlins/stages/commit_pr.py\`: rewrite to take diff context
  instead of \`impl_session_id\`.
- \`gremlins/stages/implement.py\`: optional — drop \`session_id\`
  from \`ImplStageResult\` (or leave it for telemetry; only
  remove if it has no other use).
- \`gremlins/orchestrators/gh.py\`: delete the lines 95-101
  rewind. Update the \`commit-pr\` stage wiring at lines 404-410
  to drop the \`impl_session_id\` thread.
- \`gremlins/prompts/\`: update the three commit-pr prompt
  templates (\`HeadAdvanced\`, \`DirtyOnly\`, empty-handoff) to
  open with the diff rather than continuing-the-session
  framing. Read carefully before editing — the action clauses
  encode subtle invariants per \`ImplOutcome\`.
- \`gremlins/CLAUDE.md\` and \`gremlins/stages/CLAUDE.md\`: drop
  the "commit-pr resumes the implement session" framing. Both
  currently say "commit-pr resumes the implement session" — that
  becomes "commit-pr opens its own session against the
  impl-handoff branch's diff."

## Tests

- \`gremlins/tests/test_stages_local.py\` (or wherever commit-pr
  tests live — likely \`test_orchestrator_gh.py\`): drop the
  \`resume_session\` assertions and add a positive test that
  asserts commit-pr's prompt contains the diff content from a
  fixture impl-handoff branch.
- New regression test: drive the gh orchestrator with
  \`--resume-from commit-pr\`. Pre-seed an impl-handoff branch
  with a known diff. Stub gh CLI subprocesses. Assert the
  pipeline completes through \`ghaddress\` without re-running
  implement and without raising \`EmptyImpl\`.
- \`gremlins/tests/conftest.py\` \`MINIMAL_EVENTS\` /
  \`_issue_events\`: drop the \`session-commit-pr\` reuse if it
  was depending on the resume contract.

## Verification

- \`cd gremlins && PYTHONPATH=. python -m pytest\` passes.
- \`grep -n "impl_session_id\\|resume_session" gremlins/\` returns
  far fewer hits — only telemetry / debug log uses, if any.
- Manual: launch a ghgremlin, kill it during \`commit-pr\`
  (e.g. \`/gremlins stop\`), run
  \`python -m gremlins.cli resume <id>\`. The pipeline picks up
  at \`commit-pr\` and runs through to \`ghaddress\` against the
  existing impl-handoff branch. No \`EmptyImpl\` loop.

## Alternatives considered

1. **Persist session_id to state.json.** Adds disk state and
   depends on \`claude --resume <id>\` working across process
   boundaries / time gaps — brittle. Rejected.
2. **Refuse the rewind with a clear error.** Cheaper than option
   2 but doesn't auto-recover; the operator still has to push
   the impl-handoff branch and open the PR by hand (as I did
   for #180). Rejected.

## Notes

- The consolidation chain has finished, so this issue is no
  longer blocked behind anything. Open work: just this fix.
- The verification "Same root cause likely affects
  request-copilot, ghreview, wait-copilot, ghaddress" from the
  original filing is now answered — those stages do NOT have
  this problem (each opens its own session).

Closes #181.

xbrianh:
## Summary

- Drops the cross-stage `session_id` dependency in the gh pipeline's `commit-pr` stage — the stage now computes the diff from the impl-handoff branch and runs a fresh claude session with no `--resume`.
- Removes the `--resume-from commit-pr → implement` rewind hack in `gh.py` that caused an `EmptyImpl` rescue loop when the impl-handoff branch was already present.
- Persists `impl_handoff_branch` / `impl_base_ref` to `state.json` after implement so `--resume-from commit-pr` can reconstruct the outcome and diff from disk without re-running implement.

## Test plan

- [ ] `cd gremlins && PYTHONPATH=. python -m pytest` — 236 tests pass
- [ ] `test_parse_resume_from_commit_pr` asserts no rewind message for `--resume-from commit-pr`
- [ ] `test_resume_from_commit_pr_skips_implement` drives `gh_main --resume-from commit-pr` against a pre-seeded impl-handoff branch, asserts implement is skipped, diff content appears in prompt, and `resume_session` is `None`
- [ ] `grep -n "impl_session_id\|resume_session" gremlins/**/*.py` returns only client protocol and the new `assert resume_session is None` assertion

Closes #181.

### DIFF
--- a/gremlins/orchestrators/gh.py
+++ b/gremlins/orchestrators/gh.py
@@ -26,6 +26,7 @@ from typing import Dict, List, Optional
 
 from ..clients.claude import ClaudeClient, SubprocessClaudeClient
 from ..gh_utils import extract_gh_url, get_repo, parse_issue_ref, view_issue
+from ..git import DirtyOnly, HeadAdvanced
 from ..runner import install_signal_handlers, run_stages
 from ..stages.commit_pr import run_commit_pr_stage
 from ..stages.ghaddress import run_ghaddress_stage
@@ -90,15 +91,6 @@ def _parse_gh_args(argv: List[str]) -> argparse.Namespace:
     parser.add_argument("--model", dest="model", default=None)
     parser.add_argument("instructions", nargs="*")
     args = parser.parse_args(argv)
-
-    # --resume-from commit-pr: rewind to implement (IMPL_SESSION not persisted)
-    if args.resume_from == "commit-pr":
-        sys.stderr.write(
-            "    note: --resume-from commit-pr requires IMPL_SESSION which isn't "
-            "persisted; rewinding to implement\n"
-        )
-        sys.stderr.flush()
-        args.resume_from = "implement"
 
     if args.resume_from is not None and args.resume_from not in VALID_STAGES:
         die(
@@ -399,19 +391,49 @@ def gh_main(argv: List[str], *, client: Optional[ClaudeClient] = None) -> int:
             issue_num=issue_num,
         )
         impl_result_holder["result"] = result
+        # Persist for commit-pr resume: base_ref and handoff branch are all
+        # that's needed to reconstruct the diff and outcome on a fresh process.
+        patch_state(
+            impl_handoff_branch=result.handoff_branch,
+            impl_base_ref=result.pre_state.head,
+        )
 
     def stage_commit_pr() -> None:
         set_stage("commit-pr")
         print("==> [2b/6] committing + opening PR", flush=True)
-        result: ImplStageResult = impl_result_holder["result"]  # type: ignore[assignment]
+
+        if "result" in impl_result_holder:
+            result: ImplStageResult = impl_result_holder["result"]  # type: ignore[assignment]
+            impl_outcome = result.outcome
+            impl_handoff_branch = result.handoff_branch
+            base_ref = result.pre_state.head
+        else:
+            # Resuming at commit-pr: reconstruct from state.json
+            impl_handoff_branch = _read_state_field(state_file, "impl_handoff_branch")
+            base_ref = _read_state_field(state_file, "impl_base_ref")
+            if not base_ref:
+                die(
+                    "--resume-from commit-pr: no impl_base_ref in state.json "
+                    "(rewind to implement?)"
+                )
+            if impl_handoff_branch:
+                count_r = subprocess.run(
+                    ["git", "rev-list", "--count", f"{base_ref}..{impl_handoff_branch}"],
+                    capture_output=True, text=True, check=False,
+                )
+                commit_count = int(count_r.stdout.strip() or "0")
+                impl_outcome = HeadAdvanced(commit_count=commit_count)
+            else:
+                impl_outcome = DirtyOnly()
+
         pr_url = run_commit_pr_stage(
             client=client,
             model=model,
-            impl_session_id=result.session_id,
-            pre_state=result.pre_state,
-            outcome=result.outcome,
-            handoff_branch=result.handoff_branch,
-            issue_num=issue_num,
+            impl_outcome=impl_outcome,
+            impl_handoff_branch=impl_handoff_branch,
+            base_ref=base_ref,
+            issue_url=issue_url,
+            cwd=None,
             session_dir=session_dir,
         )
         pr_url_holder["url"] = pr_url
--- a/gremlins/orchestrators/gh.py
+++ b/gremlins/orchestrators/gh.py
@@ -421,7 +421,13 @@ def gh_main(argv: List[str], *, client: Optional[ClaudeClient] = None) -> int:
                     ["git", "rev-list", "--count", f"{base_ref}..{impl_handoff_branch}"],
                     capture_output=True, text=True, check=False,
                 )
-                commit_count = int(count_r.stdout.strip() or "0")
+                if count_r.returncode != 0:
+                    die(
+                        f"--resume-from commit-pr: impl_handoff_branch '{impl_handoff_branch}' "
+                        f"not found or base_ref invalid (rewind to implement?)\n"
+                        f"{count_r.stderr.strip()}"
+                    )
+                commit_count = int(count_r.stdout.strip())
                 impl_outcome = HeadAdvanced(commit_count=commit_count)
             else:
                 impl_outcome = DirtyOnly()
--- a/gremlins/stages/commit_pr.py
+++ b/gremlins/stages/commit_pr.py
@@ -38,16 +38,19 @@ def _get_diff(
             ["git", "log", "--patch", f"{base_ref}..{impl_handoff_branch}"],
             **run_kw,
         )
-        diff = r.stdout.strip()
-        if not diff:
-            r = subprocess.run(
-                ["git", "diff", f"{base_ref}..{impl_handoff_branch}"],
-                **run_kw,
+        if r.returncode != 0:
+            raise RuntimeError(
+                f"git log --patch {base_ref}..{impl_handoff_branch} failed "
+                f"(rc={r.returncode}): {r.stderr.strip()}"
             )
-            diff = r.stdout.strip()
+        diff = r.stdout.strip()
     else:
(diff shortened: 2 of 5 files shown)


xbrianh (commented):
Clean removal of the `session_id` dependency. The state-persistence + resume reconstruction approach is correct and the regression test is well-designed. One correctness gap in the resume path, and one minor note on the diff fallback.

gremlins/orchestrators/gh.py
```diff
@@ -399,19 +391,49 @@ def stage_implement() -> None:
...
+                die(
+                    "--resume-from commit-pr: no impl_base_ref in state.json "
+                    "(rewind to implement?)"
+                )
+            if impl_handoff_branch:
+                count_r = subprocess.run(
+                    ["git", "rev-list", "--count", f"{base_ref}..{impl_handoff_branch}"],
+                    capture_output=True, text=True, check=False,
+                )
+                commit_count = int(count_r.stdout.strip() or "0")
```
xbrianh: Silent fallback to `commit_count=0` when `git rev-list` fails (e.g. `impl_handoff_branch` was deleted or `base_ref` is invalid). With `count=0`, `_get_diff` then runs `git log --patch` against a branch that doesn't exist, produces empty output, falls back to `git diff`, also empty, and the agent receives the contradictory combination of an `HeadAdvanced(commit_count=0)` action clause and `(no diff available)` in the prompt.

Suggest dying explicitly on failure:
```python
if count_r.returncode != 0:
    die(
        f"--resume-from commit-pr: impl_handoff_branch '{impl_handoff_branch}' "
        "not found or base_ref invalid (rewind to implement?)"
    )
commit_count = int(count_r.stdout.strip())
```

gremlins/stages/commit_pr.py
```diff
@@ -23,45 +24,73 @@ def _load(name: str) -> str:
...
+    run_kw: dict = dict(capture_output=True, text=True, check=False)
+    if cwd:
+        run_kw["cwd"] = cwd
+    if isinstance(outcome, HeadAdvanced):
+        r = subprocess.run(
+            ["git", "log", "--patch", f"{base_ref}..{impl_handoff_branch}"],
+            **run_kw,
+        )
+        diff = r.stdout.strip()
+        if not diff:
```
xbrianh: The `git log --patch` → `git diff` fallback is unlikely to be exercised: if there are commits (which there are for `HeadAdvanced`), `git log --patch base..branch` should always produce output; if the branch doesn't exist, both commands fail and both return empty. The only realistic case where `git log --patch` is empty but `git diff` is not would be a pathological empty-tree merge. Consider removing the fallback and letting the `"(no diff available)"` sentinel on line 52 serve as the sole empty-output guard — the two-step adds complexity without a realistic trigger.

gremlins/orchestrators/gh.py
xbrianh: Fixed. The  call now checks  and calls  with stderr on failure, so a missing branch or invalid base_ref surfaces immediately rather than silently producing .

gremlins/stages/commit_pr.py
xbrianh: Agreed. Removed the `git log --patch` → `git diff` fallback. `git log --patch` now either succeeds (non-empty output for `HeadAdvanced`) or fails (returncode non-zero → raises `RuntimeError`). The `"(no diff available)"` sentinel on line 52 remains as the sole empty-output guard.

gremlins/stages/commit_pr.py
```diff
@@ -23,45 +24,73 @@ def _load(name: str) -> str:
...
+            r = subprocess.run(
+                ["git", "diff", f"{base_ref}..{impl_handoff_branch}"],
+                **run_kw,
+            )
+            diff = r.stdout.strip()
+    else:
+        # DirtyOnly: uncommitted changes relative to HEAD
+        r = subprocess.run(["git", "diff", "HEAD"], **run_kw)
```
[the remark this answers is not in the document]
xbrianh: Fixed. Both git commands in `_get_diff` now check `returncode` and raise `RuntimeError` with rc + stderr on failure. The unreachable `git log --patch` → `git diff` fallback was removed at the same time.

gremlins/orchestrators/gh.py
```diff
@@ -399,19 +391,49 @@ def stage_implement() -> None:
...
+                impl_outcome = HeadAdvanced(commit_count=commit_count)
+            else:
+                impl_outcome = DirtyOnly()
```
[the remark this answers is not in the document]
xbrianh: Fixed. The `git rev-list --count` call now checks `returncode` and calls `die(...)` with stderr on non-zero exit, so a missing branch or invalid base_ref surfaces immediately rather than silently producing `commit_count=0`.

PR #197, opened 2026-05-01
feat(fleet): add MODEL column to /gremlins status table

### LINKED ISSUE #195: feat(gremlins): add model column to /gremlins fleet output
## Context

The \`/gremlins\` status table shows kind, id, stage, liveness, age, boss, and description — but not which Claude model the gremlin is running. When gremlins are launched with non-default models (e.g. \`--impl-model opus\`), there's no way to tell from fleet output.

## Approach

Two-phase: (1) persist the model at launch time into \`state.json\`, and (2) surface it in the fleet table.

**Persisting at launch** is the right layer — the model args are fully known before any subprocess starts, no parsing of stream files required, and \`state.json\` is the single source of truth fleet.py already reads. We store the *intended* model string (e.g. \`claude-sonnet-4-6\`, \`opus\`, or the alias as passed), not the resolved one from stream events, which is acceptable for a fleet status view.

A single \`model\` field representing the implementation model is the right default column — that's the primary workhorse stage. If plan/review use different models, a drill-in can show those, but they add clutter to the summary row.

## Tasks

- [ ] **Audit CLI arg plumbing**: trace how model flags (\`-i\`/\`--impl-model\`, \`-p\`/\`--plan-model\`, etc.) flow from \`gremlins/cli.py\` → \`launcher.launch()\` → \`launcher._write_state()\` and identify the right field name(s) to add.
- [ ] **Persist model in state.json**: in \`launcher.py\` \`_write_state()\` (or wherever the initial state dict is built), write \`impl_model\` (and optionally \`plan_model\`, \`review_model\`) from the parsed args into the state file. Fall back to the SDK/env default label if the flag is absent.
- [ ] **Extend \`build_row()\`**: in \`gremlins/fleet/render.py\`, read \`impl_model\` (or a derived \`model\` summary) from the state dict and include it in the row.
- [ ] **Update the format string**: in \`gremlins/fleet/constants.py\` (or wherever \`FMT\` / column headers live), add a \`MODEL\` column. Choose a width that fits common aliases like \`sonnet\`, \`opus\`, \`haiku\` (≤10 chars) and short model IDs.
- [ ] **Handle missing field gracefully**: older state files won't have \`impl_model\`; default to \`"—"\` or \`"?\` so the table doesn't error on pre-existing gremlins.
- [ ] **Manual smoke test**: launch a gremlin with a non-default model, run \`/gremlins\`, confirm the model column populates correctly. Also verify old state files (missing the field) render without error.

## Open questions

- Should we show all three models (plan / impl / review) as separate fields or collapse to one representative column? A single \`model\` column keeps the table readable; full detail belongs in the drill-in view.
- What string should we store — the alias as passed (\`opus\`) or the resolved model ID (\`claude-opus-4-7\`)? Alias is simpler and survives model ID churn; resolved ID is more precise. Recommend alias with a note in the drill-in.

xbrianh:
Store `impl_model` in `state.json` at launch time (extracted from `-i` for localgremlin or `--model` for gh/boss gremlins, defaulting to `'sonnet'`), then surface it as a new 10-char `MODEL` column in the fleet table between `BOSS` and `DESCRIPTION`.

## Changes
- `gremlins/fleet/constants.py`: extend `FMT` with a `%-10s` slot for MODEL
- `gremlins/fleet/render.py`: read `impl_model`/`model` from state, add `model` key to row dict, update header and row format calls
- `gremlins/launcher.py`: add `_extract_impl_model()` helper, write `impl_model` into initial state
- `gremlins/tests/test_fleet.py`: three new unit tests covering primary field, fallback field, and missing-field dash
- `gremlins/tests/test_launcher.py`: two new integration tests for explicit and default model extraction

Closes #195

### DIFF
--- a/gremlins/launcher.py
+++ b/gremlins/launcher.py
@@ -132,7 +132,7 @@ def _extract_impl_model(pipeline_args: list, kind: str) -> str:
         for i, a in enumerate(args):
             if a == "--model" and i + 1 < len(args):
                 return args[i + 1]
-            if isinstance(a, str) and a.startswith("--model="):
+            if a.startswith("--model="):
                 return a[len("--model="):]
         return "sonnet"
 
--- a/gremlins/launcher.py
+++ b/gremlins/launcher.py
@@ -116,6 +116,27 @@ def _patch_state(state_dir: pathlib.Path, **fields) -> None:
         pass
 
 
+def _extract_impl_model(pipeline_args: list, kind: str) -> str:
+    """Extract the implementation model alias from pipeline_args.
+
+    Returns the alias as passed (e.g. 'opus', 'sonnet') or 'sonnet' as the
+    default when no model flag is present — matching each orchestrator's default.
+    """
+    args = list(pipeline_args)
+    if kind == "localgremlin":
+        for i, a in enumerate(args):
+            if a == "-i" and i + 1 < len(args):
+                return args[i + 1]
+        return "sonnet"
+    else:  # ghgremlin, bossgremlin use --model
+        for i, a in enumerate(args):
+            if a == "--model" and i + 1 < len(args):
+                return args[i + 1]
+            if isinstance(a, str) and a.startswith("--model="):
+                return a[len("--model="):]
+        return "sonnet"
+
+
 def _delete_patch_state(state_dir: pathlib.Path, delete_keys: tuple, **fields) -> None:
     """Remove delete_keys and merge fields into state.json atomically. Best-effort."""
     sf = state_dir / "state.json"
@@ -286,6 +307,7 @@ def launch(
         "description_explicit": desc_explicit,
         "parent_id": parent_id or "",
         "pipeline_args": stored_pipeline_args,
+        "impl_model": _extract_impl_model(stored_pipeline_args, kind),
         "stage": "starting",
         "pid": None,
     }
--- a/gremlins/tests/test_launcher.py
+++ b/gremlins/tests/test_launcher.py
@@ -250,6 +250,34 @@ def test_launch_persists_impl_model_default(lenv):
     _wait_for_finished(_gremlins_state_root(lenv) / gr_id, timeout=60)
 
 
+def test_launch_persists_impl_model_gh_space_form(lenv_with_gh):
+    """impl_model is extracted from space-separated --model flag for ghgremlin."""
+    lenv = lenv_with_gh
+    launcher = _launcher()
+    gr_id = launcher.launch(
+        "ghgremlin",
+        pipeline_args=("--model", "opus"),
+        instructions="test gh model space",
+    )
+    state = _read_state(_gremlins_state_root(lenv) / gr_id)
+    assert state["impl_model"] == "opus"
+    _wait_for_finished(_gremlins_state_root(lenv) / gr_id, timeout=60)
+
+
+def test_launch_persists_impl_model_gh_equals_form(lenv_with_gh):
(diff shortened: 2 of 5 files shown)


gremlins/launcher.py
```diff
@@ -116,6 +116,27 @@ def _patch_state(state_dir: pathlib.Path, **fields) -> None:
...
+    if kind == "localgremlin":
+        for i, a in enumerate(args):
+            if a == "-i" and i + 1 < len(args):
+                return args[i + 1]
+        return "sonnet"
+    else:  # ghgremlin, bossgremlin use --model
+        for i, a in enumerate(args):
+            if a == "--model" and i + 1 < len(args):
+                return args[i + 1]
+            if isinstance(a, str) and a.startswith("--model="):
```
xbrianh: The `isinstance(a, str)` guard is only on the `--model=` branch — the `-i` and space-separated `--model` branches above have no such check. Since `pipeline_args` is always a sequence of strings in practice, this is redundant and inconsistent. Either add the guard to all three branches or drop it here:
```python
if a.startswith("--model="):
    return a[len("--model="):]
```

gremlins/tests/test_launcher.py
```diff
@@ -225,6 +225,31 @@ def test_launch_persists_pipeline_args(lenv):
     _wait_for_finished(_gremlins_state_root(lenv) / gr_id, timeout=60)
 
 
+def test_launch_persists_impl_model_explicit(lenv):
```
xbrianh, line 228: The two new integration tests only exercise `localgremlin` + `-i`. The `_extract_impl_model` function has three distinct parse paths: `-i` (localgremlin), `--model <value>` (gh/boss), and `--model=<value>` (gh/boss). Only the first is covered by integration tests. Consider adding at least one `ghgremlin` case with `pipeline_args=("--model", "opus")` and one with `pipeline_args=("--model=haiku",)` to cover the `startswith` branch.

xbrianh (commented):
Clean, focused implementation. Two minor observations: an inconsistent `isinstance` guard in `_extract_impl_model` and a gap in integration test coverage for `ghgremlin`/`bossgremlin` kinds and the `--model=<value>` form.

gremlins/launcher.py
xbrianh: Fixed — dropped the  guard from the `--model=` branch so all three branches are consistent.

gremlins/tests/test_launcher.py
xbrianh: Added two integration tests using `lenv_with_gh`: `test_launch_persists_impl_model_gh_space_form` (space-separated `--model opus`) and `test_launch_persists_impl_model_gh_equals_form` (`--model=haiku`).

gremlins/launcher.py
```diff
@@ -116,6 +116,27 @@ def _patch_state(state_dir: pathlib.Path, **fields) -> None:
...
+                return a[len("--model="):]
+        return "sonnet"
```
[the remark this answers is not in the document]
xbrianh: The `pipeline_args` list is always constructed programmatically in Python (never parsed from a shell string via argparse), so the concatenated `-iopus` form never arises in practice — callers always pass `("-i", "opus")` as separate elements. No change needed.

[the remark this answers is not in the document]
xbrianh: Covered by the two new integration tests added in this pass: `test_launch_persists_impl_model_gh_space_form` and `test_launch_persists_impl_model_gh_equals_form`.

